PR #16, opened 2025-06-27
defer output fetching for task/terminal to tools, add `GetTaskOutputTool`, add terminal last command context to prompt

meganrogge:
Before:

Output was being provided in the prompt agent context.

Now, we reference tools in the prompt agent context that can be invoked if needed. We also now provide the last command context to the agent.

For tasks:
- added a new `GetTaskOutput` tool 
- because they don't have an associated terminal, for now, we're finding the terminal with the name of the task until we have better API vscode#234440 and passing that in to the tool

For terminals:
- associated the `id` that's assigned to a terminal in the `RunInTerminal` tool with the Copilot terminal, as `IKnownTerminal`
- returned `IKnownTerminal[]` from `getCopilotTerminals` so that the `id` can be passed to `GetTerminalOutput` tool

Screenshot 2025-06-27 at 12 53 54 PM

Screenshot 2025-06-27 at 12 55 50 PM

Fixes #252690

### DIFF
--- a/package.json
+++ b/package.json
@@ -420,7 +420,7 @@
 				"name": "copilot_getTerminalOutput",
 				"toolReferenceName": "getTerminalOutput",
 				"displayName": "%copilot.tools.getTerminalOutput.name%",
-				"modelDescription": "Get the output of a terminal command previous started with copilot_runInTerminal",
+				"modelDescription": "Get the output of a terminal command previously started with runInTerminal",
 				"tags": [],
 				"inputSchema": {
 					"type": "object",
@@ -432,6 +432,30 @@
 					},
 					"required": [
 						"id"
+					]
+				}
+			},
+			{
+				"name": "copilot_getTaskOutput",
+				"toolReferenceName": "getTaskOutput",
+				"displayName": "%copilot.tools.getTaskOutput.name%",
+				"modelDescription": "Get the output of a task to evaluate it for errors, idle status, and more",
+				"tags": [],
+				"inputSchema": {
+					"type": "object",
+					"properties": {
+						"name": {
+							"type": "string",
+							"description": "The name of the task to check."
+						},
+						"maxLinesToRetrieve": {
+							"type": "number",
+							"description": "The maximum number of lines to retrieve from the terminal output."
+						}
+					},
+					"required": [
+						"maxLinesToRetrieve",
+						"name"
 					]
 				}
 			},
@@ -1220,7 +1244,8 @@
 				"icon": "$(terminal)",
 				"tools": [
 					"runInTerminal",
-					"getTerminalOutput"
+					"getTerminalOutput",
+					"getTaskOutput"
 				]
 			},
 			{
--- a/src/extension/tools/node/getTaskOutputTool.tsx
+++ b/src/extension/tools/node/getTaskOutputTool.tsx
@@ -1,0 +1,48 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as l10n from '@vscode/l10n';
+import type * as vscode from 'vscode';
+import { ITerminalService } from '../../../platform/terminal/common/terminalService';
+import { LanguageModelTextPart, LanguageModelToolResult } from '../../../vscodeTypes';
+import { ToolName } from '../common/toolNames';
+import { ToolRegistry } from '../common/toolsRegistry';
+
+export interface ITaskOptions {
+	name: string;
+	maxLinesToRetrieve: number;
+}
+
+/**
+ * Tool to provide output for a given task terminal.
+ */
+export class GetTaskOutputTool implements vscode.LanguageModelTool<ITaskOptions> {
+	public static readonly toolName = ToolName.GetTaskOutput;
+
+	constructor(@ITerminalService private readonly terminalService: ITerminalService) {
+
+	}
+	async invoke(options: vscode.LanguageModelToolInvocationOptions<ITaskOptions>, token: vscode.CancellationToken) {
+		// TODO:@meganrogge when there's API to determine if a terminal is a task, improve this vscode#234440
+		const terminal = this.terminalService.terminals.find(t => t.name === options.input.name);
+		if (!terminal) {
+			return;
+		}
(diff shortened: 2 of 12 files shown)


src/extension/tools/node/getTaskOutputTool.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+	}
+	async invoke(options: vscode.LanguageModelToolInvocationOptions<ITaskOptions>, token: vscode.CancellationToken) {
+		// TODO:@meganrogge when there's API to determine if a terminal is a task, improve this vscode#234440
+		const terminal = this.terminalService.terminals.find(t => t.name === options.input.name);
+		if (!terminal) {
+			return;
+		}
+		// 40 chars per line and 60k chars is about 1500 lines
+		const buffer = this.terminalService.getBufferForTerminal(terminal, Math.min(options.input.maxLinesToRetrieve, 1500));
```
meganrogge: @tyriar could use your guidance here pls

src/platform/terminal/vscode/terminalBufferListener.ts
```diff
@@ -97,8 +100,8 @@ export function getActiveTerminalShellType(): string {
 
 function appendLimitedWindow<T>(target: T[], data: T) {
 	target.push(data);
-	if (target.length > 20000) {
-		// 20000 data events should capture a minimum of about twice the typical visible area
+	if (target.length > 40) {
```
meganrogge, line 104: @Tyriar not sure what we should go with here

meganrogge: fwiw, when debugging, I never saw the target.length get even close to 40

Tyriar: I was writing up a big response to this but decided against it. Since each value could theoretically be very small (1 byte), or very large (~5000 bytes when I run ls just now), how about we just ignore what I said before and leave it as is for now as this is another perfect candidate to be moved to core.

Moving to core means not relying on the forever proposed `onDidWriteTerminalData` API, not giving the tool data events instead of the actual buffer content, and not needing to incur any cost of duplicating the events/buffer as all since we'd be pulling it from the source of truth.

src/extension/tools/node/getTaskOutputTool.tsx
Tyriar: `getActiveTerminalBuffer` and `getBufferForTerminal` do not currently return lines, but data events on separate lines, plus the regular line breaks. Here's what's inside `terminalBuffers` for this terminal:

image

image

So `maxLines` is actually incorrect here and these output functions never really worked properly since they returned a `\n` between every data event. I'd replace the `.join('\n')` in both, then cap this one to some reasonable character count based on the typical context window we have available. Then next iteration move these terminal buffer tools to core like in my other comment.

package.json
```diff
@@ -435,6 +435,30 @@
...
+			{
+				"name": "copilot_getTaskOutput",
+				"toolReferenceName": "getTaskOutput",
+				"displayName": "%copilot.tools.getTaskOutput.name%",
+				"modelDescription": "Get the output of a task to evaluate it for errors, idle status, and more",
+				"tags": [],
+				"inputSchema": {
+					"type": "object",
+					"properties": {
+						"name": {
```
roblourens: `copilot_runVsCodeTask` takes an "ID", let's make sure these are aligned.

roblourens: Please also look at `AgentTasksInstructions` and see if any alignment is needed

src/extension/tools/node/getTaskOutputTool.tsx
```diff
@@ -0,0 +1,48 @@
...
+	constructor(@ITerminalService private readonly terminalService: ITerminalService) {
+
+	}
+	async invoke(options: vscode.LanguageModelToolInvocationOptions<ITaskOptions>, token: vscode.CancellationToken) {
+		// TODO:@meganrogge when there's API to determine if a terminal is a task, improve this vscode#234440
+		const terminal = this.terminalService.terminals.find(t => t.name === options.input.name);
+		if (!terminal) {
+			return;
+		}
+		// 40 chars per line and 60k chars is about 1500 lines
```
roblourens: Then can you cap it based on chars? Lines could be very long and we need to be careful with token usage.

test/simulation/baseline.json
```diff
@@ -11805,9 +11805,9 @@
   {
     "name": "toolCalling [panel] - find all phone numbers in markdown files in the codebase",
```
meganrogge, line 11806: output for this one is:

```
Scenario toolCalling [panel] - find all phone numbers in markdown files in the codebase failed due to an error:
Error: Tool call name "read_file" does not match expected tool call names.
    at validateToolCallExpectation (/Users/meganrogge/Repos/vscode-copilot-chat/test/e2e/toolSimTest.ts:92:11)
    at /Users/meganrogge/Repos/vscode-copilot-chat/test/e2e/toolSimTest.ts:74:10
    at /Users/meganrogge/Repos/vscode-copilot-chat/test/e2e/scenarioTest.ts:226:36
    at executeTestOnce (/Users/meganrogge/Repos/vscode-copilot-chat/test/testExecutor.ts:441:3)
    at CTask.execute (/Users/meganrogge/Repos/vscode-copilot-chat/test/taskRunner.ts:20:19)
  ▼ [ 4 ->  0] toolCalling [panel] - find all phone numbers in markdown files in the codebase (️️️💸 cache miss)
  = [10] toolCalling [panel] - I'm using git, create
```

meganrogge:
@tyriar or @roblourens , since you're both already contributors to this repo, I believe it will allow you to update baseline. Could you pls do so and push to this branch? See slack for discussion

src/platform/terminal/vscode/terminalBufferListener.ts
```diff
@@ -28,10 +28,14 @@ export function getBufferForTerminal(terminal?: Terminal, maxLines: number = 100
 	if (!buffer) {
 		return '';
 	}
-	const start = Math.max(0, buffer.length - maxLines);
-	return buffer.slice(start).join('\n');
+	const joined = buffer.join('');
+	const start = Math.max(0, joined.length - maxChars);
+	return joined.slice(start);
```
roblourens, line 33: A nice trick- `joined.slice(-maxChars)` :)

roblourens (commented):
Can be in the future, but a unit test would be nice to be sure that we're rendering things nicely for all the different types of tasks and terminal states. See agentPrompt.spec.tsx

PR #27, opened 2025-06-29
Allow fetch to handle images returned from internal fetch

TylerLeonhardt:
The scenario is if they `#fetch file:///foo.png` or some other kind of non-http scheme, it'll fall back to the file service and just grab the file contents.

This needed updating to handle binary data.

Once this goes in, I can update the internal tool to actually send image data back.

ref https://github.com/microsoft/vscode/issues/250495


bpasero (approved):

TylerLeonhardt:
fyi @connor4312

PR #29, opened 2025-06-29
fetchWebPageTool stests

TylerLeonhardt:


roblourens (approved):
Nice! Do you find these helpful? I'm thinking about making these tests optional/something that we can run async but don't block PRs, what do you think?

TylerLeonhardt:
I'm cool with that. I added this because there have been 1 or 2 reports of "misses"... although I did update the description semi recently and haven't heard anyone yell yet

PR #39, opened 2025-06-30
Refactor how thinking id and summary is handled and add tests

karthiknadig:


DonJayamanne (approved):

src/platform/networking/node/stream.ts
```diff
@@ -410,6 +402,16 @@ export class SSEProcessor {
 
 					this.logChoice(choice);
 
+					if (isChainOfThoughtChoice(choice)) {
```
roblourens, line 405: Is this a temporary thing?

karthiknadig: Yes. I plan to get rid of this soon.

PR #56, opened 2025-06-30
Improve the summarization with better prompt cache 

caohy1988:
# Optimize Conversation History Summarization Cache Strategy

## Summary

This PR dramatically optimizes prompt caching for conversation history summarization by **aligning the prompt structure to reuse cache from normal agent conversations**. 

## Background

The conversation history summarization feature supports two modes:
- **Full mode**: Comprehensive summary with detailed analysis  
- **Simple mode**: Fallback mode when full mode fails or is configured

**Key Insight**: Instead of optimizing cache within summarization, we can reuse the cache that's already populated from the user's normal agent conversations.

### 🎯 **Cache Reuse from Normal Agent Conversations**
- **Identical prompt structure**: Restructured `AgentSummarizationPrompt` to match `AgentPrompt` exactly
- **Shared cache utilization**: Leverages cache already populated by regular agent usage
- **Warm cache advantage**: No cold start penalty - cache is pre-populated from normal conversations

### 🔧 **Unified Mode Handling** 
- **Same comprehensive prompt for both modes**: Eliminated prompt structure differences between simple/detailed modes
- **History-only differentiation**: Modes differ only in conversation history content (truncated vs full)
- **Shared helper function**: `getAgentInstructions()` ensures instruction logic stays synchronized

```diff
// Before: Different structures = no cache reuse
AgentPrompt: SystemMsg → Instructions → UserMsg → GlobalContext [CACHE] → History
SummaryPrompt: SystemMsg → Instructions → UserMsg → SummaryInstructions [CACHE] → History

// After: Identical structure = cache reuse
AgentPrompt: SystemMsg → Instructions → UserMsg → GlobalContext [CACHE] → History  
SummaryPrompt: SystemMsg → Instructions → UserMsg → GlobalContext [CACHE] → History + Summary Query
```

### 🏗️ **Shared Cache Architecture**

**Both Normal Agent and Summarization now share identical cached sections:**

```
┌─────────────────────────────────────┐
│     CACHED FROM NORMAL AGENT       │ ← Reused by summarization
├─────────────────────────────────────┤
│ SystemMessage                       │
│ ├── Copilot Identity Rules          │
│ └── Safety Rules                    │
├─────────────────────────────────────┤
│ Agent Instructions                  │ ← EXACT same logic
│ ├── DefaultAgentPrompt OR           │
│ └── SweBenchAgentPrompt             │
├─────────────────────────────────────┤
│ UserMessage                         │ ← Same custom instructions
│ ├── Custom Instructions             │
│ └── Mode Instructions (if any)      │
├─────────────────────────────────────┤
│ UserMessage                         │ ← CRITICAL: Same structure  
│ └── GlobalAgentContext + [CACHE]    │   as normal AgentPrompt
└─────────────────────────────────────┘
         🔄 CACHE REUSE BOUNDARY
┌─────────────────────────────────────┐
│         DYNAMIC SECTION             │ ← Only this varies
├─────────────────────────────────────┤
│ ConversationHistoryForSummarization │
│ ├── Simple: SimpleSummarizedHistory │ ← Truncated for simple mode
│ └── Full: ConversationHistory       │ ← Complete for full mode
├─────────────────────────────────────┤
│ WorkingNotebookSummary (optional)   │
├─────────────────────────────────────┤
│ UserMessage                         │ ← Comprehensive instructions
│ └── Detailed Summary Query          │   + format in final message
└─────────────────────────────────────┘
```


src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -24,28 +25,73 @@ import { ToolCallingLoop } from '../../../intents/node/toolCallingLoop';
...
+				<UserMessage>
+					<CustomInstructions languageId={undefined} chatVariables={this.props.promptContext.chatVariables} />
+					{this.props.promptContext.modeInstructions && <Tag name='customInstructions'>
+						Below are some additional instructions from the user.<br />
+						<br />
+						{this.props.promptContext.modeInstructions}
+					</Tag>}
+				</UserMessage>
+				<GlobalAgentContext enableCacheBreakpoints={false} />
 				<SystemMessage priority={this.props.priority}>
```
roblourens: This might help a bit, but since this comes before the history, it would still be a cache miss on the history, which is most of it, right?

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -24,28 +25,73 @@ import { ToolCallingLoop } from '../../../intents/node/toolCallingLoop';
...
+			/>;
+
+		// Use the same comprehensive summarization prompt for both modes
+		// The difference in output detail will come from the amount of history provided
+		const summarizationQuery = this.getSummarizationQuery();
+
 		return (
 			<>
+				{/* Reuse the same system prompt structure as AgentPrompt for maximum caching */}
+				<SystemMessage>
```
roblourens, line 79: I would refactor so that this doesn't duplicate where the main prompt is constructed, since they would have to be kept perfectly in sync.

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -148,37 +194,49 @@ export class ConversationHistorySummarizationPrompt extends PromptElement<Conver
...
-					Focus particularly on:<br />
-					- The specific agent commands/tools that were just executed<br />
-					- The results returned from these recent tool calls (truncate if very long but preserve key information)<br />
-					- What the agent was actively working on when the token budget was exceeded<br />
-					- How these recent operations connect to the overall user goals<br />
-
-					Include all important tool calls and their results as part of the appropriate sections, with special emphasis on the most recent operations.
+				<UserMessage priority={900}>
+					{summarizationQuery}
+					<cacheBreakpoint type={CacheType} />
```
roblourens: This is tricky, need cache breakpoints for full mode, but not simple mode

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -194,7 +194,8 @@ export class AgentSummarizationPrompt extends PromptElement<ConversationHistoryS
 
 					This summary should serve as a comprehensive handoff document that enables seamless continuation of all active work streams while preserving the full technical and contextual richness of the original conversation.<br />
 				</SystemMessage>
-				{/* Dynamic history components - not cached for optimal cache efficiency */}
+				<cacheBreakpoint type={CacheType} />
```
roblourens: Ah, I think I didn't explain well enough- I'm hoping to hit the prompt cache from the user's own previous normal chat requests. This could hit a cache from the user's previous summarization requests, but I don't think that will happen that often, and doesn't include most of the content anyway

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
caohy1988: refactored

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
caohy1988: updated the cache point only work for simple mode

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
caohy1988: Update the cache setting based on your feedback, can you take another look?

caohy1988:
update the PR and resolved all your feedback, please take another look, thanks @roblourens

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
caohy1988: good point, update the prompt cache structured

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -429,27 +499,11 @@ class ConversationHistorySummarizer {
...
-							description: tool.description,
-							parameters: tool.inputSchema && Object.keys(tool.inputSchema).length ? tool.inputSchema : undefined
-						}, type: 'function'
-					})),
-					(tool, rule) => {
-						this.logService.logger.warn(`Tool ${tool} failed validation: ${rule}`);
-					},
-				),
-			} : undefined;
+			// Don't include tools in the summarization request to prevent tool calling
```
roblourens: This will break the prompt cache

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -429,27 +499,11 @@ class ConversationHistorySummarizer {
...
 			summaryResponse = await endpoint.makeChatRequest('summarizeConversationHistory', ToolCallingLoop.stripInternalToolCallIds(summarizationPrompt), undefined, this.token ?? CancellationToken.None, ChatLocation.Other, undefined, {
 				temperature: 0,
 				stream: false,
-				...toolOpts
+				tool_choice: 'none' // Explicitly disable tool calling during summarization
```
roblourens: Tools must be provided to use tool_choice- I believe some provider errors when we do thsi

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -24,161 +25,218 @@ import { ToolCallingLoop } from '../../../intents/node/toolCallingLoop';
...
 
-class WorkingNotebookSummary extends PromptElement<NotebookSummaryProps> {
-	override async render(state: void, sizing: PromptSizing) {
-		return (
-			<UserMessage>
-				This is the current state of the notebook that you have been working on:<br />
-				<NotebookSummary notebook={this.props.notebook} />
-			</UserMessage>
-		);
+	private getSummarizationQuery(): string {
```
roblourens: String containing `<br />`, let's stick to prompt-tsx

roblourens (commented):
Thanks for doing this but I'd like to see in some eval that this approach works as well as using the system prompt

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
roblourens: What I meant is that we should have one place where we render the prompt, so that we aren't duplicating the same content in two places. You don't have to do that work, it might be complicated, let's decide whether this approach works as well first.

caohy1988:
> Thanks for doing this but I'd like to see in some eval that this approach works as well as using the system prompt before you spend a lot of time on it :)

would run a side by side compare using current approach and the simple system prmopt approach for summary with MSBench

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
caohy1988: resolved

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
caohy1988: add back tools

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
caohy1988: update to use prompt tsx

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
caohy1988: got it, would run benchmark first to get which approach to go

PR #66, opened 2025-06-30
edits: try to normalize indentation in apply_patch

connor4312:
This makes a change such that the apply_patch matcher will add
indentation in the patch such that the indentation in the first line
of context given in the patch matches that which was found in the file.
This fixes 4.1's habit of giving an entirely outdented patch.

You might think we should instead run save participants to use the
formatter for the language. This is fine for JS and friends but not for
languages with syntatic whitespace like Python or YAML, and this
approach seems to work well in my local testing.

swebench data coming in a few hours which'll inform mergability.
swebench is Python which should be a good acid test for this.


connor4312:
swebench score is +/- the same with this change. I'll add an EXP flag to control behavior and ship it

roblourens (dismissed):

roblourens (approved):

PR #69, opened 2025-07-01
discourage semantic_search and increase parallelization

yemohyleyemohyle:
Semantic search in the current implementation, when used, contributes the most tokens to the prompt often doubling the price. Here we propose a warning message in the instructions to avoid the tool. To balance some loss in accuracy for  sonnet4 model we propose occasional just in time reminder to parallelize tool calls. Assuming limited number of turns, it fits more tool calls increasing trajectory length hence increasing accuracy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -416,7 +416,8 @@
 						"explanation",
 						"isBackground"
 					]
-				}
+				},
+				"when": "!config.chat.agent.terminal.coreToolsEnabled"
 			},
 			{
 				"name": "copilot_getTerminalOutput",
@@ -435,7 +436,8 @@
 					"required": [
 						"id"
 					]
-				}
+				},
+				"when": "!config.chat.agent.terminal.coreToolsEnabled"
 			},
 			{
 				"name": "copilot_getTaskOutput",


src/extension/prompts/node/agent/agentInstructions.tsx
```diff
@@ -46,7 +46,9 @@ export class DefaultAgentPrompt extends PromptElement<DefaultAgentPromptProps> {
 				When reading files, prefer reading large meaningful chunks rather than consecutive small sections to minimize tool calls and gain better context.<br />
 				Don't make assumptions about the situation- gather context first, then perform the task or answer the question.<br />
 				{!this.props.codesearchMode && <>Think creatively and explore the workspace in order to make a complete fix.<br /></>}
-				Don't repeat yourself after a tool call, pick up where you left off.<br />
+				You pimary objective is to resolve the user's request while minimizing the cost to the user and the time to reach solution.<br />
```
OrenMe:
```suggestion
				You primary objective is to resolve the user's request while minimizing the cost to the user and the time to reach solution.<br />
```

PR #94, opened 2025-07-02
Indicate when no active tasks/terminals are found to avoid hallucination about active tasks/terminals that have been cancelled 

meganrogge:
Fixes https://github.com/microsoft/vscode-internalbacklog/issues/5571
Fixes https://github.com/microsoft/vscode-internalbacklog/issues/5572

Also adds some logs to investigate when there's no result.

Repro: 
- run tasks, ask about active tasks ✅ . 
- kill the tasks, ask about active tasks 🐛 says they're still running

I noticed this did not reproduce if I cleared the chat. It seems that when we provided `undefined` as the return value when there were no active tasks/terminals, it used the old context. 

The fix is to return the string indicating there are no active tasks / terminals.

Also:
- added logs to `GetTaskOutput`
- specified when `isActive` for the task in `agentPrompt`, saying that's when the `GetTaskOutput` tool could be run
- prevented `GetTaskOutput` from running for inactive task

Before:

https://github.com/user-attachments/assets/d9560d2d-7466-465c-883e-083d11bf9aea

After:

https://github.com/user-attachments/assets/86ff6106-6864-43d2-8100-b28f2d30c190

### DIFF
--- a/test/outcome/edit-toolcalling-panel.json
+++ b/test/outcome/edit-toolcalling-panel.json
@@ -2,7 +2,7 @@
 	{
 		"name": "edit (toolCalling) [panel] - does not read",
 		"requests": [
-			"e290732cf0cefa6f11bd71f2cf5ca1962d6bf2209f754f5a55674a05ebd88d4c"
+			"f0d027e3070021bfd5f8a123f59b22dc622865d06a4cfd7bf54c20a614153be1"
 		]
 	}
 ]
--- a/test/outcome/fetchwebpagetool-toolcalling-panel.json
+++ b/test/outcome/fetchwebpagetool-toolcalling-panel.json
@@ -2,25 +2,25 @@
 	{
 		"name": "fetchWebPageTool (toolCalling) [panel] - multiple URLs boundary test with 6 URLs",
 		"requests": [
-			"5de587fac8cc3ef067a89c5bb4a081b9d10825bf20c1b1e7ec4d02ff43e7e0d1"
+			"2e75476f9e0169ddcf3d9cc9e0ef8acdd869fc8fc25e74e2b2d8977582209b81"
 		]
 	},
 	{
 		"name": "fetchWebPageTool (toolCalling) [panel] - multiple URLs handling",
 		"requests": [
-			"c71f965a43acc4e43c6a746f89ccf414c1e4e399d300efbfbca6e7257bbdc676"
+			"41c94059765b2fe7a2abf1e73dbb44242476e07649a86eccca5ec3f99352db17"
 		]
 	},
 	{
 		"name": "fetchWebPageTool (toolCalling) [panel] - proper URL validation and query handling",
 		"requests": [
-			"8ae1dfca289add225ae90422c279508dd852e1dc8158e615914e1644cee82dd4"
+			"90b614cb6659ea4696166cb1827b6f2ec5729673feaf8f609adc66ca44cd0b4f"
 		]
 	},
 	{
 		"name": "fetchWebPageTool (toolCalling) [panel] - query parameter extraction",
 		"requests": [
-			"6de958ad66c1666b33ee3d02af5a0d903d8e41e7759957c6b15fbb7537e6288d"
+			"b6e5ff7774b7e0c6071983117da6f95c3fe569ce72507fa8e3798855f32959af"
 		]
 	}
 ]
--- a/test/outcome/findfilestool-toolcalling-panel.json
+++ b/test/outcome/findfilestool-toolcalling-panel.json
@@ -2,7 +2,7 @@
 	{
 		"name": "findFilesTool (toolCalling) [panel] - proper glob patterns",
 		"requests": [
-			"6202d52922ab15893305909f6f0aaf23aae6f772a6b12fc8bcf03b41ff0a1bdf"
+			"b3b5f8896144f3a9b6e7741c7f1038e749178c1189a9f1dcfc7979e596399ca2"
 		]
 	}
 ]
--- a/test/outcome/notebooks-toolcalling-panel.json
+++ b/test/outcome/notebooks-toolcalling-panel.json
@@ -2,25 +2,25 @@
 	{
 		"name": "notebooks (toolCalling) [panel] - Edit cell tool",
 		"requests": [
-			"6df4524b79c30ac2d84fa04183f35286c5456c91b3b6d5c62425fe34e8360d0a"
+			"d77abb1e79e0e447c8dc59f6343255726d19553f319f8b9656d762162ff4e9bd"
 		]
 	},
 	{
 		"name": "notebooks (toolCalling) [panel] - Run cell at a specific index",
 		"requests": [
-			"ffcec1ca32891e5143befb239df9aa9226761b28d44cb0324f37279d7cda5379"
+			"29dbbce5e336bea985843bbd8d8aae9d58dbf1398af5d7b4ce41c2244cd20995"
 		]
 	},
 	{
 		"name": "notebooks (toolCalling) [panel] - Run cell tool",
 		"requests": [
-			"c80f440ac022d1601717db496019c3316a28c6537c8f337976e175f711938d5a"
+			"408cb19aeb853f456f3d8a70cfb2ddfc694e018282e6e3c7e7cc4069dca304c3"
 		]
 	},
 	{
 		"name": "notebooks (toolCalling) [panel] - Run cell tool should avoid running markdown cells",
 		"requests": [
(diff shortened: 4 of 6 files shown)


roblourens (dismissed):
So the problem is that it sees the context about the task in a previous turn, then it's just gone in the next turn, and it doesn't realize that it's now gone? This makes sense but it's a little unfortunate that we have to include this in every single prompt even when tasks and terminals are not being used. I guess ideally we would only include this if something _changed_. If you wanted to do that, we could figure out how to track that state, likely on the ChatResult metadata.

meganrogge:
@roblourens yeah, that's what I think is happening. It's just a short string - probably very few tokens for this? Wouldn't the issue persist if we were to do that - when it changes, it's included. The next turn, it's not included, so it uses the old info?

roblourens:
The issue to me is more about directing the model's focus. We should include context for things that are likely to be important. If we say "there are no terminals" on every request then at some point some model might tend to think that you want it to create a terminal or something. Really hard to predict. This is why we need to get rid of implicit context too, because it frequently thinks that I'm only interested in my open file when I'm asking a general question, although that's also different because we include a huge chunk of content.

> Wouldn't the issue persist if we were to do that - when it changes, it's included. The next turn, it's not included, so it uses the old info?

But if it didn't change, then the previous state is not out of date right?

meganrogge:
Oh, I think I understand what you're saying now. We'd say no tasks or terminals are active the first time that is the case.

src/extension/prompts/node/base/terminalAndTaskState.tsx
```diff
@@ -16,6 +16,12 @@ export interface TerminalAndTaskStateProps extends BasePromptElementProps {
  * PromptElement that gets the current task and terminal state for the chat context.
  */
 export class TerminalAndTaskStatePromptElement extends PromptElement<TerminalAndTaskStateProps> {
+	/**
+	 * Tracks previous counts of tasks and terminals in the previous render.
+	 */
+	private _lastTaskCount: number = 0;
```
roblourens: Are you sure this works? I don't believe prompt element instances are reused, they would be instantiated for each prompt construction. 

And if it did, wouldn't you need to store it per session? And it won't be persisted across window reload, and it won't know about session undos and edits. That's why I recommend we store data on the chatresult metadata and not try to store data separately from the chat model.

package.json
```diff
@@ -645,7 +645,7 @@
 				"icon": "$(json)",
 				"userDescription": "%github.copilot.tools.createAndRunTask.userDescription%",
 				"when": "config.github.copilot.chat.agent.runTasks",
-				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run. \n-You can check the output of a task later on by using copilot_getTaskOutput.",
+				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run.",
```
meganrogge, line 648:
```suggestion
				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run.  \n-You can check the output of a task later on by using copilot_getTaskOutput. Do not say the task has exited unless you are sure it has.",
```

package.json
```diff
@@ -645,7 +645,7 @@
 				"icon": "$(json)",
 				"userDescription": "%github.copilot.tools.createAndRunTask.userDescription%",
 				"when": "config.github.copilot.chat.agent.runTasks",
-				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run. \n-You can check the output of a task later on by using copilot_getTaskOutput.",
+				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run.  \n-You can check the output of a task later on by using copilot_getTaskOutput. Do not say the task has exited unless you are sure it has.",
```
meganrogge:
```suggestion
				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run. \n-You can check the output of a task later on by using copilot_getTaskOutput. Do not say the task has exited unless you are sure it has.",
```

package.json
```diff
@@ -645,7 +645,7 @@
 				"icon": "$(json)",
 				"userDescription": "%github.copilot.tools.createAndRunTask.userDescription%",
 				"when": "config.github.copilot.chat.agent.runTasks",
-				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run. \n-You can check the output of a task later on by using copilot_getTaskOutput.",
+				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run. \n-You can check the output of a task later on by using copilot_getTaskOutput. Do not say the task has exited unless you are sure it has.",
```
meganrogge:
```suggestion
				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run. \n-You can check the output of a task later on by using copilot_getTaskOutput. Do not say the task has exited unless you are sure it has. You can notice errors, but they may not mean the task has failed unless they're at the bottom of the output.",
```

package.json
```diff
@@ -645,7 +645,7 @@
 				"icon": "$(json)",
 				"userDescription": "%github.copilot.tools.createAndRunTask.userDescription%",
 				"when": "config.github.copilot.chat.agent.runTasks",
-				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run. \n-You can check the output of a task later on by using copilot_getTaskOutput.",
+				"modelDescription": "For a workspace, this tool will create a task based on the package.json, README.md, and project structure so that the project can be built and run. \n-You can check the output of a task later on by using copilot_getTaskOutput. Do not say the task has exited unless you are sure it has. You can notice errors, but they may not mean the task has failed unless they're at the bottom of the output.",
```
connor4312: I think this is something that should be in the tool output when the model calls it. The model doesn't need to know this information in advance of the tool call, so no need to pay the token cost for every prompt

src/extension/prompts/node/agent/agentPrompt.tsx
```diff
@@ -549,10 +549,18 @@ class AgentTasksInstructions extends PromptElement {
...
+			The following tasks can be executed using the {ToolName.RunTask} tool if they are not already running:<br />
 			{taskGroups.map(([folder, tasks]) =>
 				<Tag name='workspaceFolder' attrs={{ path: this._promptPathRepresentationService.getFilePath(folder) }}>
-					{tasks.map((t, i) => <Tag name='task' attrs={{ id: `${t.type}: ${t.label || i}` }}>{this.makeTaskPresentation(t)}</Tag>)}
+					{tasks.map((t, i) => {
+						const isActive = this._tasksService.isTaskActive(t);
+						return (
+							<Tag name='task' attrs={{ id: `${t.type}: ${t.label || i}` }}>
+								{this.makeTaskPresentation(t)}
+								{isActive && <> (This task is currently running. You can use the {ToolName.GetTaskOutput} tool to view its output. Do not say the task has exited unless you are sure it has. You can notice errors, but they may not mean the task has failed unless they're at the bottom of the output.)</>}
```
connor4312: Likewise here, if a task is here that means it hasn't exited, and I don't the model would say that until it tried to view the output

src/extension/tools/node/getTaskOutputTool.tsx
```diff
@@ -30,16 +31,19 @@ export class GetTaskOutputTool implements vscode.LanguageModelTool<ITaskOptions>
...
+			this.logService.logger.debug('getTaskOutputTool returning undefined: no matching label for task ' + options.input.id);
 			return;
 		}
 		// TODO:@meganrogge when there's API to determine if a terminal is a task, improve this vscode#234440
 		const terminal = this.terminalService.terminals.find(t => t.name === label);
 		if (!terminal) {
+			this.logService.logger.debug('getTaskOutputTool returning undefined: no terminal for task: ' + options.input.id + ' label: ' + label + ' terminal names: ' + this.terminalService.terminals.map(t => t.name).join(', '));
 			return;
 		}
 		const buffer = this.terminalService.getBufferForTerminal(terminal, Math.min(options.input.maxCharsToRetrieve ?? 16000, 16000));
```
connor4312, line 49: Probably down after "Output for task" we should say "Task is still running" etc. to reflect its state

src/extension/prompts/node/agent/agentPrompt.tsx
meganrogge: based on my testing, sometimes it would say that. but maybe I should move this into the get task output tool?

PR #101, opened 2025-07-02
Draft: Tool filtering experiment

digitarald:
For discussion.

This PR introduces experimental semantic filtering for language model tools, specifically targeting MCP (Model Context Protocol) servers when the number of available tools exceeds 128.

**Scope**
- Reduces tool set size when too many MCP tools are available
- Improves relevance of available tools based on user query
- Preserves all non-MCP tools to avoid breaking existing functionality

**How it works**
1. **Tool Embeddings Cache** (src/extension/tools/common/toolEmbeddingsCache.ts) caches embeddings for tools to improve performance
    - Uses `text3small_512` embedding type
    - Caches tool embeddings based on name, description, and schema
    - LRU cache with default size of 256 entries
    - Supports both tool and query embeddings
2. **Semantic Tool Filtering** src/extension/tools/vscode-node/toolsService.ts)
    1. Triggers when tools > 128 and feature is enabled
    2. Separates MCP tools from others (only MCP tools get filtered)
    3. Uses semantic similarity (min score 0.3) to rank MCP tools
    4. Keeps top relevant MCP tools + all non-MCP tools (max 128 total)
3. **Configuration** `github.copilot.chat.tools.semanticSearch.enabled`, Boolean (default: false)

**TODO**
- [ ] Gets called way too often, and each filtering (cached) takes about 300ms on my M4 Pro
- [ ] Misses context from custom instructions and attached files

### DIFF
--- a/package.json
+++ b/package.json
@@ -2433,6 +2433,14 @@
 						],
 						"markdownDescription": "%github.copilot.config.agent.thinkingTool%"
 					},
+					"github.copilot.chat.tools.semanticSearch.enabled": {
+						"type": "boolean",
+						"default": false,
+						"tags": [
+							"experimental"
+						],
+						"markdownDescription": "%github.copilot.config.tools.semanticSearch.enabled%"
+					},
 					"github.copilot.chat.edits.suggestRelatedFilesFromGitHistory": {
 						"type": "boolean",
 						"default": true,
--- a/src/extension/tools/vscode-node/toolsService.ts
+++ b/src/extension/tools/vscode-node/toolsService.ts
@@ -4,6 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from 'vscode';
+import { ConfigKey, IConfigurationService } from '../../../platform/configuration/common/configurationService';
+import { IEmbeddingsComputer } from '../../../platform/embeddings/common/embeddingsComputer';
 import { ILogService } from '../../../platform/log/common/logService';
 import { Lazy } from '../../../util/vs/base/common/lazy';
 import { IInstantiationService } from '../../../util/vs/platform/instantiation/common/instantiation';
@@ -51,9 +53,11 @@ export class ToolsService extends BaseToolsService {
 
 	constructor(
 		@IInstantiationService instantiationService: IInstantiationService,
-		@ILogService logService: ILogService
+		@ILogService logService: ILogService,
+		@IEmbeddingsComputer embeddingsComputer: IEmbeddingsComputer | undefined,
+		@IConfigurationService private readonly configurationService: IConfigurationService
 	) {
-		super(logService);
+		super(logService, embeddingsComputer);
 		this._copilotTools = new Lazy(() => new Map(ToolRegistry.getTools().map(t => [t.toolName, instantiationService.createInstance(t)] as const)));
 	}
 
@@ -75,11 +79,11 @@ export class ToolsService extends BaseToolsService {
 		// Can't actually implement this in prod, name is not exposed
 		throw new Error('This method for tests only');
 	}
-
-	getEnabledTools(request: vscode.ChatRequest, filter?: (tool: vscode.LanguageModelToolInformation) => boolean | undefined): vscode.LanguageModelToolInformation[] {
+	async getEnabledTools(request: vscode.ChatRequest, filter?: (tool: vscode.LanguageModelToolInformation) => boolean | undefined): Promise<vscode.LanguageModelToolInformation[]> {
 		const toolMap = new Map(this.tools.map(t => [t.name, t]));
 
-		return this.tools.filter(tool => {
+		// First, apply explicit filtering to reduce the tool set
+		let candidateTools = this.tools.filter(tool => {
 			// 0. Check if the tool was disabled via the tool picker. If so, it must be disabled here
 			const toolPickerSelection = request.tools.get(getContributedToolName(tool.name));
 			if (toolPickerSelection === false) {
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -2433,6 +2433,14 @@
 						],
 						"markdownDescription": "%github.copilot.config.agent.thinkingTool%"
 					},
+					"github.copilot.chat.tools.semanticSearch.enabled": {
+						"type": "boolean",
+						"default": false,
```
marklicata, line 2438: Is there any reason we wouldn't want this feature enabled by default?

src/extension/tools/vscode-node/toolsService.ts
```diff
@@ -113,5 +117,46 @@ export class ToolsService extends BaseToolsService {
...
+		let semanticQuery: string | undefined;
+		if (request.prompt) {
+			// Use the user's prompt as the semantic query
+			semanticQuery = request.prompt;
+		}
+
+		// After explicit filtering, apply semantic filtering if we still have too many tools
+		// Only apply semantic filtering to tools tagged with 'mcp'
+		const isSemanticSearchEnabled = this.configurationService.getConfig(ConfigKey.ToolsSemanticSearchEnabled);
+		if (semanticQuery && candidateTools.length > 128 && isSemanticSearchEnabled) {
```
marklicata, line 131: Should there be a config for minimum threshold of tools? Right now it's hard-coded to 128 and likely very hard to run down if this condition was the reason we ran a semantic match.

src/extension/tools/vscode-node/toolsService.ts
```diff
@@ -113,5 +117,46 @@ export class ToolsService extends BaseToolsService {
...
+			try {
+				// Separate MCP tools from other tools
+				const mcpTools = candidateTools.filter(tool => tool.tags.includes('mcp'));
+				const otherTools = candidateTools.filter(tool => !tool.tags.includes('mcp'));
+
+				// Calculate the maximum number of MCP tools we can keep based on the other tools count
+				const maxMcpTools = Math.max(0, 128 - otherTools.length);
```
marklicata, line 138: This design makes me nervous. We know LLC performance starts to degrade long before 128 tools and that 10 is most optimal. Why would we not try to semantic match to the top 10 tools where "10" is a configurable value? Would that be a follow on experiment?

digitarald: Doesn't have to be hard-coded, just meant to discuss the approach. The limit is currently set in https://github.com/microsoft/vscode-copilot-chat/blob/fc85ee93fa3ed01469d656f056885e633262abee/src/platform/configuration/common/configurationService.ts#L502 , as GH doesn't return the current model's tool limit capabilities.

src/extension/tools/vscode-node/toolsService.ts
digitarald: See https://github.com/microsoft/vscode-copilot-chat/pull/101#discussion_r2192992881

package.json
digitarald: Yes, it would run as a/b experiment but default off.

digitarald:
cc @connor4312 for input on the approach.

src/extension/tools/vscode-node/toolsService.ts
```diff
@@ -113,5 +117,46 @@ export class ToolsService extends BaseToolsService {
...
+					semanticallyFilteredMcpTools.some(semanticTool => semanticTool.name === tool.name)
+				);
+				const semanticTime = performance.now() - start;
+
+				// Combine filtered MCP tools with all other tools (non-MCP tools are not semantically filtered)
+				candidateTools = [...filteredMcpTools, ...otherTools];
+
+				// Log removed tools
+				const removedTools = mcpTools.filter(tool => !filteredMcpTools.some(filteredTool => filteredTool.name === tool.name));
+				console.log(`Removed ${removedTools.length} tools after semantic filtering (${semanticTime.toFixed(2)} ms): ${removedTools.map(tool => tool.name).join(', ')}`);
```
connor4312, line 154: Looks cool overall, we should have telemetry for this data

src/extension/tools/vscode-node/toolsService.ts
```diff
@@ -113,5 +117,46 @@ export class ToolsService extends BaseToolsService {
...
+			} catch (error) {
+				this.logService.logger.error('Failed to apply semantic filtering, using explicitly filtered tools', error);
+			}
+		}
+
+		return candidateTools;
```
connor4312, line 160: We should indicate to the user somehow that their tools were filtered. That doesn't necessarily need to be done here, I would be fine if we have a ⚠️ on the tool picker indicating filtering is going to happen in the core side

src/extension/tools/vscode-node/toolsService.ts
```diff
@@ -113,5 +117,46 @@ export class ToolsService extends BaseToolsService {
 
 			return false;
 		});
+
+		// Extract semantic query from the request if available
+		let semanticQuery: string | undefined;
+		if (request.prompt) {
+			// Use the user's prompt as the semantic query
+			semanticQuery = request.prompt;
```
connor4312, line 125: I'm not sure the prompt has enough knowledge to do this off the bat. Imo the better data source for this is after the first LM turn where it'll usually plan out what it's going to do. It does mean we'd throw out the first cache breakpoint though, cc @roblourens

digitarald: Also misses the conversation and open files as context; which might carry a lot of meaning.

joshfree:
@digitarald does this approach work with context-sensitive prompting in a long multi-prompt session though? Eg the user says something like "ok but that again only in east us 2"

JeffreyRichter:
I would not get an embedding vector that includes the input Schema or the tool name as these are cryptic and likely to match words in the prompt. 
I wouldn't have a cache size maximum - you need to put ALL the tools in the vector DB or the results will not consider some tools and you can't ignore some tools completely based on some being evicted.
I would always use the results of this and not just when >128 entries. 
A min score of .3 is worth considering - I'd like to know how well this works in practice.

PR #114, opened 2025-07-03
Enhance TypeScript tests and simulation workspace services

dfajardoCNE:
- Refactored function signature tests to use a shared TestSession for p12, improving test efficiency and readability.
- Increased global test timeout to 10 seconds in Vite configuration to accommodate longer-running TypeScript tests.
- Updated SimulationTerminal state to include an undefined shell property for better state management.

### DIFF
--- a/src/extension/typescriptContext/serverPlugin/src/node/test/simple.spec.ts
+++ b/src/extension/typescriptContext/serverPlugin/src/node/test/simple.spec.ts
@@ -196,6 +196,12 @@ suite('Method - Signature', () => {
 });
 
 suite('Function signature', () => {
+	let p12Session: testing.TestSession;
+
+	beforeAll(() => {
+		p12Session = create(path.join(root, 'p12'));
+	});
+
 	test('complete function signature types', () => {
 		const session: testing.TestSession = create(path.join(root, 'p7'));
 		const expected: testing.ExpectedCodeSnippet[] = [{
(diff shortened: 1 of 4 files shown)


src/extension/typescriptContext/serverPlugin/src/node/test/simple.spec.ts
```diff
@@ -222,51 +228,47 @@ suite('Function signature', () => {
 
 		const context = computeContext(session, path.join(root, 'p7/source/f2.ts'), { line: 6, character: 0 }, ContextKind.Snippet);
 		assertContextItems(context, expected);
-	});
+	}, 10000);
```
dbaeumer, line 231: Why do you think this is necessary. Tests should usually be fast and bumping that to 10 seconds is IMO not good since it will cover up slow tests./

vite.config.ts
```diff
@@ -21,6 +21,7 @@ export default defineConfig(({ mode }) => ({
 		include: ['**/*.spec.ts', '**/*.spec.tsx'],
 		exclude,
 		env: loadEnv(mode, process.cwd(), ''),
+		testTimeout: 10000, // Increase global timeout to 10 seconds for TypeScript tests
```
dbaeumer, line 24: Same here.

src/platform/test/node/simulationWorkspaceServices.ts
```diff
@@ -844,7 +844,7 @@ class SimulationTerminal extends Disposable implements vscode.Terminal {
 	) {
 		super();
 		this.name = creationOptions.name ?? '';
-		this.state = { isInteractedWith: false };
+		this.state = { isInteractedWith: false, shell: undefined };
```
dbaeumer, line 847: Can you explain why you think this is necessary?

dbaeumer:
I am closing the PR due to inactivity. Please ping if you want to continue on this and I will reopen it.

PR #125, opened 2025-07-04
qualify name of byi models with provider

aeschli:
For https://github.com/microsoft/vscode/issues/253131


aeschli:
Retired in favor of https://github.com/microsoft/vscode/issues/253131

PR #145, opened 2025-07-07
add tests for `TerminalAndTaskState` PromptElement

meganrogge:
fix [#253097](https://github.com/microsoft/vscode/issues/253097)


src/extension/prompts/node/agent/test/agentPrompt.spec.tsx
```diff
@@ -233,4 +234,179 @@ suite('AgentPrompt', () => {
...
+			assert(output.includes('npm: build'));
+			assert(output.includes('npm: watch'));
+			assert(output.includes('No active terminals found.'));
+		});
+		test('Copilot terminals and no active tasks', async () => {
+
+			const tasksService: any = {};
+			const terminalService: any = {};
+
+			const uri = URI.from({ path: 'workspace:/', scheme: 'file' });
```
roblourens: `workspace:/` doesn't look like a path, is that right?

src/extension/prompts/node/agent/test/agentPrompt.spec.tsx
```diff
@@ -233,4 +234,179 @@ suite('AgentPrompt', () => {
...
+				} else if (term.id === '2') {
+					return { commandLine: 'npm test', cwd: '/workspace', exitCode: 1 };
+				}
+				return undefined;
+			};
+
+			const prompt = new TerminalAndTaskStatePromptElement({}, tasksService, terminalService);
+			const rendered = await prompt.render();
+			const output = typeof rendered === 'string' ? rendered : JSON.stringify(rendered) ?? '';
+			assert(output.includes('Active Tasks:'));
```
roblourens: This is fine but the snapshots that I use in the other tests in this file are a really nice way to validate the whole output, that it's in the right order and formatting, and doesn't both include this string and `No active tasks found` and things like that, you might want to look at that if you aren't familiar with them, I've find them very useful.

src/extension/prompts/node/agent/test/agentPrompt.spec.tsx
```diff
@@ -233,4 +234,179 @@ suite('AgentPrompt', () => {
...
+			terminalService.getLastCommandForTerminal = (term: { id: string }) => {
+				if (term.id === '1') {
+					return { commandLine: 'npm run build', cwd: '/workspace', exitCode: 0 };
+				} else if (term.id === '2') {
+					return { commandLine: 'npm test', cwd: '/workspace', exitCode: 1 };
+				}
+				return undefined;
+			};
+
+			const prompt = new TerminalAndTaskStatePromptElement({}, tasksService, terminalService);
```
roblourens: You might want to have tests that just render the whole agent prompt, like the other tests in this file, because that lets you test that this element is actually used in the agent prompt and gets the right props passed in. Either is fine.

roblourens (commented):
If you want to test `TerminalAndTaskStatePromptElement` on its own, could you move this to its own test file so it stays focused?

meganrogge:
In the middle of other stuff ATM, so went with just a separate test file.

PR #153, opened 2025-07-08
Sample TypeScript context provider telemetry

dbaeumer:


src/extension/typescriptContext/vscode-node/languageContextService.ts
```diff
@@ -366,34 +366,46 @@ class TelemetrySender {
...
-			{
-				requestId: context.requestId,
-				source: context.source ?? KnownSources.unknown,
-				originalRequestId: originalRequestId
-			},
-			{
-				numberOfItems: numberOfItems
-			}
-		);
+		const sampleTelemetry = context.sampleTelemetry ?? 1;
```
alexr00: @dbaeumer if `context.sampleTelemetry` is `0`, then I think this will evaluate to `0`. Is that intended?

alexr00 (approved):

PR #165, opened 2025-07-08
Add metadata headers to ado requests

mjbvz:
Useful for tracking if the request was from stable/insiders


Frulfump:
I always think about ADO.NET when I see ADO but this is in reference to Azure DevOps (AzDO) right?

PR #166, opened 2025-07-08
monitor inactive or idol task status, poll for info

meganrogge:
fixes https://github.com/microsoft/vscode/issues/254667
fixes https://github.com/microsoft/vscode/issues/254671

This fixes a bug where the agent runs a task and it fails, but reports that the task is running in the background successfully.

Previously, the agent has only been aware of active tasks. We're now including information about inactive tasks too.

This makes use of our new `taskExecution.terminal` API to ensure we're mapping the task to the right terminal. We have to map a `taskExecution` to a terminal before it is removed from `taskExecutions` in order to be able to track the output for that task.

We poll until a task is idol or inactive. When that happens, we fetch the output and return that information to the agent and user.

Screenshot 2025-07-09 at 4 25 10 PM

Screenshot 2025-07-09 at 4 25 05 PM
0-cdf44005e666" />

Screenshot 2025-07-09 at 4 58 58 PM

### DIFF
--- a/src/extension/prompts/node/agent/test/terminalAndTaskPrompt.spec.tsx
+++ b/src/extension/prompts/node/agent/test/terminalAndTaskPrompt.spec.tsx
@@ -201,7 +201,7 @@ suite('TerminalAndTaskStatePromptElement', () => {
 
 		// Convert rendered output to string for assertions
 		const output = typeof rendered === 'string' ? rendered : JSON.stringify(rendered) ?? '';
-		assert(output.includes('No active tasks found.'));
+		assert(output.includes('No tasks found.'));
 		assert(output.includes('Terminal 1'));
 		assert(output.includes('Terminal 2'));
 		assert(output.includes('npm run build'));
@@ -228,6 +228,6 @@ suite('TerminalAndTaskStatePromptElement', () => {
 		const rendered = await prompt.render();
 
 		const output = typeof rendered === 'string' ? rendered : JSON.stringify(rendered) ?? '';
-		assert(output.includes('No active tasks or Copilot terminals found.'));
+		assert(output.includes('No tasks or Copilot terminals found.'));
 	});
 });


Frulfump:
Title could be better (and description ofc). I think it's referring to this issue? https://github.com/microsoft/vscode/issues/254671

src/platform/tasks/vscode/tasksService.ts
```diff
@@ -42,6 +57,19 @@ export class TasksService implements ITasksService {
 		return existingTasks.some(t => this.matchesTask(t, def));
 	}
 
+	/**
+	 * This is needed because when tasks exit, they're removed from the taskExecutions, but we might want to review the output of the task
```
meganrogge, line 71: here's how that keyedTaskId is created 
https://github.com/microsoft/vscode/blob/0748e889c18201cf0040c9022e5f87d4aaabf249/src/vs/workbench/contrib/tasks/common/tasks.ts#L1296-L1318

src/platform/tasks/vscode/tasksService.ts
```diff
@@ -42,6 +57,44 @@ export class TasksService implements ITasksService {
...
+			if ((taskDefinition.type === key.type &&
+				(!key.label || taskDefinition.label === key.label) &&
+				(!key.script || taskDefinition.script === key.script) &&
+				(!key.command || taskDefinition.command === key.command))) {
+				return terminal;
+			}
+		}
+	}
+
+	getTaskId(taskDefinition: vscode.TaskDefinition): string | undefined {
```
meganrogge: this comes from here https://github.com/microsoft/vscode/blob/0748e889c18201cf0040c9022e5f87d4aaabf249/src/vs/workbench/contrib/tasks/common/tasks.ts#L1296-L1317

src/platform/tasks/vscode/tasksService.ts
```diff
@@ -19,14 +19,29 @@ import { IWorkspaceService } from '../../workspace/common/workspaceService';
...
-	) { }
+	) {
+		super();
+		this.add(vscode.tasks.onDidStartTask(e => {
+			const terminal: vscode.Terminal | undefined = (e.execution as any).terminal;
+			if (!terminal) {
+				return;
+			}
+			this.taskDefinitionToTerminal.set(e.execution.task.definition, terminal);
+			this.add(vscode.window.onDidCloseTerminal(closedTerminal => {
```
connor4312: This leaks a listener every time a task gets started. Should make sure to clear this listener when appropriate

src/extension/tools/node/runTaskTool.tsx
```diff
@@ -55,6 +83,22 @@ class RunTaskTool implements vscode.LanguageModelTool<IRunTaskToolInput> {
 		return new LanguageModelToolResult([new LanguageModelTextPart(output)]);
 	}
 
+	private async _evaluateOutputForErrors(output: string, token: CancellationToken): Promise<string> {
+		const endpoint = await this._endpointProvider.getChatEndpoint('gpt-4o-mini');
+
+		const fetchResult = await endpoint.makeChatRequest(
+			'taskOutputEvaluation',
+			[{ role: ChatRole.System, content: [{ type: ChatCompletionContentPartKind.Text, text: `Review this output to determine if the task exited or if there are errors ${output}. If it has exited, explain why.` }] }],
```
connor4312: `output` should be in a user message

package.json
```diff
@@ -128,7 +128,8 @@
 		"taskProblemMatcherStatus",
 		"contribLanguageModelToolSets",
 		"textDocumentChangeReason",
-		"resolvers"
+		"resolvers",
+		"taskExecutionTerminal"
```
Tyriar, line 132: You'll need to update the vscode engine version also, otherwise this could get released and updated in older versions and throw runtime exceptions.

src/extension/prompts/node/agent/test/terminalAndTaskPrompt.spec.tsx
```diff
@@ -67,7 +73,14 @@ suite('TerminalAndTaskStatePromptElement', () => {
...
 		const terminalService: any = {};
-
+		tasksService.getTerminalForTask = (task: any) => {
+			if (task.command === 'build') {
+				return { name: 'Terminal 1', id: '1' };
+			} else if (task.command === 'watch') {
+				return { name: 'Terminal 2', id: '2' };
+			}
+			return undefined;
+		};
```
Tyriar: This is repeated a few times. Could refactor to share the impl

src/extension/tools/node/runTaskTool.tsx
```diff
@@ -39,6 +44,33 @@ class RunTaskTool implements vscode.LanguageModelTool<IRunTaskToolInput> {
...
 		const result: TaskResult = task ? await this.tasksService.executeTask(task, token, workspaceFolder) : { status: TaskStatus.Error, error: new Error('Task not found') };
+
+		// Waits a maximum of 10 seconds
+		// If it times out, just says the task has started
+		const checkIntervals = [1000, 2000, 3000, 4000];
+
+		if (task) {
+			let terminal: vscode.Terminal | undefined;
+			for (const interval of checkIntervals) {
+				await new Promise(resolve => setTimeout(resolve, interval));
```
Tyriar: We have the `timeout` helper in this repo

src/extension/tools/node/runTaskTool.tsx
```diff
@@ -39,6 +44,33 @@ class RunTaskTool implements vscode.LanguageModelTool<IRunTaskToolInput> {
...
+				if (!terminal) {
+					terminal = this.tasksService.getTerminalForTask(task);
+					if (!terminal) {
+						continue;
+					}
+				}
+				const buffer = this.terminalService.getBufferForTerminal(terminal, 16000);
+				const inactive = !this.tasksService.isTaskActive(task);
```
Tyriar, line 65: Does active here mean the task is currently running? Don't we also want to evaluate the output for active watch tasks, say a watch task without a problem matcher/begin/end patterns?

We probably want to do something very similar here:

https://github.com/microsoft/vscode-copilot-chat/blob/8f585e3680238be4ac371eb82771636be7bfdfef/src/extension/tools/node/runInTerminalTool.tsx#L91-L135

Should try share code when that happens.

src/platform/tasks/vscode/tasksService.ts
```diff
@@ -19,14 +19,39 @@ import { IWorkspaceService } from '../../workspace/common/workspaceService';
 import { ITasksService, TaskResult, TaskStatus } from '../common/tasksService';
 
 
-export class TasksService implements ITasksService {
+export class TasksService extends DisposableStore implements ITasksService {
 	_serviceBrand: undefined;
 
+	private taskDefinitionToTerminal: Map<vscode.TaskDefinition, vscode.Terminal> = new Map();
```
Tyriar: `latestTerminalForTaskDefinition`?

src/platform/tasks/vscode/tasksService.ts
Tyriar: Could add a comment pointing there as the source for future reference

src/platform/tasks/vscode/tasksService.ts
```diff
@@ -42,6 +67,44 @@ export class TasksService implements ITasksService {
...
+			if ((taskDefinition.type === key.type &&
+				(!key.label || taskDefinition.label === key.label) &&
+				(!key.script || taskDefinition.script === key.script) &&
+				(!key.command || taskDefinition.command === key.command))) {
+				return terminal;
+			}
+		}
+	}
+
+	getTaskId(taskDefinition: vscode.TaskDefinition): string | undefined {
```
Tyriar: Is this used? Couldn't see references outside of self recursive

Tyriar (dismissed):

src/extension/tools/node/runTaskTool.tsx
meganrogge: I think we only want to evaluate once the active task becomes idol, see what I just pushed.

meganrogge: If it takes longer than 10 seconds though, we don't want to do so

meganrogge: i'm going to defer adding telemetry

meganrogge: otherwise, for a watch task, it could be hanging for endless time

meganrogge: https://github.com/microsoft/vscode/issues/254980

DonJayamanne (approved):

PR #184, opened 2025-07-09
Fix missing filterReason in telemetry event when the finishReason is known to be content filtering by snippy

jwangxx:


jwangxx:
actually, this is the wrong approach. we need to add the actual filter reason on CAPI side. closing out

PR #208, opened 2025-07-10
list all openrouter models without category

danilofalcao:


isidorn:
@danilofalcao why do you think we should list all models? I worry that most of these models will not work well in VS Code?

danilofalcao:
> @danilofalcao why do you think we should list all models? I worry that most of these models will not work well in VS Code?

@isidorn unfortunately not all models on openrouter have this tag programming set like grok-4, devstral, codestral, etc. and they get excluded from the current fetch

ping-Toven:
it might make sense to filter using our tool calling param: https://openrouter.ai/api/v1/models?supported_parameters=tools

isidorn:
I like @ping-Toven suggestion.
@danilofalcao can you update the PR if this suggestion makes sense to you (we care about models that support tool calling)

danilofalcao:
> I like @ping-Toven suggestion. @danilofalcao can you update the PR if this suggestion makes sense to you (we care about models that support tool calling)

all good using @ping-Toven's suggestion

isidorn:
Did you get a chance to test this out?

danilofalcao:
> ?supported_parameters=tools

yep, that worked great listing all models with tool support

felipesssantos:
> > ?supported_parameters=tools
> 
> yep, that worked great listing all models with tool support

same here

TylerLeonhardt (approved):

isidorn:
Thank you for your contribution 🙏  Merging in.

rthidden:
Does this update keep us from being to add models that are available on OpenRouter but not part of the list in GitHub Copilot?

icegull:
Before upgrading, I can see useful free openrouter API like: 
1. qwen/qwen3-coder:free
2. deepseek/deepseek-r1-0528:free
3. deepseek/deepseek-r1:free

Now they are all gone..... I can't see them in openrouter anymore...
Do you know why? 


image

isidorn:
Because we only want to show the models that support tool calling - since that is the agentic future of the experience.

Do any of the models you are no longer seeing actually support tool calling?

icegull:
> Because we only want to show the models that support tool calling - since that is the agentic future of the experience.
> 
> Do any of the models you are no longer seeing actually support tool calling?

I don't know. But I can use those free models in chat and edit without any problem. Why we disabled them?

danilofalcao:
image

i'm still able to see free models. qwen3-coder isn't free anymore on openrouter. this is not a vscode/extension issue.

PR #216, opened 2025-07-11
Scaffold a chat lib npm module

alexdima:
Adds `npm run extract-chat-lib` which prepares at `/tmp-chat-lib` a npm package containing .ts source code that could be consumed as a library:
* the script that extracts the code is at `script/build/extractChatLib.ts` and it traverses the deps of the `main.ts` entrypoint and some other useful entrypoints. Once it extracts the source code, tsc is invoked on the extracted code.
* the npm package has a `main.ts` entry point (located at `src/lib/node/chat-lib-main.ts`)
* the idea is that the `main.ts` file would only export types that are written in the `main.ts` file itself. This would allow us to track the API of the npm package and offer semver semantics for the API. This can be done by creating a facade-style API where options are passed in an a live object that can receive file change events, cursor change events, etc. and also be asked for next edits. The creation options could offer hooks for configuring how to make http requests, how to report telemetry, what configuration options to use, etc.
* the rest of the files are placed at `_internal`. This allows a potential consumer to use them, but they would not have any semver guarantees and they can be changed at any time.

IMHO good next steps:
* this now produces a package which compiles, but which will throw as soon as someone tries to use it due to missing services.
* explore if a package in this shape could be useful/adopted or if it presents challenges
* one-by-one, add the tens of services necessary to make instantiating the code possible
* add necessary options / API to `main.ts`.
* add CI actions to make sure this doesn't get broken
* add builds to allow for publishing this somewhere at `@vscode/...`

cc @devm33 @chrmarti


alexdima:
Superseded by https://github.com/microsoft/vscode-copilot-chat/pull/807

PR #222, opened 2025-07-11
Optimistic Next Edit Suggestions

avarayr:
## Summary

  This PR implements optimistic prefetching for next edit suggestions, eliminating delay between accepting an edit and receiving the next suggestion by assuming that user will keep pressing tab to accept sequential edits.

  ## Problem

  Currently, when users accept an inline edit suggestion, there's a noticeable delay
  before the next suggestion appears. This breaks the flow of rapid sequential edits.

  ## Solution

  Implement predictive fetching that anticipates user behavior based on the "objects in
   motion stay in motion" principle. When a NES is shown, the system
  immediately fetches the next edit in the background, assuming the user will continue
  accepting suggestions.

  All existing tests continue to pass.

  ## Performance

  - Reduces next edit latency from ~300ms to <10ms for predicted cases
  - No impact on initial suggestion performance
  - Minimal memory overhead with automatic cleanup

## Discussion

Optimistic NES will multiply the number of NES requests by 3, which may affect the rate limit/user allowance.

### DIFF
--- a/src/extension/inlineEdits/node/nextEditProvider.ts
+++ b/src/extension/inlineEdits/node/nextEditProvider.ts
@@ -5,6 +5,7 @@
 
 import type * as vscode from 'vscode';
 import { ConfigKey, IConfigurationService } from '../../../platform/configuration/common/configurationService';
+import { IGitExtensionService } from '../../../platform/git/common/gitExtensionService';
 import { DocumentId } from '../../../platform/inlineEdits/common/dataTypes/documentId';
 import { RootedEdit } from '../../../platform/inlineEdits/common/dataTypes/edit';
 import { RootedLineEdit } from '../../../platform/inlineEdits/common/dataTypes/rootedLineEdit';
@@ -20,9 +21,9 @@ import { IExperimentationService } from '../../../platform/telemetry/common/null
 import { Result } from '../../../util/common/result';
 import { createTracer, ITracer } from '../../../util/common/tracing';
 import { assert } from '../../../util/vs/base/common/assert';
-import { DeferredPromise, timeout, TimeoutTimer } from '../../../util/vs/base/common/async';
+import { DeferredPromise, TimeoutTimer } from '../../../util/vs/base/common/async';
 import { CachedFunction } from '../../../util/vs/base/common/cache';
-import { CancellationToken } from '../../../util/vs/base/common/cancellation';
+import { CancellationToken, CancellationTokenSource } from '../../../util/vs/base/common/cancellation';
 import { BugIndicatingError } from '../../../util/vs/base/common/errors';
 import { Disposable, DisposableStore, IDisposable, toDisposable } from '../../../util/vs/base/common/lifecycle';
 import { LRUCache } from '../../../util/vs/base/common/map';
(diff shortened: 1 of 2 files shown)


avarayr:
hey @ulugbekna, i'd appreciate your [& other code owners] insight on this. 

There are some kinks to iron out, but in my testing, this massively improves the UX of NES.
This is one of the things that Cursor got right, hopefully we can bridge the gap.

some points for discussion: 

Optimistic NES will multiply the number of NES requests by 3, which may affect the rate limit/user allowance.

chrmarti:
@avarayr Thanks for the PR! We'll take a detailed look. We are also working on additional approaches to improve the experience while keeping the number of requests in check. (E.g., https://github.com/microsoft/vscode/issues/255727 and https://github.com/microsoft/vscode/issues/255728.)

src/extension/inlineEdits/node/nextEditProvider.ts
```diff
@@ -630,11 +641,144 @@ export class NextEditProvider extends Disposable implements INextEditProvider<Ne
...
+		}
+	}
+
+	private async _startPrefetchChain(docId: DocumentId, documentState: StringText, depth: number): Promise<void> {
+		if (depth >= this._prefetchChainDepth) {
+			return;
+		}
+
+		// Create a unique key for this prefetch based on document state
+		const prefetchKey = `${docId}:${documentState.value.length}:${depth}`;
```
chrmarti, line 671: Since we only show one suggestion at a time, maybe there should be at most one active prefetch at a time? (Avoiding the need for a key here.)

src/extension/inlineEdits/node/nextEditProvider.ts
```diff
@@ -630,11 +641,144 @@ export class NextEditProvider extends Disposable implements INextEditProvider<Ne
...
+			this._activePrefetches.delete(prefetchKey);
+		}
+
+		const cancellationSource = new CancellationTokenSource();
+		this._activePrefetches.set(prefetchKey, cancellationSource);
+
+		try {
+
+			// Create a synthetic document that represents the future state
+			const syntheticDoc = this._workspace.getDocument(docId);
```
chrmarti, line 686: This isn't used later?

src/extension/inlineEdits/node/nextEditProvider.ts
```diff
@@ -630,11 +641,144 @@ export class NextEditProvider extends Disposable implements INextEditProvider<Ne
...
+		const nesConfigs: INesConfigs = {
+			isAsyncCompletions: this._configService.getExperimentBasedConfig(ConfigKey.Internal.InlineEditsAsyncCompletions, this._expService),
+			isRevisedCacheStrategy: this._configService.getExperimentBasedConfig(ConfigKey.Internal.InlineEditsRevisedCacheStrategy, this._expService),
+			isCacheTracksRejections: this._configService.getExperimentBasedConfig(ConfigKey.Internal.InlineEditsCacheTracksRejections, this._expService),
+			isRecentlyShownCacheEnabled: this._configService.getExperimentBasedConfig(ConfigKey.Internal.InlineEditsRecentlyShownCacheEnabled, this._expService),
+		};
+
+		// Create a synthetic document wrapper that returns our future state
+		const syntheticDocWrapper = {
+			...doc,
```
chrmarti, line 745: Not sure this is safe to copy like that, will need to take closer look.

src/extension/inlineEdits/node/nextEditProvider.ts
```diff
@@ -630,11 +641,144 @@ export class NextEditProvider extends Disposable implements INextEditProvider<Ne
...
 	}
 
 	public handleAcceptance(docId: DocumentId, suggestion: NextEditResult) {
 		this.runSnippy(docId, suggestion);
 		this._statelessNextEditProvider.handleAcceptance?.();
+
+		// Optimistic fetching already triggered in handleShown
 	}
 
 	public handleRejection(docId: DocumentId, suggestion: NextEditResult) {
```
chrmarti, line 784: Should the prefetch be canceled when the current suggestion is rejected? Similarly: What if the current suggestion is ignored (callback below).

chrmarti (changes requested):
@avarayr Thanks again for the PR! I have left a few questions. Could you rebase/merge it with the latest from `main`?

Could you also add a setting to enable/disable it? We would keep this off by default at first, so we can get it well tested and estimate the number of additional LLM calls.

Dev-KPX-Abellanosa (approved):

PR #227, opened 2025-07-12
Use lineNumberStyle for FileVariable

roblourens:
FYI @connor4312 in case this causes edit issues. Also I don't know where to add a similar telemetry check in applyPatch for this. Looks like I didn't add it for the other summary format.

image


connor4312 (dismissed):
lgtm, some vitest failures though

dbaeumer (approved):

PR #228, opened 2025-07-12
Fix BYOK Quota exceeded caused by Intent Detection with Copilot model

fvclaus:
Fixes https://github.com/microsoft/vscode/issues/251944

### Issue

Currently, the intent detection is hard-coded to use the copilot gpt-4o-mini model. This is an issue when you have exceeded your chat quota.

This will result in either one of the two errors

image

```
Sorry, your request failed. Reason: Canceled
```

This is caused after the BYOK Chat Request returns with a 200. The code will reset the copilot token, because it doesn't check that it originated from the Copilot API

```
if (response.status === 200 && authenticationService.copilotToken?.isFreeUser && authenticationService.copilotToken?.isChatQuotaExceeded) {
		authenticationService.resetCopilotToken();
}
```
This also causes unnecessary network requests

The second error I have seen (on possibly older versions of the plugin) is the Quota exceeded error

image

### Solution

I have tried to fix it in a way that is extensible later on
- Decouple the intent models from the regular models (although they are currently the same, this leaves the option to change it later)
- Design the API of the new service in a way so that the intent detection model can be set per language model (currently it is only one intent detection model for all models)
- The simplest fix would have been to use the current model of the Chat Request for intent detection, but that is potentially wasteful with expensive models and this isn't transparent to the user that another API request is sent behind the scenes
- Not mix the BYOK Code with the other code, so I made the BYOK code register its models on startup at the new service
- For all copilot models gpt-4o-mini is still used. I incorporated this into the message of the quick pick for users who want to have the same behaviour as before
- I didn't want to mix intent detection with a copilot model with a chat request with a BYOK model, because this is makes it more complicate and hard to reason about the different states of the system

With this MR there is an error message the first time a BYOK model is used
image

The button opens a quick pick 
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ coverage/
 test/simulation/cache/_base.sqlite
 
 test/aml/out
+
+# From the Booksmarks vscode extension
+.vscode/bookmarks.json
--- a/package.json
+++ b/package.json
@@ -2121,6 +2121,13 @@
 				"enablement": "github.copilot.byokEnabled"
 			},
 			{
+				"command": "github.copilot.chat.manageIntentDetectionModel",
+				"title": "Choose Intent Detection Model...",
+				"icon": "$(settings-gear)",
+				"category": "GitHub Copilot",
+				"enablement": "github.copilot.byokEnabled"
+			},
+			{
 				"command": "github.copilot.chat.debug.showElements",
 				"title": "Show Rendered Elements"
 			},
--- a/src/extension/byok/common/byokProvider.ts
+++ b/src/extension/byok/common/byokProvider.ts
@@ -110,6 +110,7 @@ export function resolveModelInfo(modelId: string, providerName: string, knownMod
 	return {
 		id: modelId,
 		name: modelName,
+		vendor: 'copilot-byok',
 		version: '1.0.0',
 		capabilities: {
 			type: 'chat',
--- a/src/extension/byok/common/byokProvider.ts
+++ b/src/extension/byok/common/byokProvider.ts
@@ -6,6 +6,7 @@ import type { ChatResponseProviderMetadata, Disposable } from 'vscode';
 import { CopilotToken } from '../../../platform/authentication/common/copilotToken';
 import { ICAPIClientService } from '../../../platform/endpoint/common/capiClient';
 import { IChatModelInformation } from '../../../platform/endpoint/common/endpointProvider';
+import { IChatEndpoint } from '../../../platform/networking/common/networking';
 import { TokenizerType } from '../../../util/common/tokenizer';
 import { localize } from '../../../util/vs/nls';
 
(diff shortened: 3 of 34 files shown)


fvclaus:
I have moved the new service to `common` to fix the test. `vscode` dependencies are imported dynamically. Maybe there is a better way. The tests run fine in my repo now.

fvclaus:
I have released the code. You can download it from here: https://github.com/fvclaus/vscode-copilot-chat/releases/tag/v0.29.0-byok-fix-v1

To install it, go to settings, click on the three dots and select "Install from VSIX..."
image

insilications:
This is because the Copilot Chat base workflow is still closely coupled with the Copilot API. It was like that before the open source release. Hopefully, Microsoft developers with intimate knowledge of the source code will decouple the intent inference functionality from the Copilot API as they appear to be doing with the NES functionality.

fvclaus:
I am surprised not more people are running into this issue. It seems that not many are using BYOK after all.

There has been a change since the last release. The BYOK models retrieved from the https://main.vscode-cdn.net/extensions/copilotChat.json changed. The OpenAI models don't have the `models/` prefix anymore. And the `Reason: Cancelled` issue started appearing again, because I used the `models/` prefix to determine what is BYOK and what isn't.

I refactored the code and introduced `vendor` property in `ChatEndpoint` to be able to distinguish what is provided by Copilot and what isn't. 

Is there something I can do to get feedback on the MR to move forward?

fvclaus:
For anybody else blocked by this as well, I have rebased my changes on the 0.29.1 version and released it as https://github.com/fvclaus/vscode-copilot-chat/releases/tag/v0.29.1-byok-fix-v4

To install you need to remove Copilot (and it will also uninstall Copilot Chat) and then install from vsix. This will pull in Copilot again.

What will work with the fix when the quota is exceeded
- Inline chat
- Chat in Ask mode

What doesn't seem to work
- Function calling (so no Agent Mode or Edit Mode)
- Code Completions

The function calling I can't reproduce currently because my quota just reset. It was reported to me by a colleague who is also running this fix.

I am also not using Agent mode. I use Roo instead so this won't be a priority for me, but I will look into it when I get some logs or my quota runs out again.

fvclaus:
@insilications BYOK is available since the March release of vscode. I am surprised that there is still such a tight coupling between many features and the internal copilot models. It is as if they haven't tested this at all, which I don't believe. What I also find surprising is that this specific issue started appearing around the time that copilot was open-sourced. I have used Copilot BYOK since April and haven't encountered any issues since around 3 weeks. Maybe I was never over the quota (I never really paid attention), but I doubt it since I am using it almost every day. Maybe it is only a coincidence and the code that broke it coincided with the open sourcing of this extension.

insilications:
> @insilications BYOK is available since the March release of vscode. I am surprised that there is still such a tight coupling between many features and the internal copilot models. It is as if they haven't tested this at all, which I don't believe. What I also find surprising is that this specific issue started appearing around the time that copilot was open-sourced. I have used Copilot BYOK since April and haven't encountered any issues since around 3 weeks. Maybe I was never over the quota (I never really paid attention), but I doubt it since I am using it almost every day. Maybe it is only a coincidence and the code that broke it coincided with the open sourcing of this extension.

I agree with you. It's kinda strange. I had already complained about it to them. Overall, the BYOK experience leaves a little to be desired. It is not good.

lramos15:
>  Overall, the BYOK experience leaves a little to be desired. It is not good.

It's not the primary target for Copilot, but instead a power user tinker feature that we're continuously polishing.


Anyways, we've pushed a server side fix which should fix the improper counting when using BYOK models and a free SKU

PR #234, opened 2025-07-14
Attaching a folder should only reference it

benibenj:
Originally, attaching a folder would lead to a semantic search scoped to the attached folders. This lead to issue when moving to the agentic flow. As the agent is now able to perform searches in files/folders, it should be enough to only mention which folder the user attached and let the agent find relevant code snippets if needed.

fixes https://github.com/microsoft/vscode/issues/254555

### DIFF
--- a/src/extension/prompts/node/panel/chatVariables.tsx
+++ b/src/extension/prompts/node/panel/chatVariables.tsx
@@ -7,12 +7,16 @@ import { BasePromptElementProps, PromptElement, PromptElementProps, PromptPiece,
 import type { Diagnostic, DiagnosticSeverity, LanguageModelToolInformation } from 'vscode';
 import { ChatFetchResponseType, ChatLocation } from '../../../../platform/chat/common/commonTypes';
 import { IEndpointProvider } from '../../../../platform/endpoint/common/endpointProvider';
+import { IFileSystemService } from '../../../../platform/filesystem/common/fileSystemService';
+import { FileType } from '../../../../platform/filesystem/common/fileTypes';
 import { ILogService } from '../../../../platform/log/common/logService';
 import { ICopilotToolCall } from '../../../../platform/networking/common/fetch';
 import { IChatEndpoint } from '../../../../platform/networking/common/networking';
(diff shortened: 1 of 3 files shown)


src/extension/prompts/node/panel/chatVariables.tsx
```diff
@@ -225,6 +256,47 @@ function getDiagnosticCode(diagnostic: Diagnostic): string {
...
+		@IInstantiationService private readonly instantiationService: IInstantiationService,
+		@IPromptPathRepresentationService private readonly promptPathRepresentationService: IPromptPathRepresentationService,
+	) {
+		super(props);
+	}
+
+	override async prepare(sizing: PromptSizing): Promise<IFileTreeData | undefined> {
+		try {
+			return this.instantiationService.invokeFunction(accessor =>
+				workspaceVisualFileTree(accessor, this.props.folderUri, { maxLength: 2000, excludeDotFiles: false }, CancellationToken.None)
```
benibenj, line 278: Not sure how to pick an appropriate max length here

src/extension/prompts/node/panel/chatVariables.tsx
```diff
@@ -121,7 +139,7 @@ function asUserMessage(element: PromptElement, priority: number | undefined): Us
 }
 
 
-export async function renderChatVariables(chatVariables: ChatVariablesCollection, includeFilepathInCodeBlocks = true, omitReferences?: boolean, isAgent?: boolean): Promise<PromptElement[]> {
+export async function renderChatVariables(chatVariables: ChatVariablesCollection, fileSystemService: IFileSystemService, includeFilepathInCodeBlocks = true, omitReferences?: boolean, isAgent?: boolean): Promise<PromptElement[]> {
```
benibenj, line 142: Not sure how we can avoid passing the fileSystemService into here without making a larger refactoring

roblourens: Dunno why we use functions for some random things

roblourens (approved):

PR #242, opened 2025-07-14
Allow all SKUs to use BYOK

pierceboggan:
This PR removes a license check that was in place for "[bring your own key](https://code.visualstudio.com/docs/copilot/language-models#_bring-your-own-language-model-key)" which previously restricted use to Copilot Free and Individual SKUs.

This also shows a warning the first time you add a BYOK model in VS Code.

### DIFF
--- a/src/extension/byok/common/byokProvider.ts
+++ b/src/extension/byok/common/byokProvider.ts
@@ -134,10 +134,13 @@ export function resolveModelInfo(modelId: string, providerName: string, knownMod
  * Determines if Bring Your Own Key (BYOK) functionality is enabled for the current user.
  *
  * BYOK availability rules:
- * - Internal users: Always enabled
- * - Individual users: Always enabled (backward compatibility)
- * - Business/Enterprise users: Requires "Editor Preview Features" to be enabled by admin
- * - GitHub Enterprise Server: Not available
+ * - GitHub Enterprise Server: Not available (cloud endpoints required)
+ * - All cloud Copilot plans (internal, individual, business, enterprise): Enabled
+ *
+ * NOTE: we previously gated Business/Enterprise tenants behind the "Editor Preview Features" org
+ * policy. That restriction has been removed. We instead surface an in-product disclaimer when a
+ * user opens the Manage Models UI to make it clear that externally configured (BYOK) models are
+ * not covered by Copilot model quality, data handling, or compliance guarantees.
  *
  * @param copilotToken The user's Copilot token (without the actual token value)
  * @param capiClientService Service to check if running on GitHub Enterprise
@@ -146,25 +149,11 @@ export function resolveModelInfo(modelId: string, providerName: string, knownMod
 export function isBYOKEnabled(copilotToken: Omit<CopilotToken, "token">, capiClientService: ICAPIClientService): boolean {
 	const isGHE = capiClientService.dotcomAPIURL !== 'https://api.github.com';
 
-	// Not available on GitHub Enterprise
+	// Not available on GitHub Enterprise Server instances (cloud only)
 	if (isGHE) {
 		return false;
 	}
 
-	// Always allow for internal users
-	if (copilotToken.isInternal) {
-		return true;
-	}
-
-	// Allow for individual users without preview features gating (backward compatibility)
-	if (copilotToken.isIndividual) {
-		return true;
-	}
-
-	// For business and enterprise users, require Editor Preview Features to be enabled
-	if (copilotToken.copilotPlan === 'business' || copilotToken.copilotPlan === 'enterprise') {
-		return copilotToken.isEditorPreviewFeaturesEnabled();
-	}
-
-	return false;
+	// Enabled for all cloud Copilot users regardless of SKU or preview policy.
+	return true;
 }
--- a/src/extension/byok/vscode-node/byokContribution.ts
+++ b/src/extension/byok/vscode-node/byokContribution.ts
@@ -2,7 +2,7 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-import { commands, Disposable as VSCodeDisposable, window } from 'vscode';
+import { commands, Uri, Disposable as VSCodeDisposable, window } from 'vscode';
 import { IAuthenticationService } from '../../../platform/authentication/common/authentication';
 import { ConfigKey, IConfigurationService } from '../../../platform/configuration/common/configurationService';
 import { ICAPIClientService } from '../../../platform/endpoint/common/capiClient';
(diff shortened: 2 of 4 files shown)


src/extension/byok/common/byokProvider.ts
```diff
@@ -130,8 +130,41 @@ export function resolveModelInfo(modelId: string, providerName: string, knownMod
 	};
 }
 
+/**
```
isidorn, line 133: No need for comments if they were not there before. I know some devs prefer code with 0 comments.

src/extension/byok/common/byokProvider.ts
```diff
@@ -130,8 +130,41 @@ export function resolveModelInfo(modelId: string, providerName: string, knownMod
...
+		return true;
+	}
+
+	// Allow for individual users without preview features gating (backward compatibility)
+	if (copilotToken.isIndividual) {
+		return true;
+	}
+
+	// For business and enterprise users, require Editor Preview Features to be enabled
+	if (copilotToken.copilotPlan === 'business' || copilotToken.copilotPlan === 'enterprise') {
```
isidorn: No need to hard code 'business', 'enterprise'. There is probably some enum somewhere you can use.

src/extension/byok/common/byokProvider.ts
```diff
@@ -130,8 +130,41 @@ export function resolveModelInfo(modelId: string, providerName: string, knownMod
...
+		return copilotToken.isEditorPreviewFeaturesEnabled();
```
isidorn: 0 enterprises have preview features enabled. 
So I would not go down this route. For now I would just enable for insiders, and if GH wants a policy to disable once they deliver we can return that boolean. And I hope that is ENABLED by default.

isidorn:
Thanks for this PR. I left some comments.
I think the bigger part of the work is actually notifying the user and making it obvious that the BYOK models are not covered with guarantess we have for Copilot models. I do not know if that is a dialog, notification, or just makeing it super clear in the docs.

src/extension/byok/common/byokProvider.ts
```diff
@@ -130,8 +130,41 @@ export function resolveModelInfo(modelId: string, providerName: string, knownMod
...
+	if (isGHE) {
+		return false;
+	}
+
+	// Always allow for internal users
+	if (copilotToken.isInternal) {
+		return true;
+	}
+
+	// Allow for individual users without preview features gating (backward compatibility)
```
DanielRosenwasser:
```suggestion
	// Allow for individual users without preview feature-gating (backward compatibility)
```

src/extension/byok/common/byokProvider.ts
```diff
@@ -130,8 +130,41 @@ export function resolveModelInfo(modelId: string, providerName: string, knownMod
...
+ *
+ * BYOK availability rules:
+ * - Internal users: Always enabled
+ * - Individual users: Always enabled (backward compatibility)
+ * - Business/Enterprise users: Requires "Editor Preview Features" to be enabled by admin
+ * - GitHub Enterprise Server: Not available
+ *
+ * @param copilotToken The user's Copilot token (without the actual token value)
+ * @param capiClientService Service to check if running on GitHub Enterprise
+ * @returns true if BYOK should be enabled for this user
```
DanielRosenwasser, line 147:
```suggestion
 * @returns whether BYOK should be enabled for this user
```

src/extension/byok/common/byokProvider.ts
DanielRosenwasser: I actually don't think there is an enum - this value is currently defined as a string literal union:

https://github.com/microsoft/vscode-copilot-chat/blob/06a748132e121645fc95de67e4939afc5dff45a6/src/platform/authentication/common/copilotToken.ts#L93

DanielRosenwasser (commented):
Not opposed to the JSDoc comment but if you do end up removing it, just inline some of the details.

pierceboggan:
@isidorn OK, give it another look. Just cleaned up the logic more. :)

src/extension/byok/vscode-node/byokContribution.ts
```diff
@@ -126,6 +127,25 @@ export class BYOKContrib extends Disposable implements IExtensionContribution {
 	}
 
 	private async registerModelCommand() {
+		// One‑time disclaimer: make it clear that BYOK models are external and not covered by Copilot guarantees.
+		if (!this._byokDisclaimerShown) {
+			this._byokDisclaimerShown = true; // ensure we only show once per session
+			const detail = 'Models you configure here are provided by external services that you choose. Your prompts and code may be sent to those services and are subject to their terms. GitHub Copilot\'s data handling, compliance, and quality guarantees do NOT apply to BYOK models.';
```
isidorn, line 133: You might need to improve this string to be Localized. I am not 100% sure how to do it from this extension, but ask agent mode.

src/extension/byok/vscode-node/byokContribution.ts
```diff
@@ -126,6 +127,25 @@ export class BYOKContrib extends Disposable implements IExtensionContribution {
...
+			const learnMore = 'Learn More';
```
isidorn, line 134: Same for all these other strings. I think all user facing strings need a special syntax to be localized.

src/extension/byok/vscode-node/byokContribution.ts
```diff
@@ -126,6 +127,25 @@ export class BYOKContrib extends Disposable implements IExtensionContribution {
 	}
 
 	private async registerModelCommand() {
+		// One‑time disclaimer: make it clear that BYOK models are external and not covered by Copilot guarantees.
+		if (!this._byokDisclaimerShown) {
+			this._byokDisclaimerShown = true; // ensure we only show once per session
```
isidorn, line 132: How often is the registerModelCommand called?

If it is called only once the user adds a BYOK model - then I think this would work.

isidorn:
Left some comments. 

Did you get a chance to test this out? If yes, can you share a screenshot / gif so I see the flow to potentially provide more feedback.

nblog:
any progress?

nblog:
No progress at all? I can understand that certain organization administrators can disable this feature, but I simply cannot comprehend why it was designed so that no user in any organization can use it?

lramos15:
>  I simply cannot comprehend why it was designed so that no user in any organization can use it?

The GitHub services team has not had time to add a policy for enterprises to opt into this feature so for now we must leave it disabled as otherwise we would enable it and no company would be able to turn it off

pavele:
Hi all, what is the state of this problem, as the "manage models" functionality is still not available in both VS Code and the insider builds? We're using the business edition, and this prevents us from using model providers as HuggingFace or others. This has been forever now :(

maxious:
https://github.blog/changelog/2025-11-20-enterprise-bring-your-own-key-byok-for-github-copilot-is-now-in-public-preview/ is now released, can we please have local BYOK for Business/Enterprise users available

alex73630:
Hi everyone,

I wanted to know if there was an effort to push this feature out as I'd really like to try out some models that are not available in GitHub Copilot.
We decided to go for GitHub Copilot Teams at my company of two mostly for billing/invoicing management and it is really bothering us getting less features and having to wait months to get a feature even though we're paying double the price than Copilot Pro.

If having to deal with 2 invoices and paying half the price is required to get access to a feature hidden behind a single boolean check, well I guess I'll be happy to give less money.

Don't get me wrong, I understand that enterprises have security and compliance obligations. But please give us, small teams, a way to disable unwanted "safeguards".

PR #253, opened 2025-07-15
tools: improve string_replace and add healing (ft. Gemini)

connor4312:
Cherry-pick of https://github.com/microsoft/vscode-copilot-chat/pull/251. What this does:

- Updates the tool description of replace_string_in_file to be more explicit for Gemini (does not appear to have much impact either way for Sonnet)
- Updates edit instructions such that the model prefers replace_string_in_file when present (note: Sonnet sees replace_string_in_file as its only edit tool, so no effect there)
- Adds a "healing" step for replace_string_in_file, can be disabled via EXP.

### DIFF
--- a/package.json
+++ b/package.json
@@ -933,7 +933,7 @@
 				"name": "copilot_replaceString",
 				"toolReferenceName": "replaceString",
 				"displayName": "%copilot.tools.replaceString.name%",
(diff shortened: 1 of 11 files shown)


connor4312:
swebench is still running but looking good, looks like there will actually be a least a few percentage point uplift, 0-7% depending how many of the remaining cases pass. Will go ahead and merge this.

PR #263, opened 2025-07-16
fix task issue

meganrogge:
The `taskDefinition` on the `taskExecution.task` fired here is extremely paired down. The definition later has more properties. We need to compare only the important things on the definition. This fixes a few issues:

1) actually do the check for task definition equality
2) the task would be returned with a terminal just by matching `type` when it had an `id` but no match due to the above issue
2) validating the definitions actually match

https://github.com/microsoft/vscode-copilot-chat/blob/40379f6cb6908f396de3214a7857358265c22f94/src/platform/tasks/vscode/tasksService.ts#L34

Screenshot 2025-07-16 at 9 56 50 AM

Screenshot 2025-07-16 at 9 56 43 AM

### DIFF
--- a/src/platform/tasks/vscode/tasksService.ts
+++ b/src/platform/tasks/vscode/tasksService.ts
@@ -78,12 +78,13 @@ export class TasksService extends DisposableStore implements ITasksService {
 		for (const [key, terminal] of this.latestTerminalForTaskDefinition.entries()) {
 			if (key.id) {
 				// Only some task definitions have IDs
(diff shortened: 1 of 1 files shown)


meganrogge:
@Tyriar there's actually no repro rn because it's checking if type === type and returning for that extremely broad condition. This just makes it more solid.

For this case, it's literally returning a terminal just because `shell === shell`
Screenshot 2025-07-16 at 10 46 14 AM

PR #269, opened 2025-07-16
wait longer for tasks to become idle

meganrogge:
Wait up to 20 seconds with polling to get task output. I'm not sure what time amount is appropriate here 🤔 , but 20 seems like a reasonable amount. Should I increase it to like 120 seconds? I imagine stuff takes that long or longer on windows for example.

fixes https://github.com/microsoft/vscode/issues/256099

### DIFF
--- a/src/extension/tools/node/runTaskTool.tsx
+++ b/src/extension/tools/node/runTaskTool.tsx
@@ -54,8 +54,11 @@ class RunTaskTool implements vscode.LanguageModelTool<IRunTaskToolInput> {
 		const taskRunDurationMs = taskEndTime - taskStartTime;
 		let totalDurationMs: number | undefined;
 
-		// Start with 1000 to ensure the buffer has content, wait up to 20 seconds for the task to become idle
-		const checkIntervals = [1000, 100, 100, 100, 100, 100, 200, 200, 200, 200, 200, 200, 500, 500, 500, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000];
+		// Use a target end time and exponential backoff for polling, up to 20 seconds
+		const maxWaitMs = 20000;
+		const maxInterval = 2000;
+		let currentInterval = 500;
+		const endTime = Date.now() + maxWaitMs;
 
 		let pollStartTime: number | undefined;
 		let pollEndTime: number | undefined;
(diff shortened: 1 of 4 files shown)


src/extension/tools/node/runTaskTool.tsx
```diff
@@ -54,8 +54,8 @@ class RunTaskTool implements vscode.LanguageModelTool<IRunTaskToolInput> {
 		const taskRunDurationMs = taskEndTime - taskStartTime;
 		let totalDurationMs: number | undefined;
 
-		// Start with 500 to ensure the buffer has content
-		const checkIntervals = [1000, 100, 100, 100, 100, 100];
+		// Start with 1000 to ensure the buffer has content, wait up to 20 seconds for the task to become idle
+		const checkIntervals = [1000, 100, 100, 100, 100, 100, 200, 200, 200, 200, 200, 200, 500, 500, 500, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000];
```
Tyriar: After some time of the terminal idling and the LLM thinking it's still in progress, maybe we should ask the user if they want to continue to wait? 20 seconds seems fine if we take that approach.

Tyriar: Easier to understand approach would be to have target time (`20000`) and then just double the interval up to a maximum. Something like this:

```ts
const endTime = Date.now() + 20000;
let currentInterval = 500;
while (Date.now() < endTime) {
  currentInterval = Math.min(currentInterval * 2, 2000);
  await timeout(currentInterval);
}
```

That also makes it wait close to 20s, as opposed to 20 seconds plus all the request durations combined.

Tyriar (changes requested):

src/extension/tools/node/runTaskTool.tsx
Tyriar: And if we have the begin/end pattern we could bump this timeout up to something like 2 minutes as it should work, we just want to catch the cases where it's misconfigured. The longer we wait, the larger the interval should be to reduce server costs.

andysalerno: @Tyriar what do you think about allowing a longer timeout time for build tasks? Specifically, tasks that are `"kind": "build"` and `"isDefault": true` ? My original concern in the bug report is that Copilot is triggering my build task to check if there are any errors (as expected) but many builds are longer than 20 seconds.

Maybe not for this PR, the 20sec is already an improvement :D Just wondering what you expect to happen for build tasks or test tasks

meganrogge: our tasks in vscode-copilot and vscode do not have end patterns. I worry mandating that will cause this to not wait for many tasks

meganrogge: @andysalerno we're going to see if there's a way to ask for user confirmation and then continue polling for longer.

meganrogge:
Closing this because will handle when the task tool is moved over to core

PR #281, opened 2025-07-17
renable test with a bit more timeout

TylerLeonhardt:
fixes https://github.com/microsoft/vscode/issues/253598

### DIFF
--- a/src/platform/workbench/test/vscode-node/workbenchServiceImpl.test.ts
+++ b/src/platform/workbench/test/vscode-node/workbenchServiceImpl.test.ts
@@ -7,7 +7,7 @@ import * as assert from 'assert';
 import { WorkbenchServiceImpl } from '../../vscode/workbenchServiceImpt';
 
 suite('HostServiceImpl', () => {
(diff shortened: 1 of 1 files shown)


src/platform/workbench/test/vscode-node/workbenchServiceImpl.test.ts
```diff
@@ -7,15 +7,15 @@ import * as assert from 'assert';
...
+	test('getAllCommands', async () => {
 		const envService = new WorkbenchServiceImpl();
 		const commands = await envService.getAllCommands();
 		assert.ok(Array.isArray(commands));
 		assert.ok(commands.length > 0);
 		assert.ok(commands[0].label);
 		assert.ok(commands[0].command);
 		assert.ok(commands[0].keybinding);
-	});
+	}).timeout(10000); // Give this test a little bit more time due to potential delays in fetching commands
```
Tyriar, line 18: That's a very long timeout, we don't want any delays near that in unit tests which should run in milliseconds?

Tyriar (changes requested):

PR #303, opened 2025-07-18
adding turn index, conv id and message id where needed

24anisha:
First pass to add fields to existing telemetry events and send internal msft telemetry where restricted telemetry is already being sent.

Needs addressed:
- Include `conversationId` and `messageId` in fastApply/editOutcome events
- Send `engine.messages` events to internal Microsoft telemetry
- Add turn index to `engine.messages` events

Based on requirements in: [17219](https://github.com/microsoft/vscode-copilot/issues/17219)


lramos15 (approved):
🫡

src/extension/mcp/vscode-node/mcpToolCallingLoop.tsx
```diff
@@ -99,7 +99,8 @@ export class McpToolCallingLoop extends ToolCallingLoop<IMcpToolCallingLoopOptio
 			firstFetchCall,
 			{
 				messageId: randomUUID(),
-				messageSource: McpToolCallingLoop.ID
+				messageSource: McpToolCallingLoop.ID,
+				turnIndex: this.options.conversation.turns.length.toString(),
```
roblourens, line 103: You probably want to send it as a number, not a string?

PR #304, opened 2025-07-18
new notebook edit intent for inline notebook chat

amunger:
https://github.com/microsoft/vscode/issues/255807

### DIFF
--- a/src/extension/prompt/node/chatParticipantRequestHandler.ts
+++ b/src/extension/prompt/node/chatParticipantRequestHandler.ts
@@ -288,17 +288,17 @@ export class ChatParticipantRequestHandler {
 		if (!command?.intent && (this.location === ChatLocation.Editor || this.location === ChatLocation.Notebook)) { // TODO@jrieken do away with location specific code
 
 			let preferredIntent: Intent | undefined;
-			if (this.documentContext && this.request.attempt === 0 && history.length === 0) {
-				if (this.location === ChatLocation.Notebook) {
-					preferredIntent = Intent.notebookEditor;
-				} else if (this.documentContext.selection.isEmpty && this.documentContext.document.lineAt(this.documentContext.selection.start.line).text.trim() === '') {
+			if (this.documentContext && this.location === ChatLocation.Notebook) {
+				preferredIntent = Intent.notebookEditor;
+			} else if (this.documentContext && this.request.attempt === 0 && history.length === 0) {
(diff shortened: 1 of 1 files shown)


src/extension/intents/node/notebookEditorIntent.ts
```diff
@@ -0,0 +1,151 @@
...
+
+	public override async getAvailableTools(): Promise<vscode.LanguageModelToolInformation[]> {
+		return getTools(this.instantiationService, this.request);
+	}
+
+	public override buildPrompt(promptContext: IBuildPromptContext, progress: vscode.Progress<vscode.ChatResponseReferencePart | vscode.ChatResponseProgressPart>, token: vscode.CancellationToken): Promise<IBuildPromptResult> {
+		const editor = this.tabsAndEditorsService.activeNotebookEditor;
+		let workingSet: IWorkingSet | undefined = undefined;
+		if (editor) {
+			const format = this.alternativeNotebookContentService.getFormat(this.endpoint);
```
rebornix: It seems `Edit2` is using chat variables (which is attachment from the chat view), we could potentionally update `promptContext.chatVariables` instead of inserting `workingSet`.

src/extension/prompt/node/chatParticipantRequestHandler.ts
```diff
@@ -260,11 +285,13 @@ export class ChatParticipantRequestHandler {
 	}
 
 	private async selectIntent(command: CommandDetails | undefined, history: Turn[]): Promise<IIntent> {
-		if (!command?.intent && this.location === ChatLocation.Editor) { // TODO@jrieken do away with location specific code
+		if (!command?.intent && (this.location === ChatLocation.Editor || this.location === ChatLocation.Notebook)) { // TODO@jrieken do away with location specific code
 
 			let preferredIntent: Intent | undefined;
-			if (this.location === ChatLocation.Editor && this.documentContext && this.request.attempt === 0 && history.length === 0) { // TODO@jrieken do away with location specific code
-				if (this.documentContext.selection.isEmpty && this.documentContext.document.lineAt(this.documentContext.selection.start.line).text.trim() === '') {
+			if (this.documentContext && this.request.attempt === 0 && history.length === 0) {
```
amunger: the intent is only selected this way for the first request (though I don't understand why), so making a follow up request in the same inline chat won't have the correct prompt.

PR #315, opened 2025-07-21
feat: support sap ai core in ask mode and agent mode (credentials from a file)

veredcon:


lramos15:
This could just be done as an extension that is separate from Copilot Chat

veredcon:
@lramos15 can you elaborate how is it possible to do this through an extension and contribute an LLM provider?

lramos15:
https://github.com/microsoft/vscode-extension-samples/tree/main/chat-model-provider-sample

You won't be able to publish it to the marketplace until we finalize the API but that should be done within the next month. In the meantime this should get you started

PR #325, opened 2025-07-22
Make logprobs field optional

devm33:
In service of omitting `logprobs` from the choice response instead of sending `"logprobs": null`


ulugbekna:
@devm33 is this behavior only in proxy or CAPI as well?

PR #326, opened 2025-07-22
Expose the string used in PositionOffsetTransformer

DonJayamanne:


src/platform/editing/common/positionOffsetTransformer.ts
```diff
@@ -25,10 +25,7 @@ export class PositionOffsetTransformer {
 		this._lineStarts = new PrefixSumComputer(lineStartValues);
 	}
 
-	/**
-	 * used for testing
-	 */
-	_getText(): string {
+	getText(): string {
```
DonJayamanne, line 28: I plan on using this class for transformation of cells into alternative formats (for use in NES).

I don't want to store the original text in another variable, as this class already does that.

alexr00 (approved):

PR #331, opened 2025-07-22
If chat completion encounters content filtered error, attempt to retry in a way that avoids the filter, using new clear() method to clear the ChatResponseStream

jwangxx:
If chat completion encounters content filtered error, attempt to retry in a way that avoids the filter, using new clear() method to clear the ChatResponseStream.


jwangxx:
@microsoft-github-policy-service agree company="Microsoft"

src/extension/prompt/node/chatMLFetcher.ts
```diff
@@ -228,6 +231,58 @@ export class ChatMLFetcherImpl extends AbstractChatMLFetcher {
...
+					// Handle FilteredRetry case with augmented messages
+					if (result.type === ChatFetchResponseType.FilteredRetry) {
+
+						if (isFilterRetry !== true) {
+							streamRecorder.callback("", 0, { text: "", retryReason: result.category });
+
+							const filteredContent = result.value[0];
+							if (filteredContent) {
+								const retryMessage = (result.category === FilterReason.Copyright) ?
+									`The previous response (copied below) was filtered due to being too similar to existing public code. Please suggest something similar in function that does not match public code. Here's the previous response: ${filteredContent}\n\n` :
```
roblourens, line 247: It's surprising that this would work? Because you are including the same request that already triggered a filter and you're including the partial response that got filtered, and the filter is so twitchy that it seems very likely to just trigger it on the next round?

My understanding of the RAI filtering problems is that there are many false positives, eg the issue where an image attachment will trigger it, or a 'hello world' function will be flagged as 'copyrighted', etc, I don't see it helping with those.

src/extension/prompt/node/chatMLFetcher.ts
```diff
@@ -228,6 +231,58 @@ export class ChatMLFetcherImpl extends AbstractChatMLFetcher {
...
+									`The previous response (copied below) was filtered due to triggering our content safety filters, which looks for hateful, self-harm, sexual, or violent content. Please suggest something similar in content that does not trigger these filters. Here's the previous response: ${filteredContent}\n\n`;
+								const augmentedMessages: Raw.ChatMessage[] = [
+									...messages,
+									{
+										role: Raw.ChatRole.User,
+										content: toTextParts(retryMessage)
```
roblourens, line 253: Adding more content to the end has the possibility to exceed the context window especially if `filteredContent` is large. Doing this work in chatMLFetcher seems like a hack, you need it merged to main just to demo?

src/extension/prompt/node/chatMLFetcher.ts
jwangxx: It's working on the stock prompts we've tried (e.g. "implement fizzbuzz in javascript") and the retry prompt gives the llm sufficient direction to avoid returning the same response as before. The filter is also deterministic (there may be other factors making it feel twitchy) but it won't trigger on a new response that avoids repeating the old one. This code does one retry, which is a relatively small amount of time to wait for a potentially big payoff of not having the entire interaction undone. Happy to share a demo of it in action if it helps convince you.

src/extension/prompt/node/chatMLFetcher.ts
jwangxx: Agreed this does feel a little like a hack -- definitely open to suggestions on how it could be better incorporated. My thinking was it needs to be low enough in the stack to have access to the first response and make the higher levels not see it as an error.

roblourens: I guess ideally we would do any prompt crafting from the place that knows how to manage the context window for the prompt. Or maybe you want to say that it's fine if it fails for that reason since this is recovering from a failure anyway.

Would be simpler if this code could just try to make a request and we could handle an error response from sending too many tokens, but we don't get a consistent error code for this from CAPI so we can only handle it before sending a request.

jwangxx: Yeah, the thinking here was definitely along the lines of "anything is better than the experience now" paired with "do no harm" and to get something out to gather data on if this approach is even effective. I also need to add some telemetry to actually get the data and wasn't sure about all the details of how to use `this._telemetryService.sendTelemetryEvent` (e.g. is this microsoft or github? how do i add new fields?) -- is there a primer on that somewhere?

roblourens:
I think the main problem is just https://github.com/microsoft/vscode/pull/257271#discussion_r2229403862 and making this work in the agent loop.

src/extension/prompt/node/chatMLFetcher.ts
roblourens: I'd like to have msft telemetry so we can understand what it's doing. gh just if anyone wants it over there. You can copy an example from anywhere and copy the GDPR comment format.

src/extension/prompt/node/chatMLFetcher.ts
```diff
@@ -228,6 +231,58 @@ export class ChatMLFetcherImpl extends AbstractChatMLFetcher {
...
+								const retryResult = await this.fetchMany(
+									debugName,
+									augmentedMessages,
+									finishedCb,
+									token,
+									location,
+									chatEndpoint,
+									source,
+									requestOptions,
+									userInitiatedRequest,
```
roblourens: I think this should not be user-initiated so we don't double-charge them

src/extension/prompt/node/chatMLFetcher.ts
jwangxx: i just added it as a property on the existing telemetry (but the 'retryAfterFilterCategory' key and value only gets added if it exists)

src/extension/prompt/node/chatMLFetcher.ts
```diff
@@ -228,6 +231,58 @@ export class ChatMLFetcherImpl extends AbstractChatMLFetcher {
 			switch (response.type) {
 				case FetchResponseKind.Success: {
 					const result = await this.processSuccessfulResponse(response, messages, chatParams.ourRequestId, maxResponseTokens, tokenCount, timeToFirstToken, baseTelemetry, chatEndpoint, userInitiatedRequest);
+
+					// Handle FilteredRetry case with augmented messages
+					if (result.type === ChatFetchResponseType.FilteredRetry) {
```
roblourens, line 239: Can we add a setting to disable this? See configurationService.ts

jwangxx: done!

rzhao271 (approved):

PR #334, opened 2025-07-22
Add `when` clauses to disable og run task tools when new setting is enabled

meganrogge:
Also modify prompt agent context to indicate new tools


src/extension/tools/common/toolNames.ts
```diff
@@ -46,6 +47,7 @@ export const enum ToolName {
 	CreateDirectory = 'create_directory',
 	RunVscodeCmd = 'run_vscode_command',
 	GetTaskOutput = 'get_task_output',
+	GetTaskOutputCore = 'get_task_output2',
```
Tyriar: I've put the core ones at the bottom of the list with a `Core` prefix, could you match that?

Tyriar (changes requested):

Tyriar (approved):

PR #339, opened 2025-07-22
feat: add thinking delta to logs

karthiknadig:


src/extension/prompt/vscode-node/requestLoggerImpl.ts
```diff
@@ -343,6 +343,19 @@ export class RequestLogger extends AbstractRequestLogger {
...
+				let thinking: string = '';
+				const thinkingId = getThinkingId(d.thinking);
+				const thinkingText = getThinkingText(d.thinking);
+				if (thinkingId) {
+					thinking += `\nthinkingId: ${thinkingId}\n`;
+				}
+				if (thinkingText) {
+					thinking += `\nthinking text: ${thinkingText}\n`;
+				}
+				text += `\n## Thinking\n~~~\n${thinking}~~~\n`;
```
roblourens, line 356: Maybe don't show this if there is no text and no id

PR #346, opened 2025-07-23
Add azure open ai support to simulator model config

vritant24:


test/simulationMain.ts
```diff
@@ -867,6 +874,8 @@ function parseModelConfigFile(modelConfigFilePath: string): IModelConfig[] {
 			id: modelId,
 			name: model.name,
 			version: model.version,
+			type: model.type,
+			useDeveloperRole: model.useDeveloperRole ?? false,
```
lramos15, line 878: Should we update the comment config for this.

src/platform/endpoint/test/node/openaiCompatibleEndpoint.ts
```diff
@@ -105,26 +107,40 @@ export class OpenAICompatibleTestEndpoint extends ChatEndpoint {
 			throw new Error(`API key environment variable ${this.modelConfig.apiKeyEnvName} is not set`);
 		}
 
+		if (this.modelConfig.type === 'azureOpenai') {
```
lramos15, line 110: Should this be in a separate azure endpoint? Just in case they continue to diverge? Maybe via na override. You can see BYOK does this

test/simulationMain.ts
vritant24: It does have `type` and `useDeveloperRole`. Is there anything else you want there?

src/platform/endpoint/test/node/openaiCompatibleEndpoint.ts
vritant24: what do you mean by `na override`?
I was on the side of keeping it all in one, and create a separate azure endpoint if it diverges a lot. I'm still unsure of how much more if-elsing is going to be needed to be able to know the right separations (probably will have more knowledge as we continue to try more types of models).

lramos15 (approved):

rzhao271 (approved):

PR #355, opened 2025-07-24
chat endpoint: allow to specify url/request-metadata in model info

ulugbekna:
to avoid verbose inheritance code just to override a static field

### DIFF
--- a/src/platform/endpoint/common/endpointProvider.ts
+++ b/src/platform/endpoint/common/endpointProvider.ts
@@ -4,12 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 
+import { RequestMetadata } from '@vscode/copilot-api';
 import type { LanguageModelChat } from 'vscode';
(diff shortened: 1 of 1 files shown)


src/platform/endpoint/common/endpointProvider.ts
```diff
@@ -51,6 +52,7 @@ type ICompletionsModelCapabilities = {
 export interface IModelAPIResponse {
 	id: string;
 	name: string;
+	urlOrRequestMetadata?: string | RequestMetadata;
```
lramos15: Should we extend this interface so it's not appearing as if this is something that is returned by CAPI? ExtendedTokenInfo is an example where we do this

lramos15 (approved):

roblourens:
I'm working in this area and it's confusing that an xtab endpoint is a "ChatEndpoint" because it's not what I think of as "chat"

PR #357, opened 2025-07-24
rm task tools

meganrogge:


rzhao271 (dismissed):

package.json
```diff
@@ -1192,15 +1065,6 @@
 					"getProjectSetupInfo",
 					"installExtension"
 				]
-			},
-			{
-				"name": "runTasks",
```
roblourens, line 1197: Is this registered from core now?

roblourens (approved):

package.json
meganrogge: yes

meganrogge: https://github.com/microsoft/vscode/blob/7a614035a9aa8c0a74c225ee04c180e767fd07a7/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task.chatAgentTools.contribution.ts#L40-L42

meganrogge: looks like I need to rename that `runTasks`, will do.

meganrogge: https://github.com/microsoft/vscode/pull/257730

PR #360, opened 2025-07-24
more model config support for simulations

vritant24:
1. Allow disabling temperature in simulation
2. remove `max_tokens` for openai endpoints (taken from `openaiEndpoint.ts`).

### DIFF
--- a/src/platform/endpoint/test/node/openaiCompatibleEndpoint.ts
+++ b/src/platform/endpoint/test/node/openaiCompatibleEndpoint.ts
@@ -26,6 +26,7 @@ export type IModelConfig = {
 	version: string;
 	type: 'openai' | 'azureOpenai';
 	useDeveloperRole: boolean;
+	disableTemperature: boolean;
 	capabilities: {
 		supports: {
 			parallel_tool_calls: boolean;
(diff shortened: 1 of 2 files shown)


lramos15 (commented):
I'd ideally like this to be more generic like the following


```
paramOverrides: {temperature: undefined, top_p: 12}
```

and when undefined we delete it

roblourens (approved):

devmuhnnad:
I'm passing overrides object into the openai compitable model config, but it dosen't work! I'm trying to change the temperature,

HERE IS THE chatLanguageModels.json

```json
[
	{
		"name": "Moonshot",
		"vendor": "customoai",
		"type": "openai",
		"apiKey": "${input:chat.lm.secret.-10f3e3f6}",
		"models": [
			{
				"id": "kimi-k2.5",
				"name": "kimi",
				"url": "https://silly-unicorn-82.webhook.cool",
				"toolCalling": true,
				"vision": true,
				"maxInputTokens": 128000,
				"maxOutputTokens": 16000,
				"overrides": {
					"temperature": 10,
					"top_p": null
				}
			}
		]
	}
]
```

PR #377, opened 2025-07-28
TextEdit to TextReplacement for Notebook in NES

DonJayamanne:


src/extension/inlineEdits/vscode-node/features/diagnosticsBasedCompletions/diagnosticsCompletions.ts
```diff
@@ -312,10 +313,21 @@ export class CodeAction {
...
 
 	getEditForWorkspaceDocument(workspaceDocument: IVSCodeObservableDocument): TextReplacement[] | undefined {
-		if (!this.edit) {
+		const edit = this.edit;
+		if (!edit) {
 			return undefined;
 		}
-		return this.edit.get(workspaceDocument.id.toUri()).map(toInternalTextEdit);
+		if (workspaceDocument.kind === 'textDocument') {
+			return edit.get(workspaceDocument.id.toUri()).map(e => toInternalTextEdit(e.range, e.newText));
```
DonJayamanne, line 321: Very similar to 
I think the best is to create a separate file with these helpers that just accepts the IVSCodeObservable.
E.g. `toDiagnostics(vscode.Diagnostic): Diagnostic` `toTextReplacement(vscode.TextEdit): TextReplacement`
etc

https://github.com/microsoft/vscode-copilot-chat/blob/1492647ce907ef141004f5626e7835b744cb6598/src/extension/inlineEdits/vscode-node/features/diagnosticsCompletionProcessor.ts#L311-L320

DonJayamanne: There could be others, 
I want to first get an idea of what all of these additional changes are, before refactoring these or introducing other layers/helpers. 
Yes this is just building up debt, but will fix these just as i did earlier with few others.

hediet: I believe `IVSCodeDocument` should offer a nice abstraction here - consumers (such as this one) shouldn't care if its a notebook or normal text document and the API should be such that they cannot use it wrongly easily. That abstraction should offer diagnostics and fixes for IObservableDocuments.

I believe `kind` should disappear with that debt-week cleanup.

PR #389, opened 2025-07-29
fix: enable apply patch for more models

karthiknadig:


src/platform/endpoint/common/chatModelCapabilities.ts
```diff
@@ -28,6 +38,13 @@ export function modelPrefersInstructionsAfterHistory(modelFamily: string) {
 export function modelSupportsApplyPatch(model: LanguageModelChat | IChatEndpoint): boolean {
 	return model.family === 'gpt-4.1' || model.family === 'o4-mini';
 }
+export async function modelSupportsApplyPatchAsync(model: LanguageModelChat | IChatEndpoint): Promise<boolean> {
```
connor4312: I see why you didn't keep one function and make it globally async, but I think instead `modelSupportsApplyPatch` should be made globally async, and we add another function `modelPrefersJsonNotebookRepresentation`. The point of these functions were to explicitly reflect model capabilities and keying json notebooks off apply patch support was a smell in the first place.

karthiknadig: Done

connor4312 (approved):

PR #399, opened 2025-07-30
When rendering the prompts, exclude turns from the history that errored due to prompt filtration

jwangxx:
This was resulting in users getting stuck in a state where, after one prompt that triggered an RAI 422 error, all their subsequent requests were being filtered.

### DIFF
--- a/src/extension/prompts/node/panel/conversationHistory.tsx
+++ b/src/extension/prompts/node/panel/conversationHistory.tsx
@@ -73,7 +73,7 @@ export class HistoryWithInstructions extends PromptElement<Omit<ConversationHist
 export class ConversationHistory extends PromptElement<ConversationHistoryProps> {
 	override render(_state: void, _sizing: PromptSizing): PromptPiece<any, any> | undefined {
 		// exclude turns from the history that errored due to prompt filtration
-		let turnHistory = this.props.history.filter(turn => turn.responseStatus !== TurnStatus.Filtered);
+		let turnHistory = this.props.history.filter(turn => turn.responseStatus !== TurnStatus.PromptFiltered);
 
 		if (this.props.inline && turnHistory.length > 0) {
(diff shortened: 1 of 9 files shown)


src/extension/prompts/node/panel/conversationHistory.tsx
```diff
@@ -72,7 +72,8 @@ export class HistoryWithInstructions extends PromptElement<Omit<ConversationHist
  */
 export class ConversationHistory extends PromptElement<ConversationHistoryProps> {
 	override render(_state: void, _sizing: PromptSizing): PromptPiece<any, any> | undefined {
-		let turnHistory = this.props.history;
+		// exclude turns from the history that errored due to prompt filtration
+		let turnHistory = this.props.history.filter(turn => turn.responseStatus !== TurnStatus.PromptFiltered);
```
johnmog, line 76: Below there's a check in the forEach loop for Filtered responses, would a similar approach work for prompt filtered responses?

jwangxx: in the loop below, we are always including the user message and then filtering out the assistant response if it has responseStatus of TurnStatus.Filtered. in contrast, for turns with responseStatus of TurnStatus.PromptFiltered, we don't want the user message included (nor the non-existent assistant response)

src/platform/chat/common/commonTypes.ts
```diff
@@ -87,6 +87,7 @@ export enum ChatFetchResponseType {
 	OffTopic = 'offTopic',
 	Canceled = 'canceled',
 	Filtered = 'filtered',
+	PromptFiltered = 'promptFiltered',
```
roblourens, line 90: Is this a new thing? Or we just never handled it before?

roblourens: And do you have an example to trigger this? I tried some things and I mostly only get "Sorry I can't assist with this"

roblourens: Nvm I got it

PR #407, opened 2025-07-30
send thinking tokens progress to chat ui

justschen:
for rendering thinking in chat


src/extension/prompt/node/pseudoStartStopConversationCallback.ts
```diff
@@ -61,8 +61,8 @@ export class PseudoStopStartResponseProcessor implements IResponseProcessor {
 			progress.prepareToolInvocation(getContributedToolName(delta.beginToolCalls[0].name));
 		}
 
-		if (delta.thinking) {
-			// progress.thinking(delta.thinking);
+		if (delta.thinking?.text) {
+			progress.thinkingProgress(delta.thinking.text);
```
karthiknadig:
```suggestion
		if (delta.thinking) {
		        const text = delta.thinking.text ?? '';
			progress.thinkingProgress(text);
```
Otherwise we don't update the `id` in thinking data service. The id/metadata should also be sent to core right?

PR #412, opened 2025-07-31
Add conversation id and message id to relevant events

24anisha:


src/extension/mcp/vscode-node/mcpToolCallingLoop.tsx
```diff
@@ -101,7 +101,8 @@ export class McpToolCallingLoop extends ToolCallingLoop<IMcpToolCallingLoopOptio
 			firstFetchCall,
 			{
 				messageId: randomUUID(),
-				messageSource: McpToolCallingLoop.ID
+				messageSource: McpToolCallingLoop.ID,
+				turnIndex: this.options.conversation.turns.length.toString(),
```
roblourens: Can you pass it as a number instead of a string?

src/extension/prompt/node/chatParticipantTelemetry.ts
```diff
@@ -451,8 +451,16 @@ export abstract class ChatTelemetry<C extends IDocumentContext | undefined = IDo
 
 		this._telemetryService.sendInternalMSFTTelemetryEvent('toolCallDetailsInternal', {
 			...toolCallProperties,
-			availableTools: JSON.stringify(availableTools.map(tool => tool.name)),
+			messageId: this.telemetryMessageId,
+			availableTools: JSON.stringify(availableTools.map(tool => tool.name))
 		}, toolCallMeasurements);
+
+		this._telemetryService.sendEnhancedGHTelemetryEvent('toolCallDetailsRestricted', {
```
roblourens: I would leave 'restricted' out of the event name

src/platform/networking/node/chatStream.ts
```diff
@@ -46,7 +47,13 @@ export function prepareChatCompletionForReturn(
 		role: Raw.ChatRole.Assistant,
 		content: toTextParts(messageContent),
 	};
-	sendEngineMessagesTelemetry(telemetryService, [rawMessageToCAPI(message)], telemetryData);
+
+	// Try to get turnIndex from telemetryData properties first, fallback to parameter
+	const effectiveTurnIndex = telemetryData.properties.turnIndex ?
```
roblourens: This is confusing, what's going on here?

src/extension/mcp/vscode-node/mcpToolCallingLoop.tsx
24anisha: I'm a little confused by this. when I try to set it to be a number, it errors because turnIndex is supposed to be a string. I see toString() a lot in the repo with turnIndex -- what determines whether it should be passed as a number or a string?

src/platform/networking/node/chatStream.ts
24anisha: if a turn index is already passed through, it uses that. if not, it calculates it and adds it

src/extension/mcp/vscode-node/mcpToolCallingLoop.tsx
roblourens: I guess this is just telemetry properties and we don't have a way to pass measurements from here. But I'm not sure you care about this case anyway? This is the mcp setup feature which doesn't actually happen in chat, it's only operating on a fake conversation here https://github.com/microsoft/vscode-copilot-chat/blob/2f73defc0d45b969fa72d4ca3cd2e0a781d0ec15/src/extension/mcp/vscode-node/commands.ts#L111

src/platform/networking/node/chatStream.ts
roblourens: So where are you calling `prepareChatCompletionForReturn` with a turnIndex?

src/extension/prompt/node/codebaseToolCalling.ts
```diff
@@ -88,7 +88,8 @@ export class CodebaseToolCallingLoop extends ToolCallingLoop<ICodebaseToolCallin
 			false,
 			{
 				messageId: randomUUID(), // @TODO@joyceerhl
-				messageSource: CodebaseToolCallingLoop.ID
+				messageSource: CodebaseToolCallingLoop.ID,
+				turnIndex: this.options.conversation.turns.length.toString(),
```
roblourens: For this stuff, seems like it would be better to have it as a number, but we just are only passing properties here. So you could come up with some way to pass a number through (the way this gh telemetry data gets passed around is kind of crazy) or if you want to work with it as a string in telemetry, this is fine, either way.

rzhao271 (approved):

PR #443, opened 2025-08-04
Add a switch for Instant Apply to use short model for lower context prompts

vritant24:
1. Add a new endpoint that extends the `Proxy4OEndpoint` which instead uses a specified short model with the same capabilities and limits.
2. Add a short context limit that switches the model used for instant apply in code mapper when hit. the default is to use the short model for < 8k tokens.
3. Add support to change the short model and short context limit through exp service.


src/platform/endpoint/node/proxy4oEndpoint.ts
```diff
@@ -34,9 +35,11 @@ export class Proxy4oEndpoint extends ChatEndpoint {
 		@ITokenizerProvider tokenizerProvider: ITokenizerProvider,
 		@IInstantiationService instantiationService: IInstantiationService,
 		@IConfigurationService configurationService: IConfigurationService,
-		@IExperimentationService experimentationService: IExperimentationService
+		@IExperimentationService experimentationService: IExperimentationService,
 	) {
-		const model = configurationService.getExperimentBasedConfig<string>(ConfigKey.Internal.InstantApplyModelName, experimentationService) ?? 'gpt-4o-instant-apply-full-ft-v66';
+		if (!model) {
+			model = configurationService.getExperimentBasedConfig<string>(ConfigKey.Internal.InstantApplyModelName, experimentationService) ?? CHAT_MODEL.GPT4OPROXY;
```
vritant24: was previously a string, changed it to use the enum

vritant24:
cc @roblourens

roblourens:
I'm trying to update the test cache for your PR but running into an issue with using lfs with a fork. If I can't figure it out I'll recreate your PR on this repo.

roblourens:
Still working on this, if I don't know the real solution by tomorrow I'll just recreate the PR so it can go in the normal way

roblourens:
Moving to https://github.com/microsoft/vscode-copilot-chat/pull/497, trying there

PR #451, opened 2025-08-05
Improve testFailure tool docs

digitarald:
Based on recent analysis that NEVER usedthe test failure tools to debug failing tests.


connor4312:
Models shouldn't really ever use this tool directly, since they get the output they need from the test run tool. This is useful for referencing in the prompt but we should actually exclude it from the model otherwise.

digitarald:
Ok, if we plan to exclude it that would be a better tool flow.

PR #463, opened 2025-08-05
nes: respect additionalUris' copilot ignore

ulugbekna:
fix https://github.com/microsoft/vscode/issues/259791


src/extension/xtab/node/xtabProvider.ts
```diff
@@ -370,10 +373,24 @@ export class XtabProvider extends ChainedStatelessNextEditProvider {
 				activeExperiments: new Map(),
 				timeBudget: debounceTime
 			};
+
+			const isSnippetIgnored = async (item: ContextItem): Promise<boolean> => {
+				if (item.kind === ContextKind.Snippet && item.additionalUris) {
```
dbaeumer: You need to check both uri and additionalUris on a snippet item. Or does this happen somewhere else?

dbaeumer (changes requested):

src/extension/xtab/node/xtabProvider.ts
ulugbekna: thank you for the catch! addressed

PR #469, opened 2025-08-05
Show token use against available context window on Chat message hover

pierceboggan:
This PR adds support for visualizing current token usage against the model's available context window upon hover on a chat message:
image

This information was already present in the Copilot Chat debug log that is available with "Developer: Show Chat Debug View", but now is more visible and present directly within the chat message.

### DIFF
--- a/src/extension/prompt/common/conversation.ts
+++ b/src/extension/prompt/common/conversation.ts
@@ -328,6 +328,16 @@ export interface IResultMetadata {
 	toolCallResults?: Record<string, LanguageModelToolResult>;
 	maxToolCallsExceeded?: boolean;
 	summary?: { toolCallRoundId: string; text: string };
+
+	/**
+	 * Token usage information for the response
+	 */
+	tokenUsage?: {
+		promptTokens: number;
+		completionTokens?: number;
+		totalTokens: number;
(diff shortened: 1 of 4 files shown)


cmtzco:
I'm a fan of the bar shown in the AI Usage Statistics. When hovered it could show the actual token usage and percentage. 

image

demiliani:
Why not showing something more clear, like the following:
IMG_1504

cmtzco:
> Why not showing something more clear, like the following:
> 
> IMG_1504
> 
> 

While I agree that more information is generally better, not all users are advanced and prefer to see all that data. Instead, displaying it in a quick glance bar, such as the context usage percentage, would be more user-friendly and less overwhelming. Additionally, with the current amount of information already present in the window, adding more stats would only clutter the screen and reduce the available space for work or chat. Hovering over the relevant information would be a great way to display this data without taking up too much space on the window.

isidorn:
Thanks @pierceboggan for starting with this PR. I will try to pick this up with making changes on the core side (as opposed to contributing this from an extension). I will go ahead and close this PR for now, but the work will continue in a new one in vscode - https://github.com/microsoft/vscode/pull/261829

PR #471, opened 2025-08-05
Add a save icon for the Copilot Chat debug view

pierceboggan:
This PR adds a save icon to the Copilot Chat log that opens after clicking an entry in the Copilot Chat debug view. The save experience defaults to using .md rather than .copilotmd (which GitHub does not like as an attachment).

image

This allows faster bug reporting from Chat logs.

### DIFF
--- a/src/extension/log/vscode-node/requestLogTree.ts
+++ b/src/extension/log/vscode-node/requestLogTree.ts
@@ -22,6 +22,7 @@ import { IExtensionContribution } from '../../common/contributions';
 const showHtmlCommand = 'vscode.copilot.chat.showRequestHtmlItem';
 const exportLogItemCommand = 'github.copilot.chat.debug.exportLogItem';
 const exportPromptArchiveCommand = 'github.copilot.chat.debug.exportPromptArchive';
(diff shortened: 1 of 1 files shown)


roblourens:
> rather than .copilotmd (which GitHub does not like as an attachment).

They just changed this :) https://vscodeteam.slack.com/archives/C03K6481BNV/p1754907742801079?thread_ts=1754563889.190279&cid=C03K6481BNV

I don't see this button, I think this PR is missing a package.json entry?

src/extension/log/vscode-node/requestLogTree.ts
```diff
@@ -133,6 +134,71 @@ export class RequestLogTree extends Disposable implements IExtensionContribution
...
+			});
+
+			if (!saveUri) {
+				return; // User cancelled
+			}
+
+			try {
+				// Read the text from the virtual document URI explicitly
+				const doc = await vscode.workspace.openTextDocument(resource);
+				await vscode.workspace.fs.writeFile(saveUri, Buffer.from(doc.getText(), 'utf8'));
```
roblourens, line 186: You might want `vscode.workspace.saveAs`

PR #473, opened 2025-08-05
Add message length events to MSFT internal telemetry 

yemohyleyemohyle:
Add a telemetry event that would have a similar structure as engine.messages events but with content replaced by length of that content. This way the messages will be much shorter, not truncated, les likely to drop, and will allow to compute input cost of each tool - total number of characters contributed by a tool to all the conversation input messages it is included in.

### DIFF
--- a/src/extension/prompts/node/agent/agentInstructions.tsx
+++ b/src/extension/prompts/node/agent/agentInstructions.tsx
@@ -5,6 +5,7 @@
 
 import { BasePromptElementProps, PromptElement, PromptSizing } from '@vscode/prompt-tsx';
 import type { LanguageModelToolInformation } from 'vscode';
+import { LanguageModelToolMCPSource } from '../../../../vscodeTypes';
 import { ToolName } from '../../../tools/common/toolNames';
 import { IToolsService } from '../../../tools/common/toolsService';
 import { InstructionMessage } from '../base/instructionMessage';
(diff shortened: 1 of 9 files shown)


src/platform/networking/node/chatStream.ts
```diff
@@ -12,12 +12,81 @@ import { APIJsonData, CAPIChatMessage, ChatCompletion, rawMessageToCAPI } from '
 import { FinishedCompletion, convertToAPIJsonData } from './stream';
 
 // TODO @lramos15 - Find a better file for this, since this file is for the chat stream and should not be telemetry related
-export function sendEngineMessagesTelemetry(telemetryService: ITelemetryService, messages: CAPIChatMessage[], telemetryData: TelemetryData) {
+export function sendEngineMessagesLengthTelemetry(telemetryService: ITelemetryService, messages: CAPIChatMessage[], telemetryData: TelemetryData, logService?: ILogService) {
+	// Determine if this is input or output based on message characteristics
+	const isOutput = messages.length === 1 && messages[0].role === 'assistant';
```
roblourens: Instead of trying to guess this from the input would it be easier to just pass a flag as a parameter or something?

src/platform/networking/node/chatStream.ts
```diff
@@ -46,7 +115,18 @@ export function prepareChatCompletionForReturn(
 		content: toTextParts(messageContent),
 	};
 
-	sendEngineMessagesTelemetry(telemetryService, [rawMessageToCAPI(message)], telemetryData);
+	// Create enhanced message for telemetry with usage information
+	const telemetryMessage = rawMessageToCAPI(message);
+
+	// Add usage information if available
+	if (c.usage) {
+		(telemetryMessage as any).usage = c.usage;
```
roblourens: I'd prefer that we don't use objects as random property bags but put this in telemetryData or somewhere else properly typed

src/platform/networking/node/chatStream.ts
```diff
@@ -12,12 +12,81 @@ import { APIJsonData, CAPIChatMessage, ChatCompletion, rawMessageToCAPI } from '
...
+		return processedMsg;
+	});
+
+	// Log the messages before sending to telemetry with modelCallId as top-level property
+	const logData = {
+		modelCallId: modelCallId,
+		messageType: messageType,
+		messages: messagesWithLength
+	};
+	logService?.info(`[TELEMETRY] engine.messages.length: ${JSON.stringify(logData, null, 2)}`);
```
roblourens: I don't think we should log stuff like this to `info`.  Outgoing telemetry is logged to the telemetry channel already anyway

DonJayamanne (approved):

src/platform/networking/node/chatStream.ts
```diff
@@ -64,7 +64,7 @@ export function sendEngineMessagesLengthTelemetry(telemetryService: ITelemetrySe
 		messageType: messageType,
 		messages: messagesWithLength
 	};
-	logService?.info(`[TELEMETRY] engine.messages.length: ${JSON.stringify(logData, null, 2)}`);
+	logService?.debug(`[TELEMETRY] engine.messages.length: ${JSON.stringify(logData, null, 2)}`);
```
roblourens, line 67: This is kind of noisy, why do you need to log it?

PR #479, opened 2025-08-06
Add "onExP" tag to enableRetryAfterFilteredResponse setting

jwangxx:


bpasero (approved):

Tyriar (approved):

roblourens:
I'm not actually sure we need this tag? It's used in some settings but I don't think it actually does anything in the extension... getting confirmation before I merge this

roblourens:
This just ensures that the UI is synced with the exp state

PR #497, opened 2025-08-07
Add a switch for Instant Apply to use short model for lower context prompts 

roblourens:
Had to port https://github.com/microsoft/vscode-copilot-chat/pull/443 to a branch on this repo


dbaeumer (approved):

roblourens:
I realized that the codemapper is used for some inline tests too, there are a pretty large set of these, I'm trying to update the cache for them now

PR #500, opened 2025-08-07
introuce networkError as a new type for CAPI errors

isidorn:
Fixes microsoft/vscode-internalbacklog/issues/5687

@lramos15 I took a stab to separate the network requests from the generic FAILED. This will allow us to better filter down the error rates in our dashboards. Let me know if there is something that I should improve in this PR.


lramos15:
Some CI failures

isidorn:
@lramos15 sorry about that. I just fixed those errors.

lramos15 (approved):

PR #507, opened 2025-08-07
send telemetry from EH as well

amunger:
follow up on https://github.com/microsoft/vscode/pull/257498 to send data over to GH

### DIFF
--- a/src/extension/conversation/vscode-node/userActions.ts
+++ b/src/extension/conversation/vscode-node/userActions.ts
@@ -185,34 +185,32 @@ export class UserFeedbackService implements IUserFeedbackService {
 					}
 				}
 				break;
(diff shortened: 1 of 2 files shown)


src/extension/conversation/vscode-node/userActions.ts
```diff
@@ -191,6 +185,34 @@ export class UserFeedbackService implements IUserFeedbackService {
 					}
 				}
 				break;
+			case 'chatEditingHunkAction':
+
+				/* __GDPR__
+					"panel.edit.action" : {
+						"owner": "joyceerhl",
```
roblourens: 😓

roblourens (approved):

PR #518, opened 2025-08-08
Update copied VS sources

roblourens:


roblourens:
Blocked on https://github.com/microsoft/vscode/issues/260651 and https://github.com/microsoft/vscode/issues/260650

PR #531, opened 2025-08-09
fix: update apply_patch format instructions to be consistent and include required space

agreaves-ms:
### ApplyPatchFormatInstructions

- Made sure the model knows to always have a space before `[context_before]` and `[context_after]`
- Made sure the model knows to always have a `-` before original old code lines
- Made sure the model knows to always have a `+` after new code lines

### ApplyPatchInstructions
- Explicitly tell GPT-5 to not use ApplyPatch for create file (*GPT-5 would try and create a file with apply_patch then fail over and over until eventually using the create_file tool*)

## Details

The apply_patch instructions and description would sometimes confuse the models and produce patches that where either missing an additional space infront of unchanging context lines, or it would consider `-` characters from the original text as old code lines.

### DIFF
--- a/src/extension/prompts/node/agent/agentInstructions.tsx
+++ b/src/extension/prompts/node/agent/agentInstructions.tsx
@@ -400,12 +400,15 @@ export class ApplyPatchFormatInstructions extends PromptElement {
 	render() {
 		return <>
 			*** Update File: [file_path]<br />
-			[context_before] -&gt; See below for further instructions on context.<br />
+			&nbsp;[context_before] -&gt; See below for further instructions on context.<br />
 			-[old_code] -&gt; Precede each line in the old code with a minus sign.<br />
 			+[new_code] -&gt; Precede each line in the new, replacement code with a plus sign.<br />
-			[context_after] -&gt; See below for further instructions on context.<br />
+			&nbsp;[context_after] -&gt; See below for further instructions on context.<br />
 			<br />
+			Prefix each line of old code with exactly one minus (-) before its original first character; keep the original indentation after it.<br />
+			Prefix each line of new code with exactly one plus (+), the remainder of the line after the plus is the full new line content.<br />
 			For instructions on [context_before] and [context_after]:<br />
+			- Prefix each unchanged context line with exactly one ASCII space (U+0020) before its original first character; keep the original indentation after it.<br />
 			- By default, show 3 lines of code immediately above and 3 lines immediately below each change. If a change is within 3 lines of a previous change, do NOT duplicate the first change's [context_after] lines in the second change's [context_before] lines.<br />
 			- If 3 lines of context is insufficient to uniquely identify the snippet of code within the file, use the @@ operator to indicate the class or function to which the snippet belongs.<br />
 			- If a code block is repeated so many times in a class or function such that even a single @@ statement and 3 lines of context cannot uniquely identify the snippet of code, you can use multiple `@@` statements to jump to the right context.
@@ -413,17 +416,18 @@ export class ApplyPatchFormatInstructions extends PromptElement {
 			You must use the same indentation style as the original code. If the original code uses tabs, you must use tabs. If the original code uses spaces, you must use spaces. Be sure to use a proper UNESCAPED tab character.<br />
 			<br />
 			See below for an example of the patch format. If you propose changes to multiple regions in the same file, you should repeat the *** Update File header for each snippet of code to change:<br />
-			<br />
-			*** Begin Patch<br />
-			*** Update File: /Users/someone/pygorithm/searching/binary_search.py<br />
-			@@ class BaseClass<br />
-			@@   def method():<br />
-			[3 lines of pre-context]<br />
-			-[old_code]<br />
-			+[new_code]<br />
-			+[new_code]<br />
-			[3 lines of post-context]<br />
-			*** End Patch<br />
+			<Tag name='applyPatchExample'>
+				*** Begin Patch<br />
+				*** Update File: /Users/someone/pygorithm/searching/binary_search.py<br />
+				@@ class BaseClass<br />
+				@@   def method():<br />
+				&nbsp;[3 lines of pre-context]<br />
+				-[old_code]<br />
+				+[new_code]<br />
+				+[new_code]<br />
+				&nbsp;[3 lines of post-context]<br />
+				*** End Patch<br />
+			</Tag>
 		</>;
 	}
 }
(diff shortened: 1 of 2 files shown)


src/extension/prompts/node/agent/agentInstructions.tsx
```diff
@@ -433,7 +437,9 @@ class ApplyPatchInstructions extends PromptElement<DefaultAgentPromptProps> {
 		const isGpt5 = this.props.modelFamily === 'gpt-5';
 		return <Tag name='applyPatchInstructions'>
 			To edit files in the workspace, use the {ToolName.ApplyPatch} tool. If you have issues with it, you should first try to fix your patch and continue using {ToolName.ApplyPatch}. If you are stuck, you can fall back on the {ToolName.EditFile} tool. But {ToolName.ApplyPatch} is much faster and is the preferred tool.<br />
-			{isGpt5 && <>Prefer the smallest set of changes needed to satisfy the task. Avoid reformatting unrelated code; preserve existing style and public APIs unless the task requires changes. When practical, complete all edits for a file within a single message.<br /></>}
+			{isGpt5 && <>
+				Never use {ToolName.ApplyPatch} for creating files.<br />
+				Prefer the smallest set of changes needed to satisfy the task. Avoid reformatting unrelated code; preserve existing style and public APIs unless the task requires changes. When practical, complete all edits for a file within a single message.<br /></>}
```
agreaves-ms, line 446: Also, do we need these instructions? Would they limit GPT-5 at all and have GPT-5 focus on only making small simple changes?

```text
Prefer the smallest set of changes needed to satisfy the task. Avoid reformatting unrelated code; preserve existing style and public APIs unless the task requires changes. When practical, complete all edits for a file within a single message.<br />
```

roblourens: Not sure what the origin of this line is. cc @bhavyaus

src/extension/prompts/node/agent/agentInstructions.tsx
```diff
@@ -433,7 +441,9 @@ class ApplyPatchInstructions extends PromptElement<DefaultAgentPromptProps> {
 		const isGpt5 = this.props.modelFamily === 'gpt-5';
 		return <Tag name='applyPatchInstructions'>
 			To edit files in the workspace, use the {ToolName.ApplyPatch} tool. If you have issues with it, you should first try to fix your patch and continue using {ToolName.ApplyPatch}. If you are stuck, you can fall back on the {ToolName.EditFile} tool. But {ToolName.ApplyPatch} is much faster and is the preferred tool.<br />
-			{isGpt5 && <>Prefer the smallest set of changes needed to satisfy the task. Avoid reformatting unrelated code; preserve existing style and public APIs unless the task requires changes. When practical, complete all edits for a file within a single message.<br /></>}
+			{isGpt5 && <>
+				Never use {ToolName.ApplyPatch} for creating files.<br />
```
connor4312, line 445: Creating a file with apply patch is fine, we should handle it

https://github.com/microsoft/vscode-copilot-chat/blob/6eeee7777628684be3f12cd2c0bb477c761c245b/src/extension/tools/node/applyPatch/parseApplyPatch.ts#L44

If we don't, that is a bug we should fix

src/extension/prompts/node/agent/agentInstructions.tsx
```diff
@@ -399,31 +399,39 @@ export class SweBenchAgentPrompt extends PromptElement<DefaultAgentPromptProps>
...
-			[context_after] -&gt; See below for further instructions on context.<br />
+			{[
+				`*** Update File: [file_path]`,
+				` [context_before] -> See below for further instructions on context.`,
+				`-[old_code] -> Precede each line in the old code with a minus sign.`,
+				`+[new_code] -> Precede each line in the new, replacement code with a plus sign.`,
+				` [context_after] -> See below for further instructions on context.`
+			].join('\n')}<br />
 			<br />
+			Prefix each line of old code with exactly one minus (-) before its original first character; keep the original indentation after it.<br />
```
connor4312, line 410: These prefix instructions don't appear in Codex's apply patch instructions, so I would tend to say we should not include them here unless we have strong reason to believe they improve behavior https://github.com/openai/codex/blob/main/codex-rs/apply-patch/apply_patch_tool_instructions.md

src/extension/prompts/node/agent/agentInstructions.tsx
agreaves-ms: I can do some additional testing today as I'm working on things and report back regarding apply_patch failing on creating a file.

src/extension/prompts/node/agent/agentInstructions.tsx
agreaves-ms: Good point, I can look at Codex's code first and see where this is a problem. Also I'll try testing this out further and produce smaller set of sample data to reproduce the issue.

src/extension/prompts/node/agent/agentInstructions.tsx
connor4312: We don't explicitly instruct it to use Add File afaik in our instructions, but GPT is fine-tuned on apply_patch so it might include it nevertheless

src/extension/prompts/node/agent/agentInstructions.tsx
agreaves-ms: I made a repo with contrived examples, capturing conversation, example file before conversation, results, and all logs.

https://github.com/agreaves-ms/vscode-copilot-chat-example-issues

The one example I have in there right now is specifically around what these instructions were meant to fix.

@connor4312 do you think it makes sense to share this and address these instruction changes in Codex's repo first?

I don't believe it's possible to work around this issue in apply_patch's logic as it would always see something like `- list item` differently than ` - list item`.

Captured screenshot of the problem:

image

`### 🔐 SSL and Kubernetes Integration (SSE-Only)` should be ` ### 🔐 SSL and Kubernetes Integration (SSE-Only)`

src/extension/prompts/node/agent/agentInstructions.tsx
bhavyaus: This was recommended from OAI. I believe this was to partly fix issues such as https://github.com/microsoft/vscode/issues/261052 to keep the model from being very verbose with the generated code and comments. 
Since
> Would they limit GPT-5 at all and have GPT-5 focus on only making small simple changes?

 isn't an behavior we're observing, please leave this one in for now.

PR #539, opened 2025-08-10
feature: chat export

maifeeulasad:
closes https://github.com/microsoft/vscode/issues/260769

### DIFF
--- a/src/extension/exportChat/vscode-node/exportChatCommands.ts
+++ b/src/extension/exportChat/vscode-node/exportChatCommands.ts
@@ -1,0 +1,207 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { l10n } from 'vscode';
+import { ILogService } from '../../../platform/log/common/logService';
+import { ITelemetryService } from '../../../platform/telemetry/common/telemetry';
+import { DisposableStore, IDisposable } from '../../../util/vs/base/common/lifecycle';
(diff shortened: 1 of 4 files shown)


maifeeulasad:
ping @justschen

maifeeulasad:

src/extension/exportChat/vscode-node/exportChatCommands.ts
```diff
@@ -0,0 +1,411 @@
+/*---------------------------------------------------------------------------------------------
```
tariqshams, line 1: Great idea implementing this! 

As a suggestion for future maintainability, you could break this file down into smaller functions that could be imported from helper files. For example, you could create functions to handle certain functionality to reduce duplication and help readability.

```typescript
function renderTurn(turn): string { ... }
function renderToolCalls(toolCallRounds): string { ... }
function renderRequest(request): string { ... }
function renderResponse(response): string { ... }
```

maifeeulasad:
Hey @tariqshams thanks for the review, I have done so, care to check again.

If you think it requires further modifications, please let me know.

maifeeulasad:
@tariqshams @justschen please let me know, if we need both export as `.json` and `.md`, or just only one. Right now, this branch has both. (Personally I prefer json)

Do we need export selected chat? Or all chats? (right now it's only doing the selected/last chat)

I will happily do it.

vivganes:
I am not a maintainer.  However here's some info that can be useful in this PR.

Currently, the "Copy All" function in the chat's context menu copies the markdown of the entire conversation into the clipboard. It could be useful to build upon that instead of rendering our own markdown.  This makes sure that in case of changes, we will do it in one place instead of two.

Image

maifeeulasad:
@vivganes thanks for your feedback. I have taken your comment into accounting, and I will raise another merge request for this.

Sorry for the late reply!

PR #546, opened 2025-08-11
Use .NET SDK to search for NuGet packages, emit events, add tests for all package types

joelverhagen:
Progress on https://github.com/microsoft/vscode/issues/257679 and https://github.com/microsoft/vscode/issues/259901 (switch to .NET SDK to allow for custom package sources in the future -- still hard coded to nuget.org right now).

Depends on https://github.com/microsoft/vscode/pull/259634 (NuGet can return a server.json instead of a vscode MCP.json)

Summary of changes:
- Switch to .NET 10 for `setup:dotnet` and make it work on Windows, this is for local use
- Use `actions/setup-dotnet` for setting up .NET on GitHub Actions, since this is the "blessed" way to do it, like we are for Python and node.js.
- Add tests for npm, pip, nuget, and docker package types
  - These pass locally but I may need some iterations with the CI
- Add telemetry events for MCP tool installation
  - This is the first time `PublicNonPersonalData` is used in the repo, but I think it's the best fit for package names based on my investigation of the vscode repo (these are similar to extension names I think)
- Switch NuGet implementation to using the .NET SDK, extract the server.json
  - I put my CLI executor in the `nuget.ts` file, but it could be moved to a shared place, or perhaps there's an existing implementation I should use.

Verified with:
- npm: `@azure/mcp` (old flow, no server.json)
- Python: `mcp-server-fetch` (old flow, no server.json)
- Docker: `mcp/node-code-sandbox` (old flow, no server.json)
- NuGet: `NuGet.Mcp.Server` (new flow), `Knapcode.SampleMcpServer` (new flow, this one has an input)

## Happy path (with server.json)

[Video.webm](https://github.com/user-attachments/assets/5685b27a-1e71-4e36-8807-d7351e3217ca)

## `dotnet` not found

[dotnet-not-found.webm](https://github.com/user-attachments/assets/232f0c59-ae7b-4c14-9e01-58b1f9a4b847)


src/extension/mcp/vscode-node/commands.ts
```diff
@@ -74,51 +92,123 @@ interface DockerHubResponse {
...
 		// we start doing the prompt in the background so the first call is speedy
 		const done = (async () => {
+
+			// if the package has a server manifest, we can fetch it and use it instead of a tool loop
+			if (pendingArgs.getServerManifest) {
+				let serverManifest;
+				try {
+					serverManifest = await pendingArgs.getServerManifest(canPrompt.p);
+				} catch (error) {
+					this.logService.warn(`Unable to fetch server manifest for ${validateArgs.type} package ${pendingArgs.name}@${pendingArgs.version}. Configuration will be generated from the package README.
```
kartheekp-ms, line 214: It looks like strings (error messages, log messages) are in English language in the [previous version](https://github.com/microsoft/vscode-copilot-chat/pull/546/files#diff-8fd1d0a38b63e50b726af382a240e95e0ab81067585a9f12c3d053ee5ea53952L90-L92) and in this PR. Is there a different process to localize the strings?

joelverhagen: It looks like there is a `localize` abstraction available in this repo. I hadn't looked until you mentioned it! I noticed it's more broadly used in the vscode repo, and not all the times in this vscode-copilot-chat extension repo.

joelverhagen: Log messages are generally not localized in this repo nor the vscode repo. User facing messages ARE however. I will do a pass to make sure we're localizing things properly in these modules I changed.

joelverhagen: Fixed.

joelverhagen:
@connor4312 - could you review this?
It is blocked on https://github.com/microsoft/vscode/pull/259634 but I think can be reviewed.

src/extension/mcp/vscode-node/nuget.ts
```diff
@@ -2,82 +2,367 @@
...
-	const serviceIndex = await serviceIndexResponse.json() as NuGetServiceIndexResponse;
-	const searchBaseUrl = serviceIndex.resources?.find(resource => resource['@type'] === 'SearchQueryService/3.5.0')?.['@id'];
-	if (!searchBaseUrl) {
-		return { state: 'error', error: `Package search URL not found in the NuGet.org registry service index` };
+	// dnx is used for running .NET MCP servers and it was shipped with .NET 10
+	const dotnetMajorVersion = parseInt(dotnetVersion.split('.')[0]);
+	if (dotnetMajorVersion < 10) {
+		return {
+			state: 'error',
+			error: localize("mcp.setup.badCommandVersion", "The installed .NET SDK must be version 10 or newer. Found {0}.", dotnetVersion),
```
joelverhagen:
```suggestion
			error: localize("mcp.setup.badDotnetSdkVersion", "The installed .NET SDK must be version 10 or newer. Found {0}.", dotnetVersion),
```

src/extension/mcp/vscode-node/commands.ts
```diff
@@ -215,7 +306,7 @@ export class McpSetupCommands extends Disposable {
 			} else if (args.type === 'pip') {
 				const response = await fetch(`https://pypi.org/pypi/${encodeURIComponent(args.name)}/json`);
 				if (!response.ok) {
-					return { state: 'error', error: `Package ${args.name} not found in PyPI registry` };
+					return { state: 'error', errorType: 'NotFound', error: localize("mcp.setup.pypiNotFound", "Package {0} not found in PyPI registry", args.name) };
```
joelverhagen:
```suggestion
					return { state: 'error', errorType: 'NotFound', error: localize("mcp.setup.pipNotFound", "Package {0} not found in PyPI registry", args.name) };
```

src/extension/mcp/test/vscode-node/commands.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+import { ILogService } from '../../../../platform/log/common/logService';
+import { createExtensionUnitTestingServices } from '../../../test/node/services';
+import { McpSetupCommands } from '../../vscode-node/commands';
+
+describe('get MCP server info', { timeout: 30_000 }, () => {
+	const testingServiceCollection = createExtensionUnitTestingServices();
+	const accessor = testingServiceCollection.createTestingAccessor();
+	const logService = accessor.get(ILogService);
+
+	it('npm returns package metadata', async () => {
```
connor4312, line 28: We shouldn't hit real services in our unit tests. We can record and store mock responses instead and inject them into tests.

src/extension/mcp/vscode-node/commands.ts
```diff
@@ -40,12 +43,28 @@ export interface IPendingSetupArgs {
 	name: string;
 	version?: string;
 	readme?: string;
+	getServerManifest?(installConsent: Promise<void>): Promise<any>;
 }
 
+type ValidatePackageErrorType = 'NotFound' | 'UnknownPackageType' | 'UnhandledError' | 'MissingCommand' | 'BadCommandVersion';
+type FlowFinalState = 'Done' | 'Failed' | 'NameMismatch';
```
connor4312: Prefer to use const enums for these

src/extension/mcp/vscode-node/nuget.ts
```diff
@@ -2,82 +2,367 @@
...
 }
 
-export async function getNuGetPackageMetadata(args: IValidatePackageArgs): Promise<ValidatePackageResult> {
-	// read the service index to find the search URL
-	// https://learn.microsoft.com/en-us/nuget/api/service-index
-	const serviceIndexUrl = `https://api.nuget.org/v3/index.json`;
-	const serviceIndexResponse = await fetch(serviceIndexUrl);
-	if (!serviceIndexResponse.ok) {
-		return { state: 'error', error: `Unable to load the NuGet.org registry service index (${serviceIndexUrl})` };
+export async function getNuGetPackageMetadata(id: string, logService: ILogService, dotnet: DotnetCli = { command: 'dotnet', args: [] }): Promise<ValidatePackageResult> {
```
connor4312: Are all the commands and flows here considered stable contract? I'm okayish adding them but I would not want to have to be in charge of updating them in each new dotnet/nuget version

joelverhagen: Yup, the directory structure of the packages dir and the `dotnet` commands we're using here have stable CLI interfaces.

src/extension/mcp/test/vscode-node/commands.spec.ts
joelverhagen: Is there a suite of tests I can move these to in order to test the integration? I'm wondering how we can detect if there are changes in the protocol.

connor4312: There is the risk of change, but 

1. We want to be able to reproduce old CI builds even if current live services change
1. In the past we've been broken by services going down to rate limiting at inopportune times. A random 3rd party service being unavailable should not break our builds

joelverhagen: Makes sense. Is there an "outer ring" that runs apart from builds than can tolerate these kind of error? Since the end user is hitting a live service and can be broken by an integration issue, it seems like something we should know about in an automated fashion ... somehow.

src/extension/mcp/vscode-node/commands.ts
joelverhagen: Fixed.

src/extension/mcp/test/vscode-node/commands.spec.ts
connor4312: We don't currently have such tests at the moment.

Btw, for doing the injection, we have an `IFetcherService` that you can inject in and then mock the `fetch` method of. I probably should have done that to begin with, but I was less familiar with networking code when I initially wrote this

joelverhagen: Okay, I can refactor it to use fetcher service for npm, PyPI, and Docker and then a local (disk-based) package source for `dotnet`, for testing purposes.

joelverhagen: Done.

connor4312 (approved):

aeschli (approved):

PR #554, opened 2025-08-11
Remove the message copilot_cache_control field from a context where there is no understanding of it

shaunm-msft:
Remove the message copilot_cache_control field if it exists when calling an OpenAI compatible endpoint where it might not be understood. The result of not understanding the field can be benign or result in a rejection.


roblourens (approved):

roblourens:
Thanks!

Tyriar (approved):

PR #567, opened 2025-08-12
Add hint about heredocs in fish

Tyriar:
Fixes microsoft/vscode#261222


Tyriar:
Seem to be hitting some prompt caching problem. This changes the `UserShellPrompt` part of the prompt, but it's always returning zsh here?

Screenshot 2025-08-12 at 6 06 12 am

Tyriar:
Confirmed it would be returning fish if it was up to date:

Screenshot 2025-08-12 at 6 09 13 am

PR #572, opened 2025-08-12
Support external cache layers

sbatten:
Allows passing a path to additional cache layers that will be applied on top of the cache layers in this repo.

### DIFF
--- a/test/base/cache.ts
+++ b/test/base/cache.ts
@@ -20,6 +20,12 @@ const decompress = promisify(zlib.brotliDecompress);
 
 const DefaultCachePath = process.env.VITEST ? path.resolve(__dirname, '..', 'simulation', 'cache') : path.resolve(__dirname, '..', 'test', 'simulation', 'cache');
 
+async function getGitRoot(cwd: string): Promise<string> {
+	const execAsync = promisify(exec);
+	const { stdout } = await execAsync('git rev-parse --show-toplevel', { cwd });
+	return stdout.trim();
+}
+
 export class Cache extends EventEmitter {
 	private static _Instance: Cache | undefined;
 	static get Instance() {
@@ -230,14 +236,14 @@ export class Cache extends EventEmitter {
 			this.activeLayer = (async () => {
 				const execAsync = promisify(exec);
 
-				const targetPath = this.externalLayersPath || this.layersPath;
+				const activeLayerPath = this.externalLayersPath ?? this.layersPath;
 				const gitStatusPath = this.externalLayersPath
-					? `${path.relative(await this._getGitRoot(targetPath), targetPath)}/*`
+					? `${path.relative(await getGitRoot(activeLayerPath), activeLayerPath)}/*`
 					: 'test/simulation/cache/layers/*';
 
 				// Check git for an uncommitted layer database file
 				try {
-					const gitRoot = await this._getGitRoot(targetPath);
+					const gitRoot = await getGitRoot(activeLayerPath);
 					const { stdout: statusStdout } = await execAsync(`git status -z ${gitStatusPath}`, { cwd: gitRoot });
 					if (statusStdout !== '') {
 						const layerDatabaseEntries = statusStdout.split('\0').filter(entry => entry.endsWith('.sqlite'));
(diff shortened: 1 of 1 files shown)


test/base/cache.ts
```diff
@@ -217,32 +230,46 @@ export class Cache extends EventEmitter {
 			this.activeLayer = (async () => {
 				const execAsync = promisify(exec);
 
+				const targetPath = this.externalLayersPath || this.layersPath;
```
joaomoreno:
```suggestion
				const targetPath = this.externalLayersPath ?? this.layersPath;
```

test/base/cache.ts
```diff
@@ -217,32 +230,46 @@ export class Cache extends EventEmitter {
...
+				const gitStatusPath = this.externalLayersPath
+					? `${path.relative(await this._getGitRoot(targetPath), targetPath)}/*`
+					: 'test/simulation/cache/layers/*';
```
joaomoreno: I'm confused. If this path is meant to be external... why are we doing git operations on it? Are people going to be checking these into the git repo?

Should we:

- Assume this path isn't within the git repo and avoid doing any git status calls on it?
- Use just a single external layer? Subsequent runs could use the previous external layer as read-only and populate a new external layer, keeping only the used keys.
- Alternatively, use a separate mechanism to control which file is active? Maybe keep a `.active-layer` file with a pointer?

joaomoreno: @sbatten says this the path lies inside a separate git repo. This makes sense to me.

test/base/cache.ts
```diff
@@ -217,32 +230,46 @@ export class Cache extends EventEmitter {
...
-
 				return activeLayer;
 			})();
 		}
 
 		return this.activeLayer;
 	}
 
+	private async _getGitRoot(cwd: string): Promise<string> {
+		const execAsync = promisify(exec);
```
joaomoreno: Let's extract this to top-level in the module.

test/base/cache.ts
```diff
@@ -217,32 +230,46 @@ export class Cache extends EventEmitter {
...
+						if (layerDatabaseEntries.length > 0) {
+							const regex = /([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\.sqlite$/;
+							const match = layerDatabaseEntries[0].match(regex);
+							if (match && this.layers.has(match[1])) {
+								return this.layers.get(match[1])!;
+							}
 						}
 					}
+				} catch (error) {
+					// If git operations fail, continue to create new layer
```
joaomoreno, line 259: Not sure. What if `externalLayersPath` isn't inside a git repo? Shouldn't we actually throw in that case?

test/base/cache.ts
```diff
@@ -217,32 +230,46 @@ export class Cache extends EventEmitter {
 			this.activeLayer = (async () => {
 				const execAsync = promisify(exec);
 
+				const targetPath = this.externalLayersPath ?? this.layersPath;
```
joaomoreno: Also should we rename `targetPath` to `activeLayersPath`?

test/base/simulationOptions.ts
```diff
@@ -180,6 +182,7 @@ export class SimulationOptions {
 			`  --n                                Run each scenario N times`,
 			`  --ci                               Equivalent to --n=${BASELINE_RUN_COUNT} but throws if the baseline is not up-to-date`,
 			`  --gc                               Used with --require-cache to compact cache layers into the baseline cache`,
+			`  --external-cache-layers-path       Used to specify the path to the external cache layers`,
```
joaomoreno, line 185:
```suggestion
			`  --external-cache-layers-path       Used to specify the path to the external cache layers (needs to be inside a git repository)`,
```

test/base/cache.ts
sbatten: The current approach supports someone creating a local folder for their tests they don't want to check in as well. Is there a reason you prefer to prevent that?

test/base/simulationOptions.ts
sbatten: See comment about preventing that case above.

test/base/cache.ts
sbatten: no strong feeling, but there can only be one "active layer" i.e. accepting writes, all layers in both paths are "active" underneath i.e. contribute entries.

PR #576, opened 2025-08-13
don't send random thinking tokens all the time

justschen:
cc. @karthiknadig


karthiknadig (dismissed):

karthiknadig:
@justschen You need to run vitest with update flag to update the snapshot

DonJayamanne (dismissed):

PR #577, opened 2025-08-13
Support for returning next edits from other cells

DonJayamanne:


DonJayamanne:
Will merge only when the new VS Code insiders is out and after we update the package.json as well.

src/extension/vscode.proposed.inlineCompletionsAdditions.d.ts
```diff
@@ -46,6 +46,10 @@ declare module 'vscode' {
 		action?: Command;
 
 		displayLocation?: InlineCompletionDisplayLocation;
+		/**
+		 * The Uri of the document this completion item is associated with.
```
DonJayamanne: Here's the related PR https://github.com/microsoft/vscode/pull/261376

src/extension/inlineEdits/vscode-node/inlineCompletionProvider.ts
```diff
@@ -35,11 +35,19 @@ import { learnMoreCommandId, learnMoreLink } from './inlineEditProviderFeature';
...
+	title: l10n.t('Learn More'),
+	command: learnMoreCommandId,
+	tooltip: learnMoreLink
+};
+
+interface NesCompletionItem extends InlineCompletionItem {
 	readonly telemetryBuilder: NextEditProviderTelemetryBuilder;
 	readonly info: NesCompletionInfo;
 	wasShown: boolean;
+	isEditInAnotherDocument?: boolean;
```
DonJayamanne, line 50: Only used internally, when we accept the completion, see line 424

src/extension/inlineEdits/vscode-node/inlineCompletionProvider.ts
```diff
@@ -303,6 +301,69 @@ export class InlineCompletionProviderImpl implements InlineCompletionItemProvide
...
+			isEditInAnotherDocument: true
+		};
+	}
+
+	private createComletionList(
+		doc: IVSCodeObservableDocument,
+		document: TextDocument,
+		position: Position,
+		range: Range,
+		result: NonNullable<(NextEditResult | DiagnosticsNextEditResult)['result']>,
```
DonJayamanne, line 337: Preferred Omitting so that most of the common stuff is left out, else too many args and the like.
This function will build only whats required.

DonJayamanne:
@hediet I've updated this based on our last discussion.
I will update the code to include DisplayLocationKind in a separate PR.

src/extension/inlineEdits/vscode-node/inlineCompletionProvider.ts
```diff
@@ -213,64 +223,43 @@ export class InlineCompletionProviderImpl implements InlineCompletionItemProvide
...
+			if (!documents.length) {
+				tracer.trace('no next edit suggestion');
+			} else if (documents[0][0] === document) {
+				// nes is for this same document.
+				range = documents[0][1];
+				const allowInlineCompletions = this.model.inlineEditsInlineCompletionsEnabled.get();
+				isInlineCompletion = allowInlineCompletions && isInlineSuggestion(position, document, range, result.edit.newText);
+				completionItem = serveAsCompletionsProvider && !isInlineCompletion ?
+					undefined :
+					this.createComletionList(doc, document, position, range, result);
```
hediet: This I don't understand. You assign a list to an item? Also, there is a typo in `createComletionList` ;)

src/extension/inlineEdits/vscode-node/inlineCompletionProvider.ts
```diff
@@ -213,64 +223,43 @@ export class InlineCompletionProviderImpl implements InlineCompletionItemProvide
...
+			let isInlineCompletion: boolean = false;
+			let completionItem: Omit<NesCompletionItem, 'telemetryBuilder' | 'info' | 'showInlineEditMenu' | 'action' | 'wasShown'> | undefined;
+
+			const documents = doc.fromOffsetRange(result.edit.replaceRange);
+			if (!documents.length) {
+				tracer.trace('no next edit suggestion');
+			} else if (documents[0][0] === document) {
+				// nes is for this same document.
+				range = documents[0][1];
+				const allowInlineCompletions = this.model.inlineEditsInlineCompletionsEnabled.get();
```
hediet, line 236: This is a new code duplication. Is there a way to avoid that?

DonJayamanne: Do you mean in line 355. i don't want to pass that into it as an argument, doesn't feel right.
I.e. just passing the parameter `allowInlineCompletions` feels like a weird argument when all we want is to `create a completion item`
Happy to pass that, if you want.

DonJayamanne: The method createCompletionItem should determine whether `isLineEdit` is true or false, I don't think the calling method should be able to pass that in
I guess another way is to ensure `isInlineEdit` is never returned by the method but initialized in the calling code.

DonJayamanne: Will do that.

PR #590, opened 2025-08-13
Revised context menu

eli-w-king:
Addresses: https://github.com/microsoft/vscode/issues/255962 with the following changes:
1. Renamed 'Review and Comment' to 'Code Review' to match change made in SCM panel.
2. Removed 'Explain' from nested menu.

Changes made in vscode: https://github.com/microsoft/vscode/pull/261547
1. Renamed 'Inline Editor Chat' and 'Inline Terminal Chat' to 'Open Inline Chat'.
2. 'Add Selection to Chat' only shows when text is highlighted in editor, this follows the terminal behavior.
3. Added 'Explain' from nested menu to inline item.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2796,29 +2796,24 @@
 			],
 			"editor/context/chat": [
 				{
-					"command": "github.copilot.chat.explain",
-					"when": "!github.copilot.interactiveSession.disabled",
-					"group": "copilotAction@1"
-				},
-				{
 					"command": "github.copilot.chat.fix",
 					"when": "!github.copilot.interactiveSession.disabled && !editorReadonly",
-					"group": "copilotAction@2"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2795,30 +2795,25 @@
 				}
 			],
 			"editor/context/chat": [
```
bpasero, line 2797: Related to my feedback in https://github.com/microsoft/vscode/pull/261547#discussion_r2276480621, we cannot really remove the "Explain" entry from the extension unless it is implemented by core properly, so this PR depends on the resolution in core.

bpasero (changes requested):

bpasero:
Continues in https://github.com/microsoft/vscode-copilot-chat/pulls

PR #573, opened 2025-08-12
Add multi_replace_string tool for parallel replace_string edits

yemohyleyemohyle:

PR #593, opened 2025-08-13
tools: add multi-replace-string tool

connor4312:
Builds on https://github.com/microsoft/vscode-copilot-chat/pull/573, reopening here due to LFS issues in forks 😵‍💫 

This adds a multi-replace-string tool which is enabled via a setting `"github.copilot.chat.advanced.multiReplaceString.enabled": true,`, or via EXP. Benchmarks show this reduces turn count and cost (increasing speed) without reducing benchmark score. This PR includes:

1. Prompt changes to reference the new tool when it's available (existing prompts are unchanged when the tool is not enabled)
2. Enablement of the tool based on the setting/EXP
3. A cleanup of the old Gemini replace_string experiment which is now at 100%
4. Some refactor of replace_string to enable its use in multi_replace_string

The tool previously just returned a success or error for an edit. We now prepare edits ahead of time, and in the multi case we merge them (so that workspace and notebook edits get applied in the right order) and then apply them. Edit telemetry is generally the same but it gets a new `isMulti` boolean that indicates whether it came from a multi-string-replace call, and there is a new event for the multi tool as well.


TylerLeonhardt (approved):

src/extension/tools/node/multiReplaceStringTool.tsx
```diff
@@ -0,0 +1,143 @@
...
+			"multiStringReplaceCall" : {
+				"owner": "connor4312",
+				"comment": "Tracks how much percent of the AI edits survived after 5 minutes of accepting",
+				"requestId": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The id of the current request turn." },
+				"model": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The model used for the request." },
+				"successes": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The number of successful edits.", "isMeasurement": true },
+				"failures": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The number of failed edits.", "isMeasurement": true },
+				"uniqueUris": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The number of unique URIs edited.", "isMeasurement": true },
+				"individualEdits": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The number of individual text edits made.", "isMeasurement": true }
+			}
```
OrenMe, line 54: @connor4312 did u mean to leave this here?

PR #595, opened 2025-08-14
Unify embeddings paths

mjbvz:
Our code has two embeddings paths for historical reasons:

- The capi api based one based around `IEmbeddingEndpoint`. This only supports text3small. This is used everywhere except `#codebase` and is better supported for our test infrastructure 

- The new github embeddings endpoint. This one supports additional embedding types and options. However it is currently only used in `#codebase`

This change switches everything to use the new github embeddings endpoint instead

cc @lramos15 Checking to see if this raises any concerns with you. I've already checked in with GitHub about using the new embeddings endpoint for these call paths

### DIFF
--- a/src/extension/test/node/services.ts
+++ b/src/extension/test/node/services.ts
@@ -59,7 +59,6 @@ export function createExtensionUnitTestingServices(currentTestRunInfo?: any, mod
 		new SyncDescriptor(TestEndpointProvider, [
 			modelConfig?.smartChatModel ?? modelConfig?.chatModel,
 			modelConfig?.fastChatModel ?? modelConfig?.chatModel,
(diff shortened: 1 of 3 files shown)


DonJayamanne (approved):

lramos15 (approved):
Looks fine, so just to confirm how will this work if the embeddings model changes in the future. Since we're no longer using CAPI's models.

Should we also remove from the RequestType.Embeddings from the node module?

mjbvz:
Thanks @lramos15. This endpoint should support multiple embedding models so old clients will still work when we roll out newer models. GitHub also has an api to mark some models as preferred or deprecated. I’m using that for workspace but not yet for all of these other users. Will explore adopting that for everyone

PR #607, opened 2025-08-14
Add command executor abstraction, only run dotnet CLI tests on CI

joelverhagen:
Address https://github.com/microsoft/vscode/issues/261524.
Uses CI + BUILD_ARTIFACTSTAGINGDIRECTORY similar to https://github.com/microsoft/vscode-copilot-chat/blob/e42aea64c5a9608c018c66aac9e6c784f3ff2517/src/util/vs/base/common/platform.ts#L82.


joelverhagen:
@connor4312 - is this one you would like to review?

PR #608, opened 2025-08-14
Update to consume new API

lramos15:
Extension side of https://github.com/microsoft/vscode/pull/261645


roblourens (approved):

src/extension/vscode.proposed.chatProvider.d.ts
```diff
@@ -5,14 +5,6 @@
 
 declare module 'vscode' {
```
roblourens, line 6: You should set the api version too for changes like this. Although note that going from no version to a version means that old extensions would still be considered compatible.

lramos15: I'm not super well versed in how the new api version stuff works. I still bump the engine then too?

roblourens: Pushed a couple PRs to add api versions. Engine version is fine

PR #619, opened 2025-08-15
Fix incorrect model switching on free models (#480)

lramos15:
Candidate PR of #480


oggynjack:
I have same issue i am also unable to use GPT-5 mini and GPT-4o when out of quota.

Ebaoone (approved):
Ebaone

PR #620, opened 2025-08-15
Update ExP Service to have user account based filters

sbatten:
+ filter by internal
+ filter by sku

This PR also greatly reduces the service API to simplify things. We also refresh treatments every 1 hour. 

@sandy081 I'd love to get your feedback as this is related to the work you did in core and it affects configurations.


src/platform/telemetry/common/nullExperimentationService.ts
```diff
@@ -57,45 +36,20 @@ export interface IExperimentationService {
...
-	isFlightEnabledAsync(_flight: string): Promise<boolean> {
-		return Promise.resolve(false);
-	}
+	private readonly _onDidTreatmentsChange = new Emitter<void>();
+	readonly onDidTreatmentsChange = this._onDidTreatmentsChange.event;
 
+	async hasTreatments(): Promise<void> { return Promise.resolve(); }
+	async hasAccountBasedTreatments(): Promise<void> { return Promise.resolve(); }
 	getTreatmentVariable<T extends boolean | number | string>(_configId: string, _name: string): T | undefined {
+		console.log('[ALERT!ExP] NullExperimentationService: getTreatment', _configId, _name);
```
lramos15: We probably don't want to keep this do we?

sbatten: no I'm keeping this while this is in draft as I'm using them to debug

lramos15 (approved):
Looks great!

lramos15 (dismissed):

src/platform/telemetry/common/nullExperimentationService.ts
```diff
@@ -3,52 +3,31 @@
...
-
 	readonly _serviceBrand: undefined;
 
 	/**
-	 * Promise indicating that the experimentation service has been
-	 * initialized, so it's safe to make a call to isFlightEnabled.
+	 * Emitted whenever a treatment values have changes based on user account information or refresh.
 	 */
-	readonly initializePromise: Promise<void>;
+	onDidTreatmentsChange: Event<void>;
```
sandy081: Looks like this event is not used anywhere and it might not be true that treatments have changed on refresh, so this event will be triggered falsely.

sbatten: @sandy081 It is true, the exp package doesn't actually tell us this information, so in order to be really accurate we'd have to keep a list of every queried treatment and a cache of its value and if they are not equivalent, we could fire a more accurate event. However, I don't know how valuable it will be, most consumers can just check the value for free when the event signals and decide if something should happen. The primary consumer of this would be the configuration service, which I tried hooking up, but got into a dependency cycle. Since we don't do that refresh today, I'd like to take that as a second PR.

sandy081: Ok. That is the reason I thought having this change event is misleading. In VS Code core, I ask for settings related experiments every 60 minutes instead of adding a change event in experiment service.

https://github.com/microsoft/vscode/blob/143bd85516a97ccca08818d6b04cd8ddd8997475/src/vs/workbench/services/configuration/browser/configurationService.ts#L1355

So, I suggest to do the same here.

sbatten: Although config service will be the primary consumer, I won't necessarily be the only consumer as some ExPs don't have settings. Also, we have a separate consideration now introduced in the PR. Your authentication state can cause the treatments to change, not just upon a general refresh. I plan to work with you to bring this behavior to core as well.

sandy081: But I would not prefer to use the change event if it is not sure about the values being changed or not.

sandy081 (approved):

src/platform/telemetry/common/nullExperimentationService.ts
sbatten: As discussed offline, I've updated the treatments change event to emit a list of known changed keys (ones that have been queried previously).

lramos15 (approved):

PR #644, opened 2025-08-18
Repo context for commit message

steinfurt:
Fixes microsoft/vscode#256731
It also addresses most of microsoft/vscode#238255

### DIFF
--- a/src/extension/prompt/common/repository.ts
+++ b/src/extension/prompt/common/repository.ts
@@ -7,3 +7,11 @@ export interface RecentCommitMessages {
 	readonly repository: string[];
 	readonly user: string[];
 }
+
+export interface GitCommitRepoContext {
+	readonly repositoryName?: string;
+	readonly owner?: string;
+	readonly headBranchName?: string;
+	readonly defaultBranch?: string;
+	readonly pullRequest?: { title: string; url: string };
+}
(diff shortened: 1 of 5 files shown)


steinfurt:

steinfurt:
Hello @lszomoru,

I hope you don't mind me asking here about the status of this PR. Obviously, I know one has to be patient when submitting an open source contribution. You've probably had more than enough other topics to take care of in recent weeks. Still, I wanted to quickly check in if there is anything missing, anything to be changed here or if you've got any other kind of feedback.

I originally believed this PR to almost be a no-brainer, given that the linked issue has been created by a VS code team member, there is another issue with quite a few upvotes, the missing implementation has been discussed in the AMA of the v1.102 release and this just adds the suggested implementation in a minimal way (though not re-using the existing component).

I don't want to be impolite, but after a month I thought it would be okay to ask for any kind of reaction... Obviously, no problem if it takes a little longer...

lszomoru:
@steinfurt, sorry for not getting back to you on this until now. The latest Copilot Chat pre-release extension contains the repository/branch name in the context. Could you please try to use those in custom instructions and let me know if it works as expected? Thanks!

steinfurt:
@lszomoru, yes, sure. For reference, I just had a look and can see you implemented a simplified variant in #1756.

In principle, it works fine. I also quickly tested adding a few commits with a different ticket reference from the one in the branch name to check if the relative priority between the `'repository-context'` and `'user-commits'` parts of the prompt are good. Using `priority={850}` looks about right to me.

But I've also seen these custom instructions can be overwhelmed by too much other context in the prompt, e.g. when there are large code changes. But for that to work properly one would probably have to do a bigger change like switching the model from `gpt-4o-mini` (which will happen some day anyway, I guess) or do a (slow) two-step prompt "summarize + format" which I don't think is worth it yet.

Regarding other parts of the prompt which you left out: I don't mind and only added them, because they were referenced [here](https://github.com/microsoft/vscode/issues/256731#issuecomment-3090247795). For me, personally, "branch name" is enough. Thanks!

lszomoru:
@steinfurt, thanks for looking at the code. Yes, we are planning on switching to a new model in the near future.

PR #645, opened 2025-08-18
[WIP] Enable more features during Scenario Automation

rwoll:
During Scenario Automation more of the extension must be allowed to run. It's therefore somewhere between `ExtensionMode.Production` and `ExtensionMode.Test`.

During local development the extension automation is run under `ExtensionMode.Test`, but when testing different versions of Marketplace-downloaded releases, `ExtensionMode.Production` is set.

Instead of inlining `ExensionMode.*` and `isScenarioAutomation` checks throughout the whole codebase making it difficult to understand which services are enabled in what combinations of modes, this PR co-locates these checks into `EnvServiceImpl`. It wraps the logic in more specific function names dependent on where/what is being guarded.

Suggested followup: custom ESLint Rule that bans raw `ExtensionMode.*` checks outside safelisted files.

### DIFF
--- a/src/extension/extension/vscode-node/services.ts
+++ b/src/extension/extension/vscode-node/services.ts
@@ -147,7 +147,7 @@ export function registerServices(builder: IInstantiationServiceBuilder, extensio
 		builder.define(ICopilotTokenManager, getOrCreateTestingCopilotTokenManager());
 	}
 
-	if (envService.isScenarioAutomation()) {
+	if (envService.useStaticGitHubAuthenticationService()) {
 		builder.define(IAuthenticationService, new SyncDescriptor(StaticGitHubAuthenticationService, [getStaticGitHubToken]));
 	}
 	else {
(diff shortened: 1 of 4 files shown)


src/extension/extension/vscode-node/services.ts
```diff
@@ -198,18 +200,18 @@ export function registerServices(builder: IInstantiationServiceBuilder, extensio
...
 		builder.define(IExperimentationService, new SyncDescriptor(MicrosoftExperimentationService));
 	} else {
 		builder.define(IExperimentationService, new NullExperimentationService());
 	}
 }
 
-function setupTelemetry(builder: IInstantiationServiceBuilder, extensionContext: ExtensionContext, internalAIKey: string, internalLargeEventAIKey: string, externalAIKey: string) {
+function setupTelemetry(builder: IInstantiationServiceBuilder, extensionContext: ExtensionContext, internalAIKey: string, internalLargeEventAIKey: string, externalAIKey: string, envService: IEnvService) {
 
-	if (ExtensionMode.Production === extensionContext.extensionMode) {
```
rwoll, line 212: Without refactor we would need to write something like:

```
ExtensionMode.Production === extensionContext.extensionMode && !isScenarioAutomation
```

PR #651, opened 2025-08-18
don't use url for anthropic models

justschen:
ref https://github.com/microsoft/vscode-copilot-issues/issues/194

this is a temporary solution while capi resolves their issue


src/extension/prompts/node/panel/toolCalling.tsx
```diff
@@ -406,7 +406,11 @@ class PrimitiveToolResult<T extends IPrimitiveToolResultProps> extends PromptEle
 		const uploadsEnabled = this.configurationService && this.experimentationService
 			? this.configurationService.getExperimentBasedConfig(ConfigKey.Internal.EnableChatImageUpload, this.experimentationService)
 			: false;
-		const effectiveToken = uploadsEnabled ? githubToken : undefined;
+
+		// Anthropic (from CAPI) currently does not support image uploads from tool calls.
+		const isAnthropic = this.endpoint.family.startsWith('claude') || this.endpoint.family.startsWith('Anthropic');
```
DonJayamanne: You might want to add a function into `chatModelCapabilities.ts` as thats where we store model specific capabilities

PR #653, opened 2025-08-18
allow more services in Scenario Mode

rwoll:
When running in Scenario Automation (distinct from `stest` mode and set via an env var), as much production code as possible should be used so the real user experience is measured. However, certain codepaths — like telemetry — should be disabled regardless of the formal VSCode `ExtensionMode`, and certain codepaths (like full OAuth flow) must not be run.

When running local builds, `extensionMode` is `ExtensionMode.Test` while Marketplace downloaded builds are `ExtensionMode.Production` so relying solely on `extensionMode` is not possible.

[An abandoned iteration of this PR](https://github.com/microsoft/vscode-copilot-chat/pull/645/files#diff-b4875421464d8c9a34b81c9d2711895f33ed7a9df2e3635207f31f851af93537R82) attempted to colocate all this logic within one central location but I found that made the callsites that dependend on that logic particularly hard to follow, hence this PR proposes a less DRY, but more readable approach.

### DIFF
--- a/package.json
+++ b/package.json
@@ -796,7 +796,7 @@
 				"name": "copilot_editNotebook",
 				"toolReferenceName": "editNotebook",
 				"displayName": "%copilot.tools.editNotebook.name%",
-				"modelDescription": "This is a tool for editing an existing Notebook file in the workspace. Generate the \"explanation\" property first.\nThe system is very smart and can understand how to apply your edits to the notebooks.\nWhen updating the content of an existing cell, ensure newCode preserves whitespace and indentation exactly and does NOT include any code markers such as (...existing code...).",
+				"modelDescription": "This is a tool for editing an existing Notebook file in the workspace. Generate the \"explanation\" property first.\nThe system is very smart and can understand how to apply your edits to the notebooks.\nWhen updating the content of an existing cell, ensure newCode preserves whitespace and indentation exactly and does NOT include any code markers such as (...existing code...). Do NOT use this to read the notebook file or the contents of cell(s). To read contents of a specific cell use the copilot_getNotebookSummary followed by the read_file tool.",
 				"tags": [
 					"enable_other_tool_copilot_getNotebookSummary"
 				],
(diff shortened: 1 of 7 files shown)


src/extension/onboardDebug/vscode-node/copilotDebugCommandContribution.ts
```diff
@@ -186,7 +187,7 @@ export class CopilotDebugCommandContribution extends Disposable implements vscod
 	}
 
 	private getVersionNonce() {
-		if (this.context.extensionMode !== vscode.ExtensionMode.Production) {
+		if (this.context.extensionMode !== vscode.ExtensionMode.Production && !isScenarioAutomation) {
```
connor4312: This one is fine to keep as-is

```suggestion
		if (this.context.extensionMode !== vscode.ExtensionMode.Production) {
```

src/extension/prompts/node/panel/definitionAtPosition.tsx
```diff
@@ -78,7 +79,7 @@ export class DefinitionAtPosition extends PromptElement<Props, State> {
 			return { k: 'ignored' };
 		}
 
-		const timeout = this._vscodeExtensionCtxService.extensionMode === ExtensionMode.Test
+		const timeout = this._vscodeExtensionCtxService.extensionMode === ExtensionMode.Test && !isScenarioAutomation
```
connor4312, line 82:
```suggestion
		const timeout = this._vscodeExtensionCtxService.extensionMode === ExtensionMode.Test
```

I think we still want a timeout here. No timeout is good for debugging but could stall real agent runs

src/extension/prompts/node/panel/referencesAtPosition.tsx
```diff
@@ -64,7 +65,7 @@ export class ReferencesAtPosition extends PromptElement<Props> {
 			return <ignoredFiles value={[this.props.document.uri]} />;
 		}
 
-		const timeout = this.extensionContext.extensionMode === ExtensionMode.Test
+		const timeout = this.extensionContext.extensionMode === ExtensionMode.Test && !isScenarioAutomation
```
connor4312, line 68: same

src/extension/prompts/node/panel/symbolDefinitions.tsx
```diff
@@ -82,7 +83,7 @@ export class SymbolDefinitions extends PromptElement<Props, State> {
 			return { ...emptyState, isIgnored: true };
 		}
 
-		const timeout = this.extensionContext.extensionMode === ExtensionMode.Test
+		const timeout = this.extensionContext.extensionMode === ExtensionMode.Test && !isScenarioAutomation
```
connor4312, line 86: same

src/extension/prompts/node/panel/definitionAtPosition.tsx
rwoll: `0` means no timeout, so my change is ensuring that when in Scenario Automation mode, it's the non-zero, production option.

src/extension/prompts/node/panel/referencesAtPosition.tsx
rwoll: See related comment.

PR #661, opened 2025-08-19
move function calls outside of jsx

meganrogge:
fix #262243

Not sure why, but this seems to fix a bug for me

### DIFF
--- a/src/extension/prompts/node/base/terminalAndTaskState.tsx
+++ b/src/extension/prompts/node/base/terminalAndTaskState.tsx
@@ -117,11 +117,12 @@ export class TerminalAndTaskStatePromptElement extends PromptElement<TerminalAnd
 					)}
 				</>
 			);
-
+			const renderedTasks = renderTasks();
+			const renderedTerminals = renderTerminals();
 			return (
 				<>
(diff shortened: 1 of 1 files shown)


src/extension/prompts/node/base/terminalAndTaskState.tsx
```diff
@@ -117,11 +117,12 @@ export class TerminalAndTaskStatePromptElement extends PromptElement<TerminalAnd
 					)}
 				</>
 			);
-
+			const renderedTasks = renderTasks();
```
roblourens, line 120: How would this fix the bug? The _data_ (eg `resultTasks`) has already been computed by this point, and the problem is with collecting the tasks and terminals, no?

meganrogge: Idk - I was seeing a strange result before this change and not seeing it after

roblourens: This code is equivalent as far as I can tell. I'm sure there's an element of randomness here

meganrogge:
This change can't hurt, so keeping it since it seemed to improve things for me. Pls reopen the linked issue @roblourens if you see it again though

PR #672, opened 2025-08-19
Feature: add JSON export of chat prompt logs in chat debug view

zhichli:
Add a new feature to allow exporting logs of individual chat prompt node or all nodes in the copilot debug view in JSON. 

Usage: 
- offline evaluation and trajectory transformation
- dependency of WIP chat replay feature


src/extension/log/vscode-node/requestLogTree.ts
```diff
@@ -22,20 +22,80 @@ import { IExtensionContribution } from '../../common/contributions';
...
+			if (!treeItem || !treeItem.children) {
+				return [];
+			}
+
+			const logEntries = treeItem.children.map(child => {
+				if (child instanceof ChatRequestItem || child instanceof ToolCallItem || child instanceof ChatElementItem) {
+					return child.info;
+				}
+				return undefined; // Skip non-loggable items
+			}).filter((entry): entry is LoggedInfo => !!entry);
```
connor4312: we have an `isDefined` helper

```suggestion
			}).filter(isDefined);
```

src/extension/log/vscode-node/requestLogTree.ts
```diff
@@ -22,20 +22,80 @@ import { IExtensionContribution } from '../../common/contributions';
...
+
+			const promptLogs: any[] = [];
+
+			// Process each log entry and get its JSON content using the content provider
+			for (const logEntry of logEntries) {
+				try {
+					// Get the JSON content using the virtual document URI
+					const virtualUri = vscode.Uri.parse(ChatRequestScheme.buildUri({ kind: 'request', id: logEntry.id }, 'json'));
+					const document = await vscode.workspace.openTextDocument(virtualUri);
+					const jsonContent = document.getText();
```
connor4312: You can use `vscode.workspace.fs.readFile` with the URI rather than opening a document to get the contents. Or go directly to the IRequestLogger with a new method. Reading a file to parse its JSON is sort of roundabout.

src/extension/log/vscode-node/requestLogTree.ts
```diff
@@ -306,6 +357,144 @@ export class RequestLogTree extends Disposable implements IExtensionContribution
...
+				await vscode.workspace.fs.writeFile(saveUri, Buffer.from(finalContent, 'utf8'));
+
+				// Show success message with option to reveal the file
+				const revealAction = 'Reveal in Explorer';
+				const openAction = 'Open File';
+				const result = await vscode.window.showInformationMessage(
+					`Successfully exported prompt with ${promptObject.logCount} log entries to ${saveUri.fsPath}`,
+					revealAction,
+					openAction
+				);
```
connor4312, line 393: This notification is extraneous since we show the file after we save it anyway, I would avoid the extra noise

```suggestion
```

src/extension/log/vscode-node/requestLogTree.ts
```diff
@@ -306,6 +357,144 @@ export class RequestLogTree extends Disposable implements IExtensionContribution
...
+				await vscode.workspace.fs.writeFile(saveUri, Buffer.from(finalContent, 'utf8'));
+
+				// Show success message with option to reveal the file
+				const revealAction = 'Reveal in Explorer';
+				const openAction = 'Open File';
+				const result = await vscode.window.showInformationMessage(
+					`Successfully exported ${allPromptsContent.length} prompts with ${totalLogEntries} log entries to ${saveUri.fsPath}`,
+					revealAction,
+					openAction
+				);
```
connor4312, line 480: Similarly here

```suggestion
```

src/extension/prompt/vscode-node/requestLoggerImpl.ts
```diff
@@ -24,6 +24,104 @@ import { safeStringify } from '../../../util/vs/base/common/objects';
...
+		public readonly id: string,
+		public readonly name: string,
+		public readonly args: unknown,
+		public readonly response: LanguageModelToolResult2,
+		public readonly chatRequest: any | undefined,
+		public readonly time: number,
+		public readonly thinking?: ThinkingData
+	) { }
+
+	async toJson(): Promise<any> {
```
connor4312: I suggest naming this `toJSON` -- the default JSON stringifier actually looks for that method name and will use that to stringify any object that implements it. Then below you can just `JSON.stringify(entry)` directly

src/extension/prompt/vscode-node/requestLoggerImpl.ts
```diff
@@ -24,6 +24,104 @@ import { safeStringify } from '../../../util/vs/base/common/objects';
...
+		const result: string[] = [];
+		for (const content of this.response.content as (LanguageModelTextPart | LanguageModelPromptTsxPart)[]) {
```
connor4312: Tools can also return `LanguageModelDataPart` which we should handle here (e.g. images, most commonly returned by MCP tools today)

src/extension/prompt/vscode-node/requestLoggerImpl.ts
```diff
@@ -24,6 +24,104 @@ import { safeStringify } from '../../../util/vs/base/common/objects';
...
+class LoggedElementInfo implements ILoggedElementInfo {
+	public readonly kind = LoggedInfoKind.Element;
+
+	constructor(
+		public readonly id: string,
+		public readonly name: string,
+		public readonly tokens: number,
+		public readonly maxTokens: number,
+		public readonly trace: HTMLTracer,
+		public readonly chatRequest: any | undefined
```
connor4312: Prefer referencing `ChatRequest` instead of `any` here

connor4312 (approved):
Looks sane overall, some suggestions

src/platform/requestLogger/node/requestLogger.ts
```diff
@@ -77,13 +91,15 @@ export interface ILoggedElementInfo {
 	maxTokens: number;
 	trace: HTMLTracer;
 	chatRequest: ChatRequest | undefined;
+	toJson(): any;
```
amunger: can you just use `object` instead of `any` for all these return types? (in the impl file as well)

DonJayamanne (approved):

src/extension/log/vscode-node/requestLogTree.ts
zhichli: I'm not sure which code you mean that does the "show the file after we save". Without this notification, there is no feedback in ui whether the file is saved or where to find it. I also think real action is useful besides openaction.

connor4312: Oh, I suppose only if `result === revealAction` or `openAction`. We could display the notification only if it's not one of those

connor4312 (approved):

src/extension/log/vscode-node/requestLogTree.ts
zhichli: are you assuming the file is auto-opened after the `vscode.workspace.fs.writeFile` call ? As tested, the current ux is when the user clicks the `export as json` button, it will write/save the file under the scene and won't open the file saved, then the notification comes in as the only way to render those 2 actions user can subsequently take (including open and see the file). So I'm still confused here in what scenarios we don't need to show notification 🥲 - pardon me if I misunderstood anything.

PR #678, opened 2025-08-19
Ungroup tools based on embedding rankings

vritant24:
Add embeddings-based tool ranking and caching for virtual tool grouper to predict which tools are most relevant to user queries, improving the tool selection experience.

The feature is currently behind an experiment flag (`chat.virtualTools.embeddingRanking`)


vritant24:
cc @connor4312

src/extension/prompt/node/defaultIntentRequestHandler.ts
```diff
@@ -585,7 +585,7 @@ class DefaultToolCallingLoop extends ToolCallingLoop<IDefaultToolLoopOptions> {
 		const token = CancellationToken.None;
 		const allTools = await this.options.invocation.getAvailableTools?.() ?? [];
 		const grouping = this.toolGroupingService.create(this.options.conversation.sessionId, allTools);
-		const computed = await grouping.compute(token);
+		const computed = await grouping.compute(delta.text, token);
```
connor4312: This should use


```suggestion
		const computed = await grouping.compute(this.options.request.prompt, token);
```

src/extension/intents/node/toolCallingLoop.ts
```diff
@@ -121,7 +121,7 @@ export abstract class ToolCallingLoop<TOptions extends IToolCallingLoopOptions =
 	protected abstract buildPrompt(buildPromptContext: IBuildPromptContext, progress: Progress<ChatResponseReferencePart | ChatResponseProgressPart>, token: CancellationToken): Promise<IBuildPromptResult>;
 
 	/** Gets the tools that should be callable by the model. */
-	protected abstract getAvailableTools(outputStream: ChatResponseStream | undefined, token: CancellationToken): Promise<LanguageModelToolInformation[]>;
+	protected abstract getAvailableTools(query: string, outputStream: ChatResponseStream | undefined, token: CancellationToken): Promise<LanguageModelToolInformation[]>;
```
connor4312: you don't really need another arg here, you already have the query on `this.options.request.prompt`

src/extension/tools/common/test/toolEmbeddingsCache.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { beforeEach, describe, expect, it, Mock, vi } from 'vitest';
+import { Embedding, EmbeddingType, rankEmbeddings } from '../../../../platform/embeddings/common/embeddingsComputer';
+import { TestLogService } from '../../../../platform/testing/common/testLogService';
+import { CancellationToken } from '../../../../util/vs/base/common/cancellation';
+import { ToolEmbeddingsComputer } from '../virtualTools/toolEmbeddingsCache';
+
+vi.mock('../../../../platform/embeddings/common/embeddingsComputer', async (importActual) => {
```
connor4312: Prefer to use service mocks rather than trying to mock modules like this.

src/extension/tools/common/virtualTools/toolEmbeddingsCache.ts
```diff
@@ -0,0 +1,182 @@
...
+import { CancellationToken } from '../../../../util/vs/base/common/cancellation';
+import { IInstantiationService } from '../../../../util/vs/platform/instantiation/common/instantiation';
+
+export const EMBEDDING_TYPE_FOR_TOOL_GROUPING = EmbeddingType.text3small_512;
+
+export class PreComputedToolEmbeddingsCache {
+	private readonly cache: IEmbeddingsCache;
+	private embeddingsMap: Map<string, Embedding> | undefined;
+
+	constructor(instantiationService: IInstantiationService, envService: IEnvService, private readonly _logService: ILogService) {
```
connor4312: You can dependency-inject all of these via `@Decorators` and then use `instantiatioService.creeatInstance` to create the class and fill them in

src/extension/tools/common/virtualTools/toolEmbeddingsCache.ts
```diff
@@ -0,0 +1,182 @@
...
+		if (token.isCancellationRequested) {
+			return;
+		}
+
+		const missingTools: string[] = [];
+		toolNames.forEach(t => {
+			if (!this.embeddingsStore.has(t)) {
+				missingTools.push(t);
+			}
+		});
```
connor4312:
```suggestion
		const missingTools = [...toolNames].filter(t => !this.embeddingsStore.has(t));
```

src/extension/tools/test/node/virtualTools/virtualToolGrouper.spec.ts
```diff
@@ -536,24 +536,267 @@ describe('Virtual Tools - Grouper', () => {
...
+			mockExpService = {};
+			mockConfigurationService = {
+				getExperimentBasedConfig: vi.fn().mockReturnValue(true)
+			};
+
+			// Replace the dependencies in the grouper instance
+			(grouper as any).embeddingsComputer = mockEmbeddingsComputer;
+			(grouper as any).toolEmbeddingsComputer = mockToolEmbeddingsComputer;
+			(grouper as any)._configurationService = mockConfigurationService;
+			(grouper as any)._expService = mockExpService;
```
connor4312: Use dependency injection or add methods/overrides to `TestVirtualToolGrouper` if you need to instead of overwriting private properties with `any` casts

src/extension/tools/common/virtualTools/virtualToolGrouper.ts
```diff
@@ -24,15 +30,24 @@ const SUMMARY_PREFIX = 'Call this tool when you need access to a new category of
...
 		@IToolGroupingCache private readonly _cache: IToolGroupingCache,
 		@ITelemetryService private readonly _telemetryService: ITelemetryService,
 		@ILogService private readonly _logService: ILogService,
+		@IEmbeddingsComputer private readonly embeddingsComputer: IEmbeddingsComputer,
+		@IConfigurationService private readonly _configurationService: IConfigurationService,
+		@IExperimentationService private readonly _expService: IExperimentationService,
+		@IInstantiationService _instantiationService: IInstantiationService,
+		@IEnvService envService: IEnvService,
 	) {
+		const embeddingsCache = new PreComputedToolEmbeddingsCache(_instantiationService, envService, this._logService);
```
connor4312: If the VirtualToolGrouper itself doesn't need to care about this, just instantiate the embeddingsCache within the ToolEmbeddingsComputer

src/extension/tools/common/virtualTools/virtualToolGrouper.ts
```diff
@@ -82,6 +97,14 @@ export class VirtualToolGrouper implements IToolCategorization {
...
 
+		const virtualToolEmbeddingRankingEnabled = this._configurationService.getExperimentBasedConfig(ConfigKey.Internal.VirtualToolEmbeddingRanking, this._expService);
+
+		if (virtualToolEmbeddingRankingEnabled) {
+			const predictedTools = await this._getPredictedTools(query, tools, token);
+
+			this._expandGroupsWithPredictedTools(root, predictedTools);
+		}
+
 		this._reExpandToolsToHitBudget(root);
```
connor4312: How about we refactor this so `_reExpandToolsToHitBudget` is generic and takes a ranker or comparator argument. By default it would be something like `this._reExpandToolsToHitBudget(root, g => g.tools.length)` but for embeddings it would be like `this._reExpandToolsToHitBudget(root, g => this._getGroupPredictedRelevancy(g, predictedTools))`

connor4312:
I get an error when we try to compare here

https://github.com/vritant24/vscode-copilot-chat/blob/cbf04b59ee7dfb9e94b5d32d4b534ad780c8c2e1/src/platform/embeddings/common/embeddingsComputer.ts#L162

`embedding` is simply an array of the embedding values, not the 

```
export interface Embedding {
	readonly type: EmbeddingType;
	readonly value: EmbeddingVector;
}
```

as expected

src/extension/tools/common/virtualTools/virtualToolGrouper.ts
vritant24: Thanks, that's a pretty great way structure it. I've made that change but allowing for more tools with the embedding based prediction as opposed to what's the limit today. If that feeld incorrect do let me know and I can keep the limits the same

vritant24:
> I get an error when we try to compare here
> 
> https://github.com/vritant24/vscode-copilot-chat/blob/cbf04b59ee7dfb9e94b5d32d4b534ad780c8c2e1/src/platform/embeddings/common/embeddingsComputer.ts#L162
> 
> `embedding` is simply an array of the embedding values, not the
> 
> ```
> export interface Embedding {
> 	readonly type: EmbeddingType;
> 	readonly value: EmbeddingVector;
> }
> ```
> 
> as expected

thanks, somehow I didn't run into this when testing but was able to repro it. It's now fixed

PR #682, opened 2025-08-20
don't include tasks for a second time in prompt

meganrogge:
fixes [#262243](https://github.com/microsoft/vscode/issues/262243)

Filter out task terminals - `processId` is always undefined @tyriar, which was unexpected 🤔

### DIFF
--- a/src/extension/prompts/node/base/terminalState.tsx
+++ b/src/extension/prompts/node/base/terminalState.tsx
@@ -27,14 +27,7 @@ export class TerminalStatePromptElement extends PromptElement<TerminalStateProps
 		const resultTasks: ITaskPromptInfo[] = [];
 		const allTasks = this.tasksService.getTasks()?.[0]?.[1] ?? [];
 		const tasks = Array.isArray(allTasks) ? allTasks : [];
-		const taskTerminalPids = new Set<number>();
-		await Promise.all(tasks.map(async (task) => {
-			const terminal = await this.tasksService.getTerminalForTask(task);
-			const terminalPid = terminal ? await terminal.processId : undefined;
-			if (terminalPid) {
-				taskTerminalPids.add(terminalPid);
-			}
-		}));
+		const activeTaskNames = tasks.filter(t => this.tasksService.isTaskActive(t)).map(t => t.label);
 
 		if (this.terminalService && Array.isArray(this.terminalService.terminals)) {
 			const terminals = await Promise.all(this.terminalService.terminals.map(async (term) => {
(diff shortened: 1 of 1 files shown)


Tyriar (approved):

src/extension/prompts/node/base/terminalState.tsx
```diff
@@ -97,19 +62,16 @@ export class TerminalAndTaskStatePromptElement extends PromptElement<TerminalAnd
 											Exit Code: {term.lastCommand.exitCode ?? '(unknown)'}<br />
 										</>
 									) : ''}
-									Output: {'{'}Use {ToolName.CoreGetTerminalOutput} for terminal with ID: {term.pid}.{'}'}<br />
+									Output: {'{'}Use {ToolName.CoreGetTerminalOutput} for terminal.{'}'}<br />
```
roblourens, line 65: Is this hint still helpful if it doesn't have the pid?

src/extension/prompts/node/base/terminalState.tsx
```diff
@@ -97,19 +62,16 @@ export class TerminalAndTaskStatePromptElement extends PromptElement<TerminalAnd
...
 					)}
 				</>
 			);
-			const renderedTasks = renderTasks();
-			const renderedTerminals = renderTerminals();
 			return (
 				<>
-					{resultTasks.length > 0 ? renderedTasks : 'Tasks: No tasks found.\n'}
-					{resultTerminals.length > 0 ? renderedTerminals : 'Terminals: No terminals found.\n'}
+					{resultTerminals.length > 0 ? renderTerminals() : 'Terminals: No terminals found.\n'}
```
roblourens, line 74: Is the "no terminals found" hint helpful? Maybe we can just omit the hint in that case?

src/extension/prompts/node/base/terminalState.tsx
```diff
@@ -56,32 +41,12 @@ export class TerminalAndTaskStatePromptElement extends PromptElement<TerminalAnd
...
 				} as ITerminalPromptInfo;
 			}));
-			const resultTerminals = terminals.filter(t => !!t);
+			const resultTerminals = terminals.filter(t => !!t && !activeTaskNames.includes(t.name));
 
 			if (resultTerminals.length === 0 && resultTasks.length === 0) {
 				return 'No tasks or terminals found.';
 			}
 
-			const renderTasks = () =>
```
roblourens, line 65: So this is removing all tasks but also they won't be rendered in the list of terminals? So no running task info?

src/extension/prompts/node/base/terminalState.tsx
```diff
@@ -56,32 +41,12 @@ export class TerminalAndTaskStatePromptElement extends PromptElement<TerminalAnd
 					} : undefined
 				} as ITerminalPromptInfo;
 			}));
-			const resultTerminals = terminals.filter(t => !!t);
+			const resultTerminals = terminals.filter(t => !!t && !activeTaskNames.includes(t.name));
 
 			if (resultTerminals.length === 0 && resultTasks.length === 0) {
 				return 'No tasks or terminals found.';
```
roblourens, line 47: Still mentioning tasks here but not below?

src/extension/prompts/node/base/terminalState.tsx
meganrogge: running task info is here already https://github.com/microsoft/vscode-copilot-chat/blob/e563a59af0fa1444354830faa9d0a461e00decd0/src/extension/prompts/node/agent/agentPrompt.tsx#L636-L650

PR #685, opened 2025-08-20
Fix bad ordering of thinking and tool prep chat response parts

roblourens:
Fixes an issue causing the chat to be rendered like this due to a response part being sent out of order before the last markdown part.

image

I don't really understand the `isBlockPart`, why don't we flush before every non-md part? Do you know @mjbvz?


roblourens:
I'm wondering what you remember from https://github.com/microsoft/vscode-copilot/issues/10026 and https://github.com/microsoft/vscode-copilot/pull/10347 too...

DonJayamanne (approved):

PR #691, opened 2025-08-20
rm task setting

meganrogge:
fixes https://github.com/microsoft/vscode/issues/262493

### DIFF
--- a/src/extension/intents/node/agentIntent.ts
+++ b/src/extension/intents/node/agentIntent.ts
@@ -80,7 +80,7 @@ const getTools = (instaService: IInstantiationService, request: vscode.ChatReque
 		}
 
 		allowTools[ToolName.RunTests] = await testService.hasAnyTests();
-		allowTools[ToolName.CoreRunTask] = !!(configurationService.getConfig(ConfigKey.AgentCanRunTasks) && tasksService.getTasks().length);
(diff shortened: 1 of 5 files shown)


meganrogge:
Thanks @Tyriar . Now that I see its usages, I'm wondering if we should just update the setting description vs remove this? Will see what @roblourens thinks, but going with that for now instead of this

Tyriar:
It's just a toggle for task tools isn't it? We do that here now:
image
image

roblourens:
It's fine to remove, afaik this is just from when it was experimental and we didn't have all the controls we have now

src/extension/intents/node/agentIntent.ts
```diff
@@ -80,7 +80,7 @@ const getTools = (instaService: IInstantiationService, request: vscode.ChatReque
 		}
 
 		allowTools[ToolName.RunTests] = await testService.hasAnyTests();
-		allowTools[ToolName.CoreRunTask] = !!(configurationService.getConfig(ConfigKey.AgentCanRunTasks) && tasksService.getTasks().length);
+		allowTools[ToolName.CoreRunTask] = tasksService.getTasks().length > 0;
```
roblourens, line 83: I think ideally this would be driven by a 'when' clause on the tool, not sure why we didn't do it that way originally

PR #701, opened 2025-08-21
Improve todo context prompt message

digitarald:
- Empty message points to tool
- Change tag name as manage todos tool also refers to `todoList` in its parameters


bhavyaus (approved):

bhavyaus:
We already have tool use guidance for todos. Just that we gate it behind a model check (gpt5). Should we not just relay on that? instead of having tool use guidance throughout the instructions

BAROTH1BOAT:
> * Empty message points to tool
> * Change tag name as manage todos tool also refers to `todoList` in its parameters

PR #725, opened 2025-08-22
Treat LM api request from core differently

lramos15:


roblourens:
Maybe we need a way to make sure these don't use a premium request too?

roblourens:
Although I think we mostly are just using 4o-mini so maybe it doesn't matter?

lramos15:
> Maybe we need a way to make sure these don't use a premium request too?

That's what this is doing

roblourens:
Sorry was focused on the headers... I see that you change that flag too thx

PR #727, opened 2025-08-22
Add prompt-tsx instructions

digitarald:
Based on codebase analysis, improving agentic flows on tsx files

### DIFF
--- a/.github/instructions/prompt-tsx.instructions.md
+++ b/.github/instructions/prompt-tsx.instructions.md
@@ -1,0 +1,166 @@
+---
+applyTo: '**/*.tsx'
+description: Prompt-TSX coding guidelines
+---
+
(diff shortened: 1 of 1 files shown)


.github/instructions/prompt-tsx.instructions.md
```diff
@@ -0,0 +1,166 @@
...
+### Async Components
+- Use `prepare` method for async operations before rendering
+- Pass prepared state to `render` method
+
+```tsx
+class FileContextPrompt extends PromptElement<Props, PreparedState> {
+	override async prepare(sizing: PromptSizing): Promise<PreparedState> {
+		// Async preparation logic
+		return state;
+	}
```
connor4312: I suggest removing this, `render` is async now so there is not much reason at all to use prepare

connor4312 (commented):
Looks decent, one comment

PR #733, opened 2025-08-23
Add encrypted thinking delta and clean up services

karthiknadig:
Introduce encrypted thinking delta functionality and remove the thinking service. Clean up related code and add tests to validate reasoning properties.

### DIFF
--- a/src/platform/endpoint/vscode-node/extChatEndpoint.ts
+++ b/src/platform/endpoint/vscode-node/extChatEndpoint.ts
@@ -22,6 +22,7 @@ import { TelemetryData } from '../../telemetry/common/telemetryData';
 import { ITokenizerProvider } from '../../tokenizer/node/tokenizer';
 import { CustomDataPartMimeTypes } from '../common/endpointTypes';
 import { decodeStatefulMarker, encodeStatefulMarker, rawPartAsStatefulMarker } from '../common/statefulMarkerContainer';
+import { rawPartAsThinkingData } from '../common/thinkingDataContainer';
 
 export class ExtensionContributedChatEndpoint implements IChatEndpoint {
 	private readonly _maxTokens: number;
@@ -266,6 +267,10 @@ export function convertToApiChatMessage(messages: Raw.ChatMessage[]): Array<vsco
 				const statefulMarker = rawPartAsStatefulMarker(contentPart);
 				if (statefulMarker) {
 					apiContent.push(new vscode.LanguageModelDataPart(encodeStatefulMarker(statefulMarker.modelId, statefulMarker.marker), CustomDataPartMimeTypes.StatefulMarker));
+				}
+				const thinkingData = rawPartAsThinkingData(contentPart);
+				if (thinkingData) {
+					apiContent.push(new vscode.LanguageModelThinkingPart(thinkingData.text, thinkingData.id, thinkingData.metadata));
 				}
 			}
 		}


src/platform/networking/common/networking.ts
```diff
@@ -226,13 +230,48 @@ export interface IChatEndpoint extends IEndpoint {
...
 	// python has `\ndef` and `\nclass` which must be stop words for ghost text
 	// const stops = getLanguageConfig<string[]>(accessor, ConfigKey.Stops);
 
+	const messages = [];
+	for (const message of options.messages) {
+		if (message.role === Raw.ChatRole.Assistant) {
+			let newMessage = rawMessageToCAPI(message);
+			for (const content of message.content) {
+				// opaque type
+				if (content.type === 2) {
```
roblourens: You can use real enums here

src/platform/networking/common/networking.ts
```diff
@@ -226,13 +230,48 @@ export interface IChatEndpoint extends IEndpoint {
...
+					const data = content.value as {
```
roblourens: Is that cast safe? It could be something else?

src/platform/networking/common/networking.ts
```diff
@@ -226,13 +230,48 @@ export interface IChatEndpoint extends IEndpoint {
...
+						type: string;
+						thinking: ThinkingData;
+					};
+					if (data.type === 'thinking' && data.thinking.id) {
+						if (options.reasoningPropertyType === 'AzureOpenAI') {
```
roblourens: If the goal is to say which endpoint it is, then this could be done in `createRequestBody` for the endpoint without passing a new property around

src/platform/thinking/common/thinking.ts
```diff
@@ -21,24 +31,29 @@ export type ThinkingDelta = {
...
-} |
-{
+} | {
 	text: string;
 	id?: string;
 	metadata?: string;
-	isEncrypted?: boolean;
 };
 
+export type EncryptedThinkingDelta = {
```
roblourens, line 44: In responses api, we can have encrypted content and summary text at the same time, are we losing that now?

karthiknadig: Ah! I missed that point. I will tweak this.

src/extension/byok/node/openAIEndpoint.ts
```diff
@@ -81,18 +79,26 @@ export class OpenAIEndpoint extends ChatEndpoint {
...
-	override createRequestBody(options: ICreateEndpointBodyOptions): IEndpointBody {
-		options.ignoreStatefulMarker = false;
-		const body = super.createRequestBody(options);
-		if (this.useResponsesApi) {
-			body.store = true;
-			body.n = undefined;
-			body.stream_options = undefined;
-			if (!this._modelInfo.capabilities.supports.thinking) {
-				body.reasoning = undefined;
+	protected override preprocessOptions(options: ICreateEndpointBodyOptions): ICreateEndpointBodyOptions {
```
roblourens: This method is odd because the options came from this class, they came from `makeChatRequest`

src/extension/byok/node/openAIEndpoint.ts
```diff
@@ -81,18 +79,26 @@ export class OpenAIEndpoint extends ChatEndpoint {
...
+		return { ...options, ignoreStatefulMarker: false };
+	}
+
+	protected override getCapiCallback(): RawMessageConversionCallback | undefined {
```
roblourens: Both the callback and customizeCapiBody seems like you could handle it by overriding the main methods since they are working on data that is passed out of those methods? But all the endpoint stuff would be cleaner IMO if the completions/responses handling was separate from the endpoint. But I think that's a big refactoring

karthiknadig: Made it a bit better, we can do the detangling of endpoint and completions/responses later.

src/extension/byok/node/openAIEndpoint.ts
```diff
@@ -82,18 +80,27 @@ export class OpenAIEndpoint extends ChatEndpoint {
 	}
 
 	override createRequestBody(options: ICreateEndpointBodyOptions): IEndpointBody {
-		options.ignoreStatefulMarker = false;
```
roblourens, line 85: Moving this to `makeChatRequest2` doesn't work because then it gets overridden by the `super` call, I can fix it

PR #754, opened 2025-08-25
revert prompt change

amunger:
unlist the replay capability as it is specifically for debugging specific chat replay files


test/simulation/cache/layers/f245cbe1-2962-48f7-9ce6-41a722693933.sqlite
Tyriar, line 1: Don't think you're meant to delete these as they're compacted later?

amunger: I did, they were created just for the PR that updated the prompt when it shouldn't have so they have cache entries that they shouldn't. But I also don't know what you mean by being compacted later, so I guess I'll just leave them in there to be safe.

Tyriar: I also didn't see it was just a revert, it's probably fine.

PR #756, opened 2025-08-25
Have apply patch apply to BYOK

lramos15:


connor4312 (approved):

mcowger:
Any reason this isn't enabled by default (or at least a choice to be configured by the user) for arbitrary models?

PR #760, opened 2025-08-25
Increase number of tokens full workspace can use

mjbvz:
Adjusting some of the default numbers. Tool calling was ending up with a pretty small budget for `#codebase`

### DIFF
--- a/src/extension/prompts/node/panel/codebaseAgentPrompt.tsx
+++ b/src/extension/prompts/node/panel/codebaseAgentPrompt.tsx
@@ -29,7 +29,7 @@ export class CodebaseAgentPrompt extends PromptElement<GenericBasePromptElementP
 
 	async render(state: void, sizing: PromptSizing) {
 		const { query, chatVariables, history, toolCallRounds, toolCallResults } = this.props.promptContext;
-		const isCodesearchFast = await this.workspaceChunkSearch.hasFastSearch({ endpoint: this.promptEndpoint, tokenBudget: sizing.tokenBudget, maxResultCountHint: MAX_CHUNKS_RESULTS });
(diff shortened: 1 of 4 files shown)


src/extension/prompts/node/panel/codebaseAgentPrompt.tsx
```diff
@@ -29,7 +29,7 @@ export class CodebaseAgentPrompt extends PromptElement<GenericBasePromptElementP
 
 	async render(state: void, sizing: PromptSizing) {
 		const { query, chatVariables, history, toolCallRounds, toolCallResults } = this.props.promptContext;
-		const isCodesearchFast = await this.workspaceChunkSearch.hasFastSearch({ endpoint: this.promptEndpoint, tokenBudget: sizing.tokenBudget, maxResultCountHint: MAX_CHUNKS_RESULTS });
+		const isCodesearchFast = await this.workspaceChunkSearch.hasFastSearch({ endpoint: this.promptEndpoint, tokenBudget: sizing.tokenBudget, fullWorkspaceTokenBudget: sizing.tokenBudget, maxResultCountHint: MAX_CHUNKS_RESULTS });
```
roblourens, line 32: How often does full workspace happen and what does this budget work out to? It might make more sense to not do that for agent mode so we can return small targeted results rather than pump it full of all the context all at once.

mjbvz: Full workspace is used for 10-20% of searches. Before this change we had a limit of 32k tokens for normal `@workspace`, and 20k for `#codebase` tool calls

So this change does mean that more context may be included from the tool call. The upside is that once the context is included from full workspace, we should not need to call `#codebase` or any other workspace related tools in the turn since the entire workspace is now in the context (ignoring summarization). I'll try to get the numbers but this should translate to fewer tool calling rounds and fewer input tokens over the course of the entire request

roblourens: Why is the conclusion fewer input tokens when we might be filling it with unneeded context? I don't really see why we would do full workspace in agent mode at all, it seems more helpful for the non-agent world

PR #764, opened 2025-08-26
Straighten out body.reasoning handling

roblourens:


src/extension/byok/node/test/openAIEndpoint.spec.ts
```diff
@@ -47,28 +40,11 @@ const createTestOptions = (messages: Raw.ChatMessage[]): ICreateEndpointBodyOpti
 	location: undefined as any
 });
 
-// Mock implementations
-const createMockServices = (useResponsesApi = false) => ({
```
roblourens, line 51: fyi @karthiknadig this broke when I made a few changes, we have these helpers like `createExtensionUnitTestingServices` and then I can use real test services to do things lik echange config

PR #775, opened 2025-08-26
Bubble up embeddings endpoint errors

mjbvz:
Right now we return `undefined` when computing embeddings fails. I think it makes more sense to throw exceptions with the  details of why this failed so we can track these

### DIFF
--- a/src/extension/prompts/node/panel/newWorkspace/newWorkspace.tsx
+++ b/src/extension/prompts/node/panel/newWorkspace/newWorkspace.tsx
@@ -106,29 +106,27 @@ export class NewWorkspacePrompt extends PromptElement<NewWorkspacePromptProps, N
 			else if (instruction.intent === 'Project') {
 				if (this.props.useTemplates) {
 					const result = await this.embeddingsComputer.computeEmbeddings(EmbeddingType.text3small_512, [instruction.question], {}, undefined);
-					if (result && result.values.length > 0) {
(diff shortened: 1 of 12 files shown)


src/extension/prompts/node/panel/newWorkspace/newWorkspace.tsx
```diff
@@ -106,29 +106,27 @@ export class NewWorkspacePrompt extends PromptElement<NewWorkspacePromptProps, N
...
-										url = `httpx://github.com/${org}/${repo}`;
-									} else {
-										url = path.dirname(items[0].html_url);
-									}
+						if (org && repo && repoPath) {
+							const items = await reportProgressOnSlowPromise(progress, new ChatResponseProgressPart(l10n.t('Fetching project contents...')), this.repositoryService.getRepositoryItems(org, repo, repoPath), 500);
+							if (items.length > 0) {
+								let url: string;
+								if (repoPath === '.') {
+									url = `httpx://github.com/${org}/${repo}`;
```
Tyriar, line 122: this x is meant to be there?

Tyriar (approved):

src/extension/prompts/node/panel/newWorkspace/newWorkspace.tsx
mjbvz: cc @bhavyaus Just checking if this intentional

PR #778, opened 2025-08-26
avoid runtime SHA-256 for 'grok-code' model

42HD:
Replace the runtime SHA-256 comparison in `isHiddenModelB` with a direct plain-family equality check for the known model family "grok-code". This removes unnecessary crypto overhead and ensures the function always returns a boolean Promise.


lramos15:
We would likely want to remove the isHiddenModel function all together since this model is no longer hidden. Then you would just inline grok code with the rest of the checks that this function was calling into

PR #780, opened 2025-08-27
Streamline toolsets, cleanup tool descriptions

roblourens:


src/extension/promptFileContext/vscode-node/promptFileContextService.ts
```diff
@@ -199,7 +199,7 @@ export class PromptFileContextContribution extends Disposable {
 	}
 
 	private getToolNames(): string[] {
-		return ['changes', 'codebase', 'editFiles', 'extensions', 'fetch', 'findTestFiles', 'githubRepo', 'new', 'openSimpleBrowser', 'problems', 'runCommands', 'runNotebooks', 'runTasks', 'runTests', 'search', 'searchResults', 'terminalLastCommand', 'terminalSelection', 'testFailure', 'usages', 'vscodeAPI'];
+		return ['changes', 'codebase', 'editFiles', 'extensions', 'fetch', 'findTestFiles', 'githubRepo', 'new', 'notebooks', 'openSimpleBrowser', 'problems', 'runCommands', 'runTasks', 'runTests', 'search', 'searchResults', 'terminalLastCommand', 'terminalSelection', 'testFailure', 'usages', 'vscodeAPI'];
```
DonJayamanne, line 202: @roblourens wouldn't this be a breaking change for users.
I.e. now that `editFiles` doesn't contain notebook related tools anyone who has created chat modes or the like woudn't necessarily get notebook tools if they had previously only included `editFiles` tools

roblourens: @aeschli what do you think? It's hard to say that these toolsets need to be frozen forever once they are shipped but it's true that they may be referenced in prompts or modes already. So we may have to leave them alone.

aeschli: It also breaks the enabled tools list in the global storage. At least there' all unknown (new) tool sets are considered as enabled. But for mode/prompt file, all tools sets that are not in the array are considered turned off (->no `edit` and `notebooks` tools)
-We probably should keep the declaration of the old tool sets (`editFiles`) and mark then deprecated. Deprecated tool sets will not be suggested in the tools picker, but can be used to read (old) storage and prompt files.
- The new 'notebook' set now contains write and read tools. IMO we should make it easy to deselect all editing tools with one click. I would put `editNotebook` back into 'editFiles'

roblourens: I was with you but we had a discussion in standup that people want to disable all notebook-related tools when not working with notebooks, and that is more important than grouping by being related to "editing".

PR #805, opened 2025-08-28
Separate feature flag for displaying Next Editor NES

DonJayamanne:
For https://github.com/microsoft/vscode/issues/255704

### DIFF
--- a/src/extension/inlineEdits/vscode-node/inlineCompletionProvider.ts
+++ b/src/extension/inlineEdits/vscode-node/inlineCompletionProvider.ts
@@ -101,6 +101,7 @@ export class InlineCompletionProviderImpl implements InlineCompletionItemProvide
 	private readonly _tracer: ITracer;
 
 	public readonly onDidChange: vscodeEvent<void> | undefined = Event.fromObservableLight(this.model.onChange);
+	private readonly _displayNextEditorNES: boolean;
 
 	constructor(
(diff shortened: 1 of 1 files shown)


src/extension/inlineEdits/vscode-node/inlineCompletionProvider.ts
```diff
@@ -247,7 +249,7 @@ export class InlineCompletionProviderImpl implements InlineCompletionItemProvide
 				completionItem = serveAsCompletionsProvider && !isInlineCompletion ?
 					undefined :
 					this.createCompletionItem(doc, document, position, range, result);
-			} else {
+			} else if (this._displayNextEditorNES) {
```
DonJayamanne, line 252: @hediet We want to separate the two experiments
1. Experiment 1 for Displaying next edits in other cells as 
2. Experiment 2 for how and when we display UI for next editor suggestions `Go To Next Edit`.
The plan is to try and ship 1 before 2. and 2 might change.

1 will be identical to what we have today, no ui changes, getting NES just for the cell where cursor is.

DonJayamanne: Will end up creating a separate feature/experiment setting for 2, will do that later.
Already created a separate experiment (setting) for 1.

Tyriar (approved):

PR #821, opened 2025-08-28
handle array type thinking

justschen:
needs https://github.com/microsoft/vscode/pull/263922 to not break rendering

### DIFF
--- a/src/platform/endpoint/node/responsesApi.ts
+++ b/src/platform/endpoint/node/responsesApi.ts
@@ -248,7 +248,7 @@ class OpenAIResponsesProcessor {
 							// CAPI models don't stream the reasoning summary for some reason, byok do, so don't duplicate it
 							text: this.hasReceivedReasoningSummary ?
 								undefined :
-								chunk.item.summary.map(s => s.text).join('\n\n'),
+								chunk.item.summary.map(s => s.text),
 							encrypted: chunk.item.encrypted_content,
 						} : undefined
 					});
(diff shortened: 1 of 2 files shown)


src/platform/endpoint/node/responsesApi.ts
```diff
@@ -248,7 +248,7 @@ class OpenAIResponsesProcessor {
 							// CAPI models don't stream the reasoning summary for some reason, byok do, so don't duplicate it
 							text: this.hasReceivedReasoningSummary ?
 								undefined :
-								chunk.item.summary.map(s => s.text).join('\n\n'),
+								chunk.item.summary.map(s => s.text),
```
karthiknadig, line 251: This should also be fixed in the `summary_text.delta` handling below.

karthiknadig: Essentially, we will need to use `summary_index` to create the right delta.
if `summary_index` is 0: 'text'
if `summart_index` is 1: ['','text'] 
if `summary_index` is 2: ['', '','text'] and so on.

Really once we have some models we can start testing this. OpenAI BYOK actually gets you the correct data, if you want to test and see the behavior.

justschen: `chunk.delta` is already a string here though, and we get it back as a `string` not a `string[]`

karthiknadig (approved):

src/platform/endpoint/node/responsesApi.ts
karthiknadig: Discussed offline. This will be updated as a follow up PR.

essentially, the updated point here is handling a CAPI problem where we are not getting streaming reasoning summary text. But, when using BYOK/BYOE you actually get `summary_text.delta` with `summary_index`. That could be overlapping, so that has to be handled to avoid mixing summaries between different indexes.

PR #872, opened 2025-09-01
Add INativeEnvService to avoid 'os' module import, cleanup tests

roblourens:


src/extension/extension/vscode-node/services.ts
```diff
@@ -121,6 +122,7 @@ export function registerServices(builder: IInstantiationServiceBuilder, extensio
 	builder.define(ITokenizerProvider, new SyncDescriptor(TokenizerProvider, [true]));
 	builder.define(IToolsService, new SyncDescriptor(ToolsService));
 	builder.define(IRequestLogger, new SyncDescriptor(RequestLogger));
+	builder.define(INativeEnvService, new SyncDescriptor(NativeEnvServiceImpl));
```
roblourens, line 125: cc @bpasero this was inspired by your services in core, so lmk if the way I did it is ok

roblourens: Maybe it would have been better to not register `IEnvService` and `INativeEnvService` as two separate instances but it wasn't clear how to get that right. And I'm not sure how `refineServiceDecorator` works.

bpasero: Use `refineServiceDecorator` when the one extends from the other. I would recommend to use it in this case.

Example: https://github.com/microsoft/vscode/blob/966ec1415440da5f1ef53acbd97ba2113add8853/src/vs/workbench/services/environment/browser/environmentService.ts#L23

PR #878, opened 2025-09-02
Disallow importing straight from /extHostTypes/ and avoid instanceof check in ExtHostDocumentData

alexdima:


roblourens (dismissed):
But if we have extHostTypes can we replace vscodeTypesShim with pulling types from extHostTypes? Because my process of adding new things to vscodeTypes is just copy/pasting the implementation from extHostTypes... I don't know whether some of them are different

roblourens:
Oh I guess a lot of types in Shim are already doing that... so why do we have copies in `chatTypes`? I forget

alexdima:
For vscode shims, I initially started it by copy-pasting implementations from vscode. In the meantime, we had the `copySources.ts` thing to avoid "forking" the code and to allow catching up with upstream. A few days ago, I learnt that the shims code is not used only for testing, but will also be used in the npm node module that will ship the NES code. So I've decided to try to reuse more of the upstream code via `copySources.ts`. But there is a configuration where we load both the vscode API and we make use of fake text documents (bench), so I'm trying to make that thing happy too.

PR #882, opened 2025-09-03
Pass proper header to auto mode

lramos15:
cc @roblourens. I think it makes sense to pass additional headers through `makeChatRequest2` and then see about removing `getExtraHeaders`. I'll attempt to tackle that in debt week. Let me know if you have concerns.

### DIFF
--- a/src/platform/endpoint/common/autoChatEndpoint.ts
+++ b/src/platform/endpoint/common/autoChatEndpoint.ts
@@ -4,11 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 
 import type { RequestMetadata } from '@vscode/copilot-api';
-import { ChatMessage } from '@vscode/prompt-tsx/dist/base/output/rawTypes';
+import { Raw } from '@vscode/prompt-tsx';
 import type { CancellationToken } from 'vscode';
 import { ITokenizer, TokenizerType } from '../../../util/common/tokenizer';
 import { AsyncIterableObject } from '../../../util/vs/base/common/async';
-import { Source } from '../../chat/common/chatMLFetcher';
+import { IChatMLFetcher, Source } from '../../chat/common/chatMLFetcher';
 import { ChatLocation, ChatResponse } from '../../chat/common/commonTypes';
 import { IEnvService } from '../../env/common/envService';
 import { ILogService } from '../../log/common/logService';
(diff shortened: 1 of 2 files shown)


src/platform/endpoint/common/autoChatEndpoint.ts
```diff
@@ -79,11 +80,29 @@ export class AutoChatEndpoint implements IChatEndpoint {
...
 	}
 
-	async makeChatRequest2(options: IMakeChatRequestOptions, token: CancellationToken): Promise<ChatResponse> {
-		return this._wrappedEndpoint.makeChatRequest2(options, token);
+	public async makeChatRequest2(options: IMakeChatRequestOptions, token: CancellationToken): Promise<ChatResponse> {
+		return this._makeChatRequest2({ ...options, ignoreStatefulMarker: options.ignoreStatefulMarker ?? true }, token);
 	}
 
-	async makeChatRequest(debugName: string, messages: ChatMessage[], finishedCb: FinishedCallback | undefined, token: CancellationToken, location: ChatLocation, source?: Source, requestOptions?: Omit<OptionalChatRequestParams, 'n'>, userInitiatedRequest?: boolean, telemetryProperties?: TelemetryProperties): Promise<ChatResponse> {
+	public async _makeChatRequest2(options: IMakeChatRequestOptions, token: CancellationToken) {
```
roblourens, line 87: Why are there two of them and why is it public?

lramos15: I copied this from the `ChatEndpoint` that you implemented :p. https://github.com/microsoft/vscode-copilot-chat/blob/06eeb448555c521e712ab0c3c16d046c029ad43c/src/platform/endpoint/node/chatEndpoint.ts#L335-L352

roblourens: There we call `_makeChatRequest2` multiple times (in the code that's commented out) so it makes sense IMO

lramos15: But it still doesn't make sense why `_makeChatRequest2` is public does it?

roblourens: No it doesn't, guess that slipped by me, so I just fixed it thanks :)

roblourens: https://github.com/microsoft/vscode-copilot-chat/pull/932

PR #890, opened 2025-09-03
edits: adjust confirmations for system files and outside the workspace

connor4312:
- Don't block edits to files outside the workspace, but allow them with
  confirmations.
- Always confirm on certain files (dotfiles/folders in the home
  directory and appdatas on Windows)

### DIFF
--- a/src/extension/tools/node/editFileToolUtils.tsx
+++ b/src/extension/tools/node/editFileToolUtils.tsx
@@ -3,20 +3,23 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { t } from '@vscode/l10n';
+import { homedir } from 'os';
 import type { LanguageModelChat, PreparedToolInvocation } from 'vscode';
 import { IConfigurationService } from '../../../platform/configuration/common/configurationService';
+import { ICustomInstructionsService } from '../../../platform/customInstructions/common/customInstructionsService';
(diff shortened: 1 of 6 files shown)


src/extension/tools/node/editFileToolUtils.tsx
```diff
@@ -557,35 +574,55 @@ function makeUriConfirmationChecker(configuration: IConfigurationService) {
...
 	return {
 		confirmationMessages: {
 			title: t('Allow edits to sensitive files?'),
-			message: new MarkdownString(t`The model wants to edit sensitive files (${fileParts}). Do you want to allow this?` + '\n\n' + asString()),
+			message: new MarkdownString(
+				(needsConfirmation.some(r => r.reason === ConfirmationCheckResult.Sensitive)
+					? t`The model wants to edit sensitive files (${fileParts}).`
+					: needsConfirmation.some(r => r.reason === ConfirmationCheckResult.OutsideWorkspace)
+						? t`The model wants to edit files outside of your workspace (${fileParts}).`
+						: t`The model wants to edit system files (${fileParts}).`)
```
Tyriar, line 623: system files more important that out of workspace? Should they go first?

connor4312: My thought was going configurable to non-configurable. Really only matters in a case of a multi-edit tool with multiple exlusion, but my thought here was that a user could realize they can open a workspace containing some of the files they want to change (if needed) whereas a system file doesn't really give them more information

PR #908, opened 2025-09-04
Add User NES Accept/Dismiss Counters

zelinms:


src/extension/xtab/node/xtabProvider.ts
```diff
@@ -87,6 +93,9 @@ export class XtabProvider implements IStatelessNextEditProvider {
 
 	private forceUseDefaultModel: boolean = false;
 
+	private nesAcceptCounter: number = 0;
```
benibenj, line 96: Instead of absolute counts, wouldn't it make more sense to store how many % of suggestions were accepted / rejected / ignored ?

zelinms: Yeah, that makes sense.
Currently we haven't trained the model yet. I draft this PR to verify that we do can pass such information to model prompt.
We can try different ways to define how the activity level should be calculated.

benibenj: I see. Yes that should be possible. Are you planning on passing number values to the model or a string value like: `active`, `balanced` or `passive`? We have been considering providing such a setting to the user which would lead to showing less/more suggestions and maybe slower/faster suggestions depending on the setting. Providing this information to the model would be great. 

Maybe if the user does not configure this setting, then we would try and set if for the user depending on how often they interact with NES. We could also provide a different value (`active`, `passive`, `balanced`) based on the file the user is in or reset the state when the user commits it's changes.

ulugbekna (commented):
Thank you for the PR, this looks promising (and very user-requested) ! :-) 

I agree with Ben that activity should be more of a "rate", eg `# of accepted / # of shown` (but not sure we store just # of opportunities, # of accepted, and then compute the rate) 

Activity also feels like something that can vary based on kind of work, eg sometimes I refactor code and need NES to be super fast and active, and sometimes I write new complex code and need it to be non-distracting. So potentially, activity level should be computed per-session/per git branch change. Maybe the activity level for session N should start at the activity level at the end of session N-1. 

Let me know when this PR intends to be merge-able, and I provide implementation-related comments.

zelinms:
> Thank you for the PR, this looks promising (and very user-requested) ! :-)
> 
> I agree with Ben that activity should be more of a "rate", eg `# of accepted / # of shown` (but not sure we store just # of opportunities, # of accepted, and then compute the rate)
> 
> Activity also feels like something that can vary based on kind of work, eg sometimes I refactor code and need NES to be super fast and active, and sometimes I write new complex code and need it to be non-distracting. So potentially, activity level should be computed per-session/per git branch change. Maybe the activity level for session N should start at the activity level at the end of session N-1.
> 
> Let me know when this PR intends to be merge-able, and I provide implementation-related comments.

Thanks @ulugbekna.
We will need to train a model that can change behaviors according to `ActivityLevel`, and this may take 1-2 weeks. After we get the first version of the model, I will seek your guidance to implement it on the client side.

PR #910, opened 2025-09-04
Consume changed API shape

lramos15:
Chat side of https://github.com/microsoft/vscode/pull/265213

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
 	"icon": "assets/copilot.png",
 	"pricing": "Trial",
 	"engines": {
-		"vscode": "^1.104.0-20250828",
+		"vscode": "^1.104.0-20250905",
 		"npm": ">=9.0.0",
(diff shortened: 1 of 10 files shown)


mjbvz (approved):

package.json
```diff
@@ -24,7 +24,7 @@
 	"icon": "assets/copilot.png",
 	"pricing": "Trial",
 	"engines": {
-		"vscode": "^1.104.0-20250828",
+		"vscode": "^1.104.0-20250905",
```
roblourens, line 27: FWIW I don't think changes like this need an engine version and dts version bump (and you should only ever need one of those). It looks like the only functional change is adding an optional statusIcon prop?

lramos15: Makes sense, I really didn't want to break anyone here. And we changed the core `vscode.d.ts` which doesn't have versioning

PR #951, opened 2025-09-08
Fix response handling in GHRepoService and add types to prevent futur…

TylerLeonhardt:
…e issues

ref https://github.com/microsoft/vscode/issues/265287


alexr00 (commented):
Thank you!

PR #952, opened 2025-09-08
Yemohyle/dedup messages telemetry

yemohyleyemohyle:
Refactor microsoft internal telemetry for agent mode to be sent in smaller events to avoid high drop rate. This is specifically targeted to substitute engine.messages events that attempt to send entire input to each model call, which includes all the messages from the conversation history up to the moment and contains lots of duplicates


src/platform/networking/node/chatStream.ts
```diff
@@ -58,23 +61,398 @@ export function sendEngineMessagesLengthTelemetry(telemetryService: ITelemetrySe
...
+
 export function sendEngineMessagesTelemetry(telemetryService: ITelemetryService, messages: CAPIChatMessage[], telemetryData: TelemetryData, isOutput: boolean, logService?: ILogService) {
 	const telemetryDataWithPrompt = telemetryData.extendedBy({
 		messagesJson: JSON.stringify(messages),
 	});
 	telemetryService.sendEnhancedGHTelemetryEvent('engine.messages', multiplexProperties(telemetryDataWithPrompt.properties), telemetryDataWithPrompt.measurements);
 	telemetryService.sendInternalMSFTTelemetryEvent('engine.messages', multiplexProperties(telemetryDataWithPrompt.properties), telemetryDataWithPrompt.measurements);
 
+	// Send all model telemetry events (model.request.added, model.message.added, model.modelCall.input/output, model.request.options.added)
+	// Comment out the line below to disable the new deduplicated model telemetry events
```
roblourens, line 454: Why the comment about disabling it, is that something somebody needs to do?

src/platform/networking/node/chatStream.ts
```diff
@@ -58,23 +61,398 @@ export function sendEngineMessagesLengthTelemetry(telemetryService: ITelemetrySe
...
+	}
+}
+
+/**
+ * Updates the conversationTracker with the given conversationId.
+ * If the conversationId is different from the one stored, resets to turn 1.
+ * If it's the same conversationId, increments the turn.
+ * Returns the current conversationTurn for the conversationId.
+ */
+function updateConversationTracker(conversationId: string): number {
```
roblourens, line 146: Isn't this wrong when you have more than one conversation going at a time?

src/platform/networking/node/chatStream.ts
```diff
@@ -58,23 +61,398 @@ export function sendEngineMessagesLengthTelemetry(telemetryService: ITelemetrySe
...
+	// It's called once per unique user request (identified by headerRequestId)
+	// It excludes message content and request options which are captured separately
+
+	// Extract headerRequestId to check for uniqueness
+	const headerRequestId = telemetryData.properties.headerRequestId;
+	if (!headerRequestId) {
+		return;
+	}
+
+	// Check if this is a conversation mode (has conversationId) or supplementary mode
```
roblourens, line 233: When do we not have a conversationId?

src/platform/networking/node/chatStream.ts
```diff
@@ -58,23 +61,398 @@ export function sendEngineMessagesLengthTelemetry(telemetryService: ITelemetrySe
...
+		const messageHash = hash({
+			role: message.role,
+			content: message.content,
+			headerRequestId: headerRequestId, // Include headerRequestId in hash for proper deduplication
+			...(('tool_calls' in message && message.tool_calls) && { tool_calls: message.tool_calls }),
+			...(('tool_call_id' in message && message.tool_call_id) && { tool_call_id: message.tool_call_id })
+		}).toString();
+
+		// Get existing UUID for this message content + headerRequestId combination, or generate a new one
+		let messageUuid = messageHashToUuid.get(messageHash);
```
roblourens, line 299: Can we send the hashes instead of having to map it to uuids? So that we can even track conversatations that continue after reloading the window, etc

roblourens: But I have no idea if that would be allowed.

src/platform/networking/node/chatStream.ts
```diff
@@ -58,23 +61,398 @@ export function sendEngineMessagesLengthTelemetry(telemetryService: ITelemetrySe
...
+
+			telemetryService.sendInternalMSFTTelemetryEvent('model.message.added', messageData.properties, messageData.measurements);
+		}
+	}
+
+	return messageData; // Return collected message data with UUIDs and headerRequestIds
+}
+
+function sendModelCallTelemetry(telemetryService: ITelemetryService, messageData: Array<{ uuid: string; headerRequestId: string }>, telemetryData: TelemetryData, messageDirection: 'input' | 'output', logService?: ILogService) {
+	// Get the unique model call ID
```
roblourens, line 346: I don't think I understand what this function is doing, what's a "model call"?

roblourens (commented):
It's really sad that we have to do this at this layer but thanks for doing it! I worry that telemetry could still be too lossy to reconstruct the trajectories but I hope it works. I think I mostly follow it but just a few questions.

src/platform/networking/node/chatStream.ts
yemohyleyemohyle: I do not see conversationId available when semantic_search has its own model call, or edit healing ... I called it supplementary mode.

src/platform/networking/node/chatStream.ts
yemohyleyemohyle: For every model call, there’s an input array of messages and an output array containing a single message.
This function doesn’t send the full messages directly. Instead, it sends events with an array of messageUuids (unique message identifiers).
For example, if the agent takes n turns, there are n model calls. A system message will appear at the start of each input array across all n turns. Rather than sending the entire system message n times, we send it once in a model.message.added event. Then, in each of the model.modelCall events, we only include the corresponding messageUuid and its position in the array.

src/platform/networking/node/chatStream.ts
yemohyleyemohyle: I am not sure I follow here. If the question is why don't we use hash itself instead of creating messageUuids?  I was not sure how large those hashes could be, so used a standard uuid.

src/platform/networking/node/chatStream.ts
yemohyleyemohyle: Hopefully not. At the moment the existing engine.messages event and model.... events I am adding are substitutes of each other. So after conformation that the new schema works better we would disable engine.messages events. In case the new schema (shorter events but more of them) still does not help with drop rate, it may be disabled.

src/platform/networking/node/chatStream.ts
yemohyleyemohyle: Are parallel conversations allowed by UI? It would be then. I changed the conversationTracker to allow for parallel conversations.

chrmarti (approved):

PR #978, opened 2025-09-09
Update completions core version for telemetry

benibenj:
```Copilot Generated Description:``` Update the completions core version to include telemetry sending.

fixes https://github.com/microsoft/vscode/issues/265829


dbaeumer:
Done a code review and the error case doesn't log telemetry events. Reason being:

- https://github.com/microsoft/vscode-copilot-chat/blob/release/0.31/src/extension/completions-core/extension/src/inlineCompletion.ts#L132 catches the exception but doesn't re-throw
- hence this code is never hit: https://github.com/microsoft/vscode-copilot-chat/blob/release/0.31/src/extension/completions-core/extension/src/inlineCompletion.ts#L86 and errors will not be logged into telemetry.

PR #999, opened 2025-09-10
enable auto only when editor preview features are ON

isidorn:
refs microsoft/vscode#265902


isidorn:
Thank you for the review. Merging in.
I will create a follow-up PR for the release branch.

PR #1029, opened 2025-09-12
Reduce INTERNAL_RESTRICTED

roblourens:
and delete edits linkification setting


justschen:
@roblourens looks like some checks failed

PR #1037, opened 2025-09-12
Add CAPI text-3-small endpoint support for embeddings

bhavyaus:
Bring back support for CAPI text-3-small embeddings endpoints.

This was removed as a part of https://github.com/microsoft/vscode-copilot-chat/pull/595

### DIFF
--- a/src/platform/embeddings/common/remoteEmbeddingsComputer.ts
+++ b/src/platform/embeddings/common/remoteEmbeddingsComputer.ts
@@ -272,7 +272,7 @@ export class RemoteEmbeddingsComputer implements IEmbeddingsComputer {
 				this._capiClientService,
 				endpoint,
 				token.token,
-				await createRequestHMAC(env.HMAC_SECRET), // TODO@bpasero we need web support for these environmental things
+				await createRequestHMAC(env.HMAC_SECRET),
 				'copilot-panel',
 				requestId,
 				body,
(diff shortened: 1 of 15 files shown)


src/platform/embeddings/common/remoteEmbeddingsComputer.ts
```diff
@@ -127,4 +144,161 @@ export class RemoteEmbeddingsComputer implements IEmbeddingsComputer {
...
+	 */
+	public async rawEmbeddingsFetch(
+		type: EmbeddingTypeInfo,
+		endpoint: IEmbeddingsEndpoint,
+		requestId: string,
+		inputs: readonly string[],
+		cancellationToken: CancellationToken | undefined
+	): Promise<CAPIEmbeddingResults | CAPIEmbeddingError> {
+		try {
+			const token = await this._authService.getCopilotToken();
```
bhavyaus, line 265: @bpasero how does this resolve in the scenario we're trying to test?

bpasero: @bhavyaus I think we have not yet surfaced this in the service, but probably should. There is a `authFromDevDeviceId` that can return a Copilot token from `vscode.env.devDeviceId`:

https://github.com/microsoft/vscode-copilot-chat/blob/7512c12af7e82ef095e44200af881e9eb000d9a0/src/platform/authentication/node/copilotTokenManager.ts#L109-L117

We use this in the Copilot token manager on startup only if `chat.allowAnonymousAccess` is set:

https://github.com/microsoft/vscode-copilot-chat/blob/7512c12af7e82ef095e44200af881e9eb000d9a0/src/platform/authentication/vscode-node/copilotTokenManager.ts#L61

//cc @lramos15

bpasero (changes requested):
Build seems red.

PR #1056, opened 2025-09-17
Fix switching between oswe and gpt-5 (#1035)

roblourens:
Fix #1035

### DIFF
--- a/package.json
+++ b/package.json
@@ -3009,6 +3009,43 @@
 							"experimental"
 						],
 						"description": "%github.copilot.config.grokCodeAlternatePrompt%"
+					},
+					"github.copilot.chat.useResponsesApi": {
+						"type": "boolean",
+						"default": false,
+						"markdownDescription": "%github.copilot.config.useResponsesApi%",
+						"tags": [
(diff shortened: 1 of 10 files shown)


src/extension/byok/vscode-node/baseOpenAICompatibleProvider.ts
```diff
@@ -94,9 +94,7 @@ export abstract class BaseOpenAICompatibleLMProvider implements BYOKModelProvide
 
 	private async getEndpointImpl(model: LanguageModelChatInformation): Promise<OpenAIEndpoint> {
 		const modelInfo: IChatModelInformation = await this.getModelInfo(model.id, this._apiKey);
-		const url = modelInfo.supported_endpoints?.includes(ModelSupportedEndpoint.Responses) ?
```
roblourens, line 97: @lramos15 the CAPI gpt-5 model doesn't have the 'thinking' capability set? I think that the easiest thing to do for a quick candidate right now is to disable the responses API in BYOK. Or maybe I'll think of another idea tomorrow

PR #1074, opened 2025-09-21
improve custom OpenAI Compatible model URL resolution

DGideas:
In this commit, I
* Update resolveCustomOAIUrl function to handle URLs with existing version patterns
* Remove special handling for /v1 to support any version number (v1, v2, v4, etc.)

Related Issue: https://github.com/microsoft/vscode/issues/267616#issuecomment-3315504055

Step to validate:
1. Add the following config to `settings.json`
```json
	"github.copilot.chat.customOAIModels": {
		"GLM-4.5": {
			"name": "GLM 4.5",
			"url": "https://api.z.ai/api/coding/paas/v4",
			"maxInputTokens": 131200,
			"maxOutputTokens": 96000,
			"requiresAPIKey": true,
			"thinking": true,
			"toolCalling": true,
			"vision": false,
		},
	}
```
2. After apply this patch, Copilot Chat shouldn't show any error message when works with this model endpoint


DGideas:
Before this fix:
image

After this fix:
image

PR #1075, opened 2025-09-21
errors - indicate rate limiting errors back to the client

bpasero:
@lramos15 please have a look, this is for being able to show a specific UI for no-auth users in the client when a response is rate limited.


roblourens (approved):

lramos15:
Why are we adding a new API type here? Is this to drive a sign in button? I'm wondering if we should just allow these error messages to have command buttons in them so we don't have to do this API flow

bpasero:
I think the idea is that core would be driving the UI and not the extension providing command buttons. The commands are actually implemented all in core. 

So in essence, we need a way to know a little bit more about the error kind, it does not have to be a boolean though...

lramos15:
Do we foresee the error types we want to handle from core continuing to grow? If so then yeah maybe switching to an enum makes sense

bpasero:
I can look into that.

PR #1079, opened 2025-09-22
nes: xtab: refactor: nicer references to tags

ulugbekna:


lszomoru:
@ulugbekna, PR check is 🔴

PR #1108, opened 2025-09-23
Don't send empty assistant messages

roblourens:


pierceboggan:
Tested, and fix appears to be working for me.

PR #1111, opened 2025-09-23
Grouping of BYOK custom models in model picker

johan-j:
This add new grouping and using structured data for Bring Your Own Key.

Issue:
- https://github.com/github/models-factory/issues/2796

Using the spike branch as starting point: https://github.com/microsoft/vscode-copilot-chat/commit/87451aabfc2839ac48bc111094bcae6f46288632#diff-bcf7cff25eec039e0563b151a1299f0ad1f483594076c6bb649e2132e57d07e8


johan-j:

src/extension/conversation/vscode-node/languageModelAccess.ts
```diff
@@ -153,19 +153,28 @@ export class LanguageModelAccess extends Disposable implements IExtensionContrib
 			// Counting tokens requires instantiating the tokenizers, which makes this process use a lot of memory.
 			// Let's cache the results across extension activations
 			const baseCount = await this._promptBaseCountCache.getBaseCount(endpoint);
-			let multiplierString = endpoint.multiplier !== undefined ? `${endpoint.multiplier}x` : undefined;
+			let modelDetail = endpoint.multiplier !== undefined ? `${endpoint.multiplier}x` : undefined;
```
johan-j, line 156: this was renamed to make it easier to follow the code. Its assigned to field `details` below.

src/extension/conversation/vscode-node/languageModelAccess.ts
```diff
@@ -153,19 +153,28 @@ export class LanguageModelAccess extends Disposable implements IExtensionContrib
...
-				multiplierString = 'Variable';
+				modelDetail = 'Variable';
+			}
+			let modelName = endpoint.name;
+			if (endpoint.customModel) {
+				modelDetail = endpoint.customModel.ownerName;
+				modelDescription = endpoint.customModel.keyName;
+				modelCategory = { label: localize('languageModelHeader.custom_models', "Custom Models"), order: 1 };
+
+				modelName = endpoint.name.replace(` · ${endpoint.customModel.keyName}`, '');
```
johan-j: In Copilot API, the name is added as:
`Name: fmt.Sprintf("%s · %s", m.DisplayName, m.Key.Name),`
So in here, if keyName is part of the string we remove it, since we add it already to the modelDescription.

We can clean this away in a future release if we update the API to only send the modelName.

src/platform/endpoint/node/chatEndpoint.ts
```diff
@@ -154,6 +154,7 @@ export class ChatEndpoint implements IChatEndpoint {
 	public readonly isPremium?: boolean | undefined;
 	public readonly multiplier?: number | undefined;
 	public readonly restrictedToSkus?: string[] | undefined;
+	public readonly customModel?: ICustomModel | undefined;
```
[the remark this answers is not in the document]
johan-j: following the code style of lines above

src/extension/conversation/vscode-node/languageModelAccess.ts
```diff
@@ -153,19 +153,29 @@ export class LanguageModelAccess extends Disposable implements IExtensionContrib
...
+				modelDetail = 'Variable';
+			}
+			let modelName = endpoint.name;
+			if (endpoint.customModel) {
+				const customModel = endpoint.customModel;
+				modelDetail = customModel.ownerName;
+				modelDescription = customModel.keyName;
+				modelCategory = { label: localize('languageModelHeader.custom_models', "Custom Models"), order: 1 };
+
+				modelName = endpoint.name.replace(` · ${customModel.keyName}`, '');
```
lramos15: Can we convince CAPI to not append key names like this? If they're worried about older clients than an API version would make sense here. I really dislike arbitrary string manipulation

src/platform/networking/common/networking.ts
```diff
@@ -151,6 +151,11 @@ export interface ICreateEndpointBodyOptions extends IMakeChatRequestOptions {
 	postOptions: OptionalChatRequestParams;
 }
 
+export interface ICustomModel {
```
lramos15: Lets just have whatever interface or type we use be the same as the ModelAPIResponse. Look at `ModelPolicy` for an example. 

This way we dont bother with maintaining two types

lramos15 (changes requested):

src/extension/conversation/vscode-node/languageModelAccess.ts
johan-j: yeah I will do the change in CAPI as soon as this update goes out instead! good call, keep this clean!
only a select few preview user got BYOK yet. We can ask them to update the plugin, rather than maintain the old style.

src/extension/conversation/vscode-node/languageModelAccess.ts
```diff
@@ -158,21 +158,18 @@ export class LanguageModelAccess extends Disposable implements IExtensionContrib
 			if (endpoint.model === AutoChatEndpoint.id) {
 				modelDetail = 'Variable';
 			}
-			let modelName = endpoint.name;
 			if (endpoint.customModel) {
 				const customModel = endpoint.customModel;
-				modelDetail = customModel.ownerName;
-				modelDescription = customModel.keyName;
+				modelDetail = customModel.owner_name;
+				modelDescription = customModel.key_name;
```
lramos15: As discussed in slack this should read `<MODEL> is contributed by <ORG> using <KEY_NAME>`. This can be properly localized as you see below

src/platform/endpoint/node/chatEndpoint.ts
```diff
@@ -195,8 +195,8 @@ export class ChatEndpoint implements IChatEndpoint {
 		this._policyDetails = _modelMetadata.policy;
 		if (_modelMetadata.custom_model) {
```
lramos15: Just `this.customModel = _modelMetadata.custom_model`. Same as policy

johan-j: ok thank you. Since we have 1 type now.

PR #1122, opened 2025-09-24
Cache agent

chrmarti:
https://github.com/microsoft/vscode/issues/269496

### DIFF
--- a/src/platform/networking/node/nodeFetchFetcher.ts
+++ b/src/platform/networking/node/nodeFetchFetcher.ts
@@ -6,6 +6,7 @@
 import * as undici from 'undici';
 import { IEnvService } from '../../env/common/envService';
 import { BaseFetchFetcher } from './baseFetchFetcher';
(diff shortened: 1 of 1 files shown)


src/platform/networking/node/nodeFetchFetcher.ts
```diff
@@ -32,6 +32,15 @@ export class NodeFetchFetcher extends BaseFetchFetcher {
 function getFetch(): typeof globalThis.fetch {
 	const fetch = (globalThis as any).__vscodePatchedFetch || globalThis.fetch;
 	return function (input: string | URL | globalThis.Request, init?: RequestInit) {
-		return fetch(input, { dispatcher: new undici.Agent({ allowH2: true }), ...init });
+		return fetch(input, { dispatcher: getAgent(), ...init });
```
dmitrivMS: Consider using Lazy<T> which seems to be an established pattern in the codebase.

dmitrivMS (dismissed):
Please see my comment inline

PR #1134, opened 2025-09-24
edits: fix file corruption issue with replace_string tool

connor4312:
The similarity matching in the replace_string tool incorrect was using
line numbers rather than string offsets (since inception!) which caused
file corruption issues. I initially thought it was only in the multi
edit tool, but it happens in all replace_string variants.

Closes https://github.com/microsoft/vscode/issues/265842


src/extension/tools/node/editFileToolUtils.tsx
```diff
@@ -349,42 +349,47 @@ function trySimilarityMatch(text: string, oldStr: string, newStr: string, eol: s
...
 	// Sliding window approach to find the best matching section
 	for (let i = 0; i <= lines.length - oldLines.length; i++) {
 		let totalSimilarity = 0;
+		let oldLength = 0;
 
 		// Calculate similarity for each line in the window
 		for (let j = 0; j < oldLines.length; j++) {
 			const similarity = calculateSimilarity(oldLines[j], lines[i + j]);
 			totalSimilarity += similarity;
+			oldLength += lines[i + j].length;
```
dmitrivMS, line 364: Based on name only - should this be adding length of oldLines and not lines array?

dmitrivMS (approved):

src/extension/tools/node/editFileToolUtils.tsx
connor4312: If it were an exact match, yes, but this is similarity matching and so the oldLine might not exactly equal the 'actual' old line we're matching against. The oldLength here is the actual length of the old data within the text document.

PR #1139, opened 2025-09-25
Built-in toolsets for default tools

24anisha:
Currently, we have a default toolset of 46 tools. Many of these tools are hyper-specified and often unnecessary for the task at hand. Their existence in the toolset ends up yielding worse behavior for many models, such as Sonnet 3.7 and GPT 4.1. As such, we're running an experiment to leverage the existing logic for MCP server toolsets to create built in toolsets from the default tools and only exposing a handful of tools to the user directly.

### GPT 4.1 Pass Rate
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns="http://www.w3.org/TR/REC-html40"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=PowerPoint.Slide>
<meta name=Generator content="Microsoft PowerPoint 15">

</head>

<body>

Toolset | Benchmark | Pass Rate (%)
-- | -- | --
small | MFA | 19.4
full | MFA | 13.4
small | SWEBench   C# | 8.1
full | SWEBench   C# | 7.9

</body>

</html>

### Sonnet 3.7 Success = True for Both

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns="http://www.w3.org/TR/REC-html40"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=PowerPoint.Slide>
<meta name=Generator content="Microsoft PowerPoint 15">

</head>

<body>

Case Name | Small Toolset    Steps | Full Toolset    Steps
-- | -- | --
gitmoji-cli-1248 | 7 | 18
isomorphic-git-1493 | 18 | 22
Luxon-1173 | 13 | 28
Prom-client-146 | 7 | 19

</body>

</html>

### DIFF
--- a/src/extension/tools/common/virtualTools/virtualToolGrouper.ts
+++ b/src/extension/tools/common/virtualTools/virtualToolGrouper.ts
@@ -29,12 +29,11 @@ const SUMMARY_PREFIX = 'Call this tool when you need access to a new category of
 const SUMMARY_SUFFIX = '\n\nBe sure to call this tool if you need a capability related to the above.';
 
 // categorize all tools except for the 11 default tools
-// 11 default tools = semantic_search, grep_search, read_file, create_file, apply_patch, replace_string_in_file,
-// insert_edit_into_file, run_in_terminal, list_dir, think, get_terminal_output
+// 12 default tools = semantic_search, grep_search, read_file, create_file, apply_patch, replace_string_in_file,
+// insert_edit_into_file, run_in_terminal, list_dir, think, get_terminal_output, manage_todo_list
 const BUILT_IN_TOOL_GROUPS = {
 	'Jupyter Notebook Tools': {
-		// "Call this tool when..."
-		summary: 'Tools for working with Jupyter notebooks - creating, editing, running cells, and managing notebook operations.',
+		summary: 'Call tools from this group when you need to work with Jupyter notebooks - creating, editing, running cells, and managing notebook operations.',
 		tools: [
 			'create_new_jupyter_notebook',
 			'edit_notebook_file',
@@ -50,7 +49,7 @@ const BUILT_IN_TOOL_GROUPS = {
 		]
 	},
 	'Web Interaction': {
-		summary: 'Tools for interacting with web content, browsing websites, and accessing external resources.',
+		summary: 'Call tools from this group when you need to interact with web content, browse websites, or access external resources.',
 		tools: [
 			'fetch_webpage',
 			'open_simple_browser',
@@ -58,13 +57,13 @@ const BUILT_IN_TOOL_GROUPS = {
 		]
 	},
 	'VS Code Interaction': {
-		summary: 'Tools for interacting with VS Code workspace and accessing VS Code features.',
+		summary: 'Call tools from this group when you need to interact with the VS Code workspace and access VS Code features.',
 		tools: [
 			'search_workspace_symbols',
 			'list_code_usages',
 			'get_errors',
 			'get_vscode_api',
-			// 'get_changed_files',
+			'get_changed_files',
 			'create_new_workspace',
 			'install_extension',
 			'get_project_setup_info',
@@ -72,25 +71,30 @@ const BUILT_IN_TOOL_GROUPS = {
 			'run_task',
 			'get_task_output',
 			'run_vscode_command',
-			// 'file_search',
-			// 'create_directory',
-			// 'multi_replace_string_in_file',
-			// 'terminal_selection',
-			// 'terminal_last_command',
+			'multi_replace_string_in_file',
 			'install_python_packages',
-			// 'manage_todo_list',
 			'get_search_view_results',
-			// 'edit_files',
-			'vscode_searchExtensions_internal'
+			'vscode_searchExtensions_internal',
+			'read_project_structure'
 		]
 	},
 	'Testing': {
-		summary: 'Tools for running tests, analyzing test failures, and managing test workflows.',
+		summary: 'Call tools from this group when you need to run tests, analyze test failures, and manage test workflows.',
 		tools: [
 			'run_tests',
 			'test_failure',
 			'test_search',
 			'runTests'
+		]
+	},
+	'Redundant but Specific': {
+		summary: 'These tools have overlapping functionalities but are highly specialized for certain tasks. \nTools: file_search, get_terminal_selection, get_terminal_last_command, create_directory, get_doc_info',
+		tools: [
+			'file_search',
+			'get_terminal_selection',
+			'get_terminal_last_command',
+			'create_directory',
+			'get_doc_info'
 		]
 	}
 } as const;
@@ -292,7 +296,7 @@ export class VirtualToolGrouper implements IToolCategorization {
 	private async _generateGroupsFromToolset(key: string, tools: LanguageModelToolInformation[], previous: ISummarizedToolCategory[] | undefined, token: CancellationToken): Promise<(VirtualTool | LanguageModelToolInformation)[]> {
 		// Handle built-in tools with predefined groups only if experimental setting is enabled
 		if (key === BUILT_IN_GROUP) {
-			const defaultToolGroupingEnabled = this._configurationService.getExperimentBasedConfig(ConfigKey.Internal.DefaultToolGrouping, this._expService);
+			const defaultToolGroupingEnabled = this._configurationService.getExperimentBasedConfig(ConfigKey.Internal.DefaultToolsGrouped, this._expService);
 			if (defaultToolGroupingEnabled) {
 				return this._createBuiltInToolGroups(tools);
 			} else {
--- a/src/extension/tools/common/virtualTools/virtualToolGrouper.ts
+++ b/src/extension/tools/common/virtualTools/virtualToolGrouper.ts
@@ -28,6 +28,73 @@ const CATEGORIZATION_ENDPOINT = CHAT_MODEL.GPT4OMINI;
 const SUMMARY_PREFIX = 'Call this tool when you need access to a new category of tools. The category of tools is described as follows:\n\n';
 const SUMMARY_SUFFIX = '\n\nBe sure to call this tool if you need a capability related to the above.';
 
+// categorize all tools except for the 11 default tools
+// 11 default tools = semantic_search, grep_search, read_file, create_file, apply_patch, replace_string_in_file,
+// insert_edit_into_file, run_in_terminal, list_dir, think, get_terminal_output
+const BUILT_IN_TOOL_GROUPS = {
+	'Jupyter Notebook Tools': {
+		// "Call this tool when..."
+		summary: 'Tools for working with Jupyter notebooks - creating, editing, running cells, and managing notebook operations.',
+		tools: [
+			'create_new_jupyter_notebook',
+			'edit_notebook_file',
+			'run_notebook_cell',
+			'copilot_getNotebookSummary',
+			'read_notebook_cell_output',
+			'configure_notebook',
+			'notebook_install_packages',
+			'notebook_list_packages',
+			'configure_python_environment',
+			'get_python_environment_details',
+			'get_python_executable_details'
+		]
+	},
+	'Web Interaction': {
+		summary: 'Tools for interacting with web content, browsing websites, and accessing external resources.',
+		tools: [
+			'fetch_webpage',
+			'open_simple_browser',
+			'github_repo'
+		]
+	},
+	'VS Code Interaction': {
+		summary: 'Tools for interacting with VS Code workspace and accessing VS Code features.',
+		tools: [
+			'search_workspace_symbols',
+			'list_code_usages',
+			'get_errors',
+			'get_vscode_api',
+			// 'get_changed_files',
+			'create_new_workspace',
+			'install_extension',
+			'get_project_setup_info',
+			'create_and_run_task',
+			'run_task',
+			'get_task_output',
+			'run_vscode_command',
+			// 'file_search',
+			// 'create_directory',
+			// 'multi_replace_string_in_file',
+			// 'terminal_selection',
+			// 'terminal_last_command',
+			'install_python_packages',
+			// 'manage_todo_list',
+			'get_search_view_results',
+			// 'edit_files',
+			'vscode_searchExtensions_internal'
+		]
+	},
+	'Testing': {
+		summary: 'Tools for running tests, analyzing test failures, and managing test workflows.',
+		tools: [
+			'run_tests',
+			'test_failure',
+			'test_search',
+			'runTests'
+		]
+	}
+} as const;
+
 export class VirtualToolGrouper implements IToolCategorization {
 	private readonly toolEmbeddingsComputer: ToolEmbeddingsComputer;
 
@@ -76,12 +143,9 @@ export class VirtualToolGrouper implements IToolCategorization {
 		const predictedToolsSw = new StopWatch();
 		const predictedToolsPromise = virtualToolEmbeddingRankingEnabled && this._getPredictedTools(query, tools, token).then(tools => ({ tools, durationMs: predictedToolsSw.elapsed() }));
 
+		// Now all toolsets (including built-in) go through _generateGroupsFromToolset
 		const grouped = await Promise.all(Object.entries(byToolset).map(([key, tools]) => {
-			if (key === BUILT_IN_GROUP) {
-				return tools;
-			} else {
-				return this._generateGroupsFromToolset(key, tools, previousCategorizations.get(key), token);
-			}
+			return this._generateGroupsFromToolset(key, tools, previousCategorizations.get(key), token);
 		}));
 
 		this._cache.flush();
@@ -226,6 +290,16 @@ export class VirtualToolGrouper implements IToolCategorization {
 
 	/** Top-level request to categorize a group of tools from a single source. */
 	private async _generateGroupsFromToolset(key: string, tools: LanguageModelToolInformation[], previous: ISummarizedToolCategory[] | undefined, token: CancellationToken): Promise<(VirtualTool | LanguageModelToolInformation)[]> {
+		// Handle built-in tools with predefined groups only if experimental setting is enabled
+		if (key === BUILT_IN_GROUP) {
+			const defaultToolGroupingEnabled = this._configurationService.getExperimentBasedConfig(ConfigKey.Internal.DefaultToolGrouping, this._expService);
+			if (defaultToolGroupingEnabled) {
+				return this._createBuiltInToolGroups(tools);
+			} else {
+				return tools;
+			}
+		}
+
 		if (tools.length <= Constant.MIN_TOOLSET_SIZE_TO_GROUP) {
 			return tools;
 		}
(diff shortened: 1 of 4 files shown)


24anisha:
Questions/Action items remaining:

- [x] Group the default tools without needing to alter the max toolset 
    - [x] BUG: only works for gpt-4.1 and gpt-5 because the max # tools is hardcoded to 28. Need to make it dynamic
    - [x] ISSUE: in toolGrouping.ts, isEnabled() is meant to be a fast check. So, we don't pass the chat endpoint in, which means we don't know the model family. If we do that, do non-GPT models get sent to the more costly compute step every query? How to manage?
- [x] Only apply this alteration for gpt models -- how to pass in the model family without changing the whole process?
- [x] Automatically add new tools to the built-in toolset groups (so that new tools aren't automatically considered part of the default toolset)
- [x] Confirm whether we should only apply changes for GPT models, or for all
- [x] remove unnecessary telemetry
- [x] remove isEnabled() function if it doesn't check anything meaningful
- [x] Manage duplicated logic
- [x] Re-adjust tool categorizations based on feedback
- [ ] What causes many tools to appear as core tools to the model, even though they don't show up in debugging?
- [x] edit_files -> core
- [x] handle the merge conflicts

src/extension/tools/common/virtualTools/toolGrouping.ts
```diff
@@ -47,7 +48,13 @@ export class ToolGrouping implements IToolGrouping {
 	}
 
 	public get isEnabled() {
-		return this._tools.length >= computeToolGroupingMinThreshold(this._experimentationService, this._configurationService).get();
+		// Match the logic from VirtualToolGrouper.addGroups()
+		// Enable if we could potentially trigger built-in grouping (when GPT model is used)
+		const defaultToolGroupingEnabled = this._configurationService.getExperimentBasedConfig(ConfigKey.Internal.DefaultToolsGrouped, this._experimentationService);
+		const couldTriggerBuiltInGrouping = this._tools.length > Constant.START_BUILTIN_GROUPING_AFTER_TOOL_COUNT && defaultToolGroupingEnabled;
+
```
24anisha: This doesn't check everything we need. We also need to confirm that the model is gpt-4.1 or gpt-5, but this is supposed to be a lightweight, quick function to call. 
Options:
- pass the endpoint into this function so we can check to make sure it's the correct model before going to virtual tool grouping
- let it go to virtual tool grouping every time (since START_BUILTIN_GROUPING_AFTER_TOOL_COUNT is 20) and get stopped from grouping there by the endpoint check

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+const SUMMARY_PREFIX = 'Call this tool when you need access to a new category of tools. The category of tools is described as follows:\n\n';
+const SUMMARY_SUFFIX = '\n\nBe sure to call this tool if you need a capability related to the above.';
+
+// categorize all tools except for the 12 default tools
+// 12 default tools = semantic_search, grep_search, read_file, create_file, apply_patch, replace_string_in_file,
+// insert_edit_into_file, run_in_terminal, list_dir, think, get_terminal_output, manage_todo_list
+const BUILT_IN_TOOL_GROUPS = {
+	'Jupyter Notebook Tools': {
+		summary: 'Call tools from this group when you need to work with Jupyter notebooks - creating, editing, running cells, and managing notebook operations.',
+		tools: [
```
connor4312: tool names are here: https://github.com/microsoft/vscode-copilot-chat/blob/8f7155721315874aaacfca8d1f75feedfa9aa6d6/src/extension/tools/common/toolNames.ts#L8

You can ensure categorization with a type like `const toolCategories: Record<ToolName, ToolCategory> = {};`. For any tools that are not part of that enum you'll want an additional `externalToolCategories: Record<string, ToolCategory>` and merge with that

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+			'fetch_webpage',
+			'open_simple_browser',
+			'github_repo'
+		]
+	},
+	'VS Code Interaction': {
+		summary: 'Call tools from this group when you need to interact with the VS Code workspace and access VS Code features.',
+		tools: [
+			'search_workspace_symbols',
+			'list_code_usages',
```
connor4312: These two are general search tools or might even be top-level tools

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+			'install_extension',
+			'get_project_setup_info',
+			'create_and_run_task',
+			'run_task',
+			'get_task_output',
+			'run_vscode_command',
+			'install_python_packages',
+			'get_search_view_results',
+			'vscode_searchExtensions_internal',
+			'read_project_structure'
```
connor4312: Same here

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+			'list_code_usages',
+			'get_errors',
+			'get_vscode_api',
+			'get_changed_files',
+			'create_new_workspace',
+			'install_extension',
+			'get_project_setup_info',
+			'create_and_run_task',
+			'run_task',
+			'get_task_output',
```
connor4312: Tasks are important workflows for projects that use them. I think having create_and_run_task in here is fine because that is unconditionally enabled, but run_Task and get_task_output are already conditionally enabled when tasks are enabled, and having them be top level when that's the case is good.

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+	'Redundant but Specific': {
+		summary: 'These tools have overlapping functionalities but are highly specialized for certain tasks. \nTools: file_search, terminal_selection, terminal_last_command, create_directory, get_doc_info, multi_replace_string_in_file, edit_files',
+		tools: [
+			'file_search',
+			'terminal_selection',
+			'terminal_last_command',
+			'create_directory',
+			'get_doc_info',
+			'edit_files',
+			'multi_replace_string_in_file'
```
connor4312: Editing tools should always be top level

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+			'run_tests',
+			'test_failure',
+			'test_search',
+			'runTests'
+		]
+	},
+	'Redundant but Specific': {
+		summary: 'These tools have overlapping functionalities but are highly specialized for certain tasks. \nTools: file_search, terminal_selection, terminal_last_command, create_directory, get_doc_info, multi_replace_string_in_file, edit_files',
+		tools: [
+			'file_search',
```
connor4312: File search is heavily used by all agents and we should not put it behind a group

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+			'terminal_selection',
+			'terminal_last_command',
+			'create_directory',
```
connor4312: I think this should be top-level as well. Without this the agent will use the terminal instead, requiring user approval, which is a worse experience.

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+			'test_search',
+			'runTests'
+		]
+	},
+	'Redundant but Specific': {
+		summary: 'These tools have overlapping functionalities but are highly specialized for certain tasks. \nTools: file_search, terminal_selection, terminal_last_command, create_directory, get_doc_info, multi_replace_string_in_file, edit_files',
+		tools: [
+			'file_search',
+			'terminal_selection',
+			'terminal_last_command',
```
connor4312: These can probably be in the vscode group. Generally I think they old get called if a user manually references them.

src/platform/configuration/common/configurationService.ts
```diff
@@ -735,6 +735,7 @@ export namespace ConfigKey {
 		export const PromptFileContext = defineExpSetting<boolean>('chat.advanced.promptFileContextProvider.enabled', true);
 		export const MultiReplaceString = defineExpSetting<boolean>('chat.advanced.multiReplaceString.enabled', false, INTERNAL);
 
+		export const DefaultToolsGrouped = defineExpSetting<boolean>('chat.advanced.tools.defaultToolsGrouped', true, INTERNAL);
```
connor4312: This should be off by default until we evaluate its impact.

src/extension/tools/common/virtualTools/virtualToolGrouper.ts
```diff
@@ -42,15 +45,26 @@ export class VirtualToolGrouper implements IToolCategorization {
...
 	private get virtualToolEmbeddingRankingEnabled() {
 		return this._configurationService.getExperimentBasedConfig(ConfigKey.Internal.VirtualToolEmbeddingRanking, this._expService);
 	}
 
-	async addGroups(query: string, root: VirtualTool, tools: LanguageModelToolInformation[], token: CancellationToken): Promise<void> {
+	async addGroups(query: string, root: VirtualTool, tools: LanguageModelToolInformation[], token: CancellationToken, endpoint?: IChatEndpoint): Promise<void> {
 		// If there's no need to group tools, just add them all directly;
-		if (tools.length < Constant.START_GROUPING_AFTER_TOOL_COUNT) {
+
+		// if the model is gpt 4.1 or gpt-5 and there are more than START_BUILTIN_GROUPING_AFTER_TOOL_COUNT tools, we should group built-in tools
```
connor4312: Does Sonnet not see the same benefits?

src/extension/tools/common/virtualTools/virtualToolTypes.ts
```diff
@@ -53,12 +54,12 @@ export interface IToolGrouping {
...
 	 * Internally re-reads the request and conversation state.
 	 */
-	compute(query: string, token: CancellationToken): Promise<LanguageModelToolInformation[]>;
+	compute(query: string, token: CancellationToken, endpoint?: IChatEndpoint): Promise<LanguageModelToolInformation[]>;
 
 	/**
 	 * Returns the complete tree of tools, used for diagnostic purposes.
 	 */
-	computeAll(query: string, token: CancellationToken): Promise<(LanguageModelToolInformation | VirtualTool)[]>;
+	computeAll(query: string, token: CancellationToken, endpoint?: IChatEndpoint): Promise<(LanguageModelToolInformation | VirtualTool)[]>;
```
connor4312: Should endpoint be required here? I don't think there's any consumer that is calling this outside the context of a specific endpoint...

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+				);
+				virtualTools.push(virtualTool);
+			}
+		}
+
+		// Add any remaining uncategorized tools individually
+		const uncategorizedTools = tools.filter(tool => !usedTools.has(tool.name));
+
+		// Send telemetry for built-in tool grouping
+		this._telemetryService.sendMSFTTelemetryEvent('virtualTools.generate', {
```
connor4312: This needs a GDPR comment

src/extension/tools/common/virtualTools/toolGrouping.ts
```diff
@@ -47,7 +48,13 @@ export class ToolGrouping implements IToolGrouping {
 	}
 
 	public get isEnabled() {
```
connor4312: As mentioned we shuld get rid of the isEnabled check if it doesn't reflect accurately and just handle grouping internall correctly

src/extension/tools/common/virtualTools/virtualToolGrouper.ts
```diff
@@ -237,7 +258,33 @@ export class VirtualToolGrouper implements IToolCategorization {
 	}
 
 	/** Top-level request to categorize a group of tools from a single source. */
-	private async _generateGroupsFromToolset(key: string, tools: LanguageModelToolInformation[], previous: ISummarizedToolCategory[] | undefined, token: CancellationToken): Promise<(VirtualTool | LanguageModelToolInformation)[]> {
+	private async _generateGroupsFromToolset(key: string, tools: LanguageModelToolInformation[], previous: ISummarizedToolCategory[] | undefined, token: CancellationToken, endpoint?: IChatEndpoint): Promise<(VirtualTool | LanguageModelToolInformation)[]> {
+		// Handle built-in tools with predefined groups only if experimental setting is enabled
+		if (key === BuiltInToolGroupHandler.BUILT_IN_GROUP_KEY) {
+			const defaultToolGroupingEnabled = this._configurationService.getExperimentBasedConfig(ConfigKey.Internal.DefaultToolsGrouped, this._expService);
+			if (defaultToolGroupingEnabled) {
```
connor4312: As copilot mentioned it looks like we duplicated the logic here

connor4312:
I'm also curious--what trajectory difference you saw in the small vs large toolset in your runs? Specifically for GPT-5/Sonnet. 4.1 is going away soon, we don't care too much about that any longer.

src/extension/tools/common/virtualTools/virtualToolGrouper.ts
24anisha: sonnet 3.7 does, but sonnet 4 seemed robust to the larger toolset. Also, I think we decided to avoid editing the tools for the sonnet models because of prompt caching (I believe the tools are added at the top of a request for sonnet models, so it would cause a lot of cache misses to regularly change the toolset)

24anisha:
> I'm also curious--what trajectory difference you saw in the small vs large toolset in your runs? Specifically for GPT-5/Sonnet. 4.1 is going away soon, we don't care too much about that any longer.

If you look at the second table in the PR description, you can see the deltas for GPT 5. I'm doing an in-depth qualitative dive into the trajectories today, so I should have more behavioral discrepancies to report shortly.

src/extension/tools/common/virtualTools/virtualToolGrouper.ts
connor4312: The cache behavior is the case with all our current models.

We should have these checks in the central chatModelCapabilities.ts file

24anisha:
> I'm also curious--what trajectory difference you saw in the small vs large toolset in your runs? Specifically for GPT-5/Sonnet. 4.1 is going away soon, we don't care too much about that any longer.

BTW, I completed a qualitative analysis of the GPT-5 small vs. large toolset trajectories. In MFA (where full toolset scored higher than small by 2 cases), the only tool that was regularly called in the large toolset runs that the small toolset did not have was create_and_run_task. But it didn't seem to provide that much value to the model (cases where create_and_run_task was called were not more likely to succeed). Besides that, the small toolset regularly performed at or above the level of the full toolset.

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
24anisha: is edit_files an editing tool? The description in the package.json is "This is a placeholder tool, do not use" so I figured I shouldn't add it to the core toolset

src/extension/tools/common/virtualTools/virtualToolTypes.ts
24anisha: I passed it in because _doCompute() (which is called inside the computeAll function) requires it. I can just pass in `undefined` for computeAll, if we don't want to display the built in tool groupings

24anisha:
@connor4312 question for you: when I'm debugging, I see the following tools appearing in the core toolset:
`vscode_searchExtensions_internal`, `configure_notebook`, `notebook_install_packages`, `notebook_list_packages`, `configure_python_environment`, `get_python_environment_details`, `get_python_executable_details`, `install_python_packages`. How do these tools relate to the tools in the builtin vscode toolset? What do we want to do with them?

24anisha:
> @connor4312 question for you: when I'm debugging, I see the following tools appearing in the core toolset: `vscode_searchExtensions_internal`, `configure_notebook`, `notebook_install_packages`, `notebook_list_packages`, `configure_python_environment`, `get_python_environment_details`, `get_python_executable_details`, `install_python_packages`. How do these tools relate to the tools in the builtin vscode toolset? What do we want to do with them?

For more context, here are the listed "core" tools in the debugger:
Screenshot 2025-10-10 at 4 03 25 PM

And here is the copilotmd log of the request from the extension development host:
[debuggingRunManyTools.md](https://github.com/user-attachments/files/22858915/showAllTools.md.copilotmd)

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
24anisha: In restricted telemetry, I see that they get called extremely rarely (800 and 3000 times in the last 3 days, respectively, compared to ~3 million for something like grep_search), so I don't think they make sense at the top level. Maybe I could put them in redundant but specific instead of the vs code toolset?

src/extension/tools/common/toolNames.ts
```diff
@@ -137,3 +150,107 @@ export function mapContributedToolNamesInString(str: string): string {
...
+ * Get the category for a tool, checking both ToolName enum and external tools.
+ */
+export function getToolCategory(toolName: string): ToolCategory | undefined {
+	// First check if it's a known ToolName enum value
+	for (const [, enumValue] of Object.entries(ToolName)) {
+		if (enumValue === toolName) {
+			return toolCategories[enumValue];
+		}
+	}
+}
```
connor4312, line 232:
```suggestion
export function getToolCategory(toolName: string): ToolCategory | undefined {
	return toolCategories.hasOwnProperty(toolName) ? toolCategories[toolName] : undefined
}
```

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,94 @@
...
+					},
+					groupTools
+				);
+				virtualTools.push(virtualTool);
+			}
+		}
+
+		// Add any remaining uncategorized tools individually
+		const uncategorizedTools = tools.filter(tool => !usedTools.has(tool.name));
+		return [...virtualTools, ...uncategorizedTools];
```
connor4312: I'd maybe do something like this, a little lighter and preserves the `Constant.MIN_TOOLSET_SIZE_TO_GROUP` logic.

```suggestion

		const contributedTools = tools.filter(t => !toolCategories.hasOwnProperty(t.name));
		const builtInTools = tools.filter(t => toolCategories.hasOwnProperty(t.name));

		const categories = groupBy(builtInTools, t => toolCategories[t.name as ToolName]);
		const virtualTools = Object.entries(categories).flatMap<VirtualTool|LanguageModelToolInformation>(([category, tools]) => {
			if (tools.length <= Constant.MIN_TOOLSET_SIZE_TO_GROUP) {
				return tools;
			}

			return new VirtualTool(
				VIRTUAL_TOOL_NAME_PREFIX + category.toLowerCase().replace(/\s+/g, '_'),
				SUMMARY_PREFIX + getCategorySummary(category) + SUMMARY_SUFFIX,
				0,
				{
					toolsetKey: BUILT_IN_GROUP,
					possiblePrefix: 'builtin_',
					wasExpandedByDefault: false,
					canBeCollapsed: true
				},
				tools
			);
		});

		return [...virtualTools, ...contributedTools];
```

src/extension/tools/common/virtualTools/virtualToolTypes.ts
```diff
@@ -53,12 +49,12 @@ export interface IToolGrouping {
...
```
connor4312: We don't use the endpoint anymore, should it be removed from these interfaces?

src/extension/tools/common/virtualTools/virtualToolTypes.ts
```diff
@@ -103,7 +99,7 @@ export interface IToolCategorization {
 	 * Called whenever new tools are added. The function should add each tool into
 	 * the appropriate virtual tool or top-level tool in the `root`.
 	 */
-	addGroups(query: string, root: VirtualTool, tools: LanguageModelToolInformation[], token: CancellationToken): Promise<void>;
+	addGroups(query: string, root: VirtualTool, tools: LanguageModelToolInformation[], token: CancellationToken, endpoint?: IChatEndpoint): Promise<void>;
```
connor4312: same ^

src/extension/tools/common/virtualTools/virtualTool.ts
```diff
@@ -12,6 +12,8 @@ export interface IVirtualToolMetadata {
 	wasEmbeddingsMatched?: boolean;
 	wasExpandedByDefault?: boolean;
 	canBeCollapsed?: boolean;
+	toolsetKey?: string;
+	possiblePrefix?: string;
```
connor4312: these two aren't used any more, should be safe to delete them

src/extension/tools/common/virtualTools/toolGrouping.ts
```diff
@@ -131,6 +125,7 @@ export class ToolGrouping implements IToolGrouping {
 	}
 
 	async computeAll(query: string, token: CancellationToken): Promise<(LanguageModelToolInformation | VirtualTool)[]> {
+		// Don't pass endpoint for display - prevents model-specific built-in grouping
```
connor4312:
```suggestion
```

I think this is gone now

src/extension/tools/common/virtualTools/virtualToolGrouper.ts
```diff
@@ -67,18 +88,28 @@ export class VirtualToolGrouper implements IToolCategorization {
...
+		}
 
 		// Process extension/MCP tools per-toolset with proportional slot allocation
 		if (toolsetEntries.length > 0) {
-			// Calculate available slots after accounting for builtin tools
-			const availableSlots = TOOLS_AND_GROUPS_LIMIT - builtinTools.length;
+			// Calculate available slots after accounting for builtin tools/groups
+			const builtinSlotCount = shouldGroupBuiltin
+				? groupedResults.filter(item => item instanceof VirtualTool).length
+				: builtinTools.length;
```
connor4312: I think here we should just take the `groupedResults.length` (which used to be equivalent to `builtinTools.length`). We want to distribute the number of 'slots' that other tools get based on how many are left over

connor4312 (approved):

bpasero (dismissed):

src/extension/tools/common/virtualTools/builtInToolGroupHandler.ts
```diff
@@ -0,0 +1,84 @@
...
+		case ToolCategory.Testing:
+			return 'Call tools from this group when you need to run tests, analyze test failures, and manage test workflows.';
+		case ToolCategory.RedundantButSpecific: {
+			const toolNames = getToolsForCategory(category);
+			return `These tools have overlapping functionalities but are highly specialized for certain tasks. Tools: ${toolNames.join(', ')}`;
+		}
+		case ToolCategory.Core:
+			return 'Core tools that should always be available without grouping.';
+		default:
+			return 'Tools in this category.';
```
connor4312: You can use `assertNever(category)` to have a compile-time assert that all categories are described.

connor4312 (approved):

PR #1145, opened 2025-09-25
Cache the base prompt token counts across extension activations

alexdima:
Fixes microsoft/vscode#268310


alexdima:
> Thanks! I guess a slight downside here is that we're always invalidating on updates which in insiders / pre-release can be rough because the prompt doesn't change very often. Wondering if we should be smarter with some sort of versioning of the prompt

Yes! One thing we could do is run a job in CI during packaging and add these numbers to the `package.json`, but then I thought that models are dynamic, some people have some, some people have others, then there is BYOK... 

Also sorry I forgot to ping you on this! I wanted to write a Slack message containing multiple memory / perf improvements.

PR #1157, opened 2025-09-26
Add security-sensitive file extensions to workspace indexing exclusion list

phawrylak:
This pull request updates the list of excluded file extensions in the `workspaceFileIndex.ts` file to improve security and prevent indexing of sensitive files. The change specifically adds several certificate and private key file types to the exclusion list.

Security-related exclusions:

* Added certificate and private key extensions (`pfx`, `p12`, `pem`, `crt`, `cer`, `key`, `priv`, `jks`, `keystore`, `csr`) to the `EXCLUDE_EXTENSIONS` set in `workspaceFileIndex.ts` to prevent indexing of security-sensitive files.


mjbvz (approved):
Thanks for adding these. I'll also confirm our indexing services ignore these. I believe it should already but good to make sure

roblourens (approved):

PR #1187, opened 2025-09-29
WIP: use a the chat session provider to replay the exported chat logs

amunger:
https://github.com/microsoft/vscode-copilot-evaluation/issues/357

This would make the chat look more natural and would enable us to drop the additional intent.

The API changed as I was working through this though, so it is broken in its current state.

Also, hitting F5 to debug the replay file will currently cause an infinite loop - just use the replay command to test in this state.


src/extension/replay/vscode-node/editHelper.ts
```diff
@@ -0,0 +1,70 @@
...
+			doc.positionAt(replacement.replaceRange.endExclusive)
+		);
+
+		workspaceEdit.replace(uri, range, replacement.newText);
+
+		for (const textEdit of workspaceEdit.entries()) {
+			const edits = Array.isArray(textEdit[1]) ? textEdit[1] : [textEdit[1]];
+			for (const textEdit of edits) {
+				try {
+					stream.textEdit(uri, textEdit);
```
amunger, line 63: with this solution, I'm seeing that the stream passed to the ChatSessionProvider is not able to use streamed changes in the same way that a contributed intent can, I saw an edit happen once, but without any accept/reject UI, but mostly just ignores any edits that are given to the stream. Most likely by design, but would there be a way to enable that instead? @joshspicer @rebornix 

It's also frequently failing to just open the file with the workspace service though, so maybe it's a different issue?

PR #1225, opened 2025-10-02
Fix: Cache network agent instance

chrmarti:
https://github.com/microsoft/vscode/issues/269496


TylerLeonhardt (approved):

TylerLeonhardt:
This was discussed on the Zurich side. They did programatic testing, but since it's Thursday and the VS Code release has gone out already, we need to get this one out.

PR #1231, opened 2025-10-03
feat: add custom `requestHeaders` support for custom OAI and Azure models in BYOK

zhichli:


src/extension/byok/node/openAIEndpoint.ts
```diff
@@ -96,39 +127,91 @@ export class OpenAIEndpoint extends ChatEndpoint {
...
+			// Use normalized lowercase key for storage to ensure consistency
 			sanitized[key] = sanitizedValue;
+			processedCount++;
 		}
+
 		return sanitized;
 	}
 
-	private _sanitizeHeaderValue(value: string, key: string): string | undefined {
+	private _sanitizeHeaderValue(value: unknown): string | undefined {
```
lramos15, line 207: Are we not sanitzing twice here? Both above and with this function

zhichli: So _sanitizeHeaderValue() is a helper function called by _sanitizeCustomHeaders() during initialization, not a separate sanitization step. The headers are sanitized once during object construction and then reused safely throughout the object's lifetime.

The separation is: _sanitizeCustomHeaders() handles the overall header collection logic (limits, reserved headers, **key** validation) while _sanitizeHeaderValue() focuses specifically on **value** sanitization (control characters, length limits, etc.)."

PR #1232, opened 2025-10-03
feat: Hide editFile tool message

bryanchen-d:
Hide messages of edit file tools like
`Using "Apply Patch"`
`Using "Replace String"`
to reduce the chat clutter as after the message we already show the files being modified.

This is to fix https://github.com/microsoft/vscode/issues/265037

### DIFF
--- a/src/extension/vscode.proposed.chatParticipantPrivate.d.ts
+++ b/src/extension/vscode.proposed.chatParticipantPrivate.d.ts
@@ -236,7 +236,7 @@ declare module 'vscode' {
 
 	export interface PreparedToolInvocation {
 		pastTenseMessage?: string | MarkdownString;
-		presentation?: 'hidden' | undefined;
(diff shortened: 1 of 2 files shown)


src/extension/vscode.proposed.chatParticipantPrivate.d.ts
```diff
@@ -236,7 +236,7 @@ declare module 'vscode' {
 
 	export interface PreparedToolInvocation {
 		pastTenseMessage?: string | MarkdownString;
-		presentation?: 'hidden' | undefined;
+		presentation?: 'hidden' | 'hiddenAfterComplete' | undefined;
```
roblourens, line 239: The source of truth for extension API is the vscode repo, so there should be a matching PR there that updates this

bryanchen-d: added a matching PR to update the API: https://github.com/microsoft/vscode/pull/270052

PR #1233, opened 2025-10-03
feat: add timeout in searching text in files

bryanchen-d:
Add timeout in findFiles and findTextInFiles tool so they won't hang the chat.

Fixes: https://github.com/microsoft/vscode/issues/262743

### DIFF
--- a/src/extension/tools/node/findFilesTool.tsx
+++ b/src/extension/tools/node/findFilesTool.tsx
@@ -11,7 +11,7 @@ import { URI } from '../../../util/vs/base/common/uri';
 import * as l10n from '@vscode/l10n';
 import { ISearchService } from '../../../platform/search/common/searchService';
 import { IWorkspaceService } from '../../../platform/workspace/common/workspaceService';
-import { raceCancellationError, raceTimeout } from '../../../util/vs/base/common/async';
+import { raceTimeoutAndCancellationError } from '../../../util/common/racePromise';
 import { CancellationToken, CancellationTokenSource } from '../../../util/vs/base/common/cancellation';
 import { IInstantiationService } from '../../../util/vs/platform/instantiation/common/instantiation';
 import { ExtendedLanguageModelToolResult, LanguageModelPromptTsxPart, MarkdownString } from '../../../vscodeTypes';
@@ -48,19 +48,10 @@ export class FindFilesTool implements ICopilotTool<IFindFilesToolParams> {
 		// also in the case of the parent token being cancelled, it will cancel this one too
 		const searchCancellation = new CancellationTokenSource(token);
 
-		async function raceTimeoutAndCancellationError<T>(promise: Promise<T>, timeoutMessage: string): Promise<T> {
-			const result = await raceTimeout(raceCancellationError(promise, token), timeoutInMs);
-			if (result === undefined) {
-				// we have timed out, so cancel the search
-				searchCancellation.cancel();
-				throw new Error(timeoutMessage);
-			}
-
-			return result;
-		}
-
 		const results = await raceTimeoutAndCancellationError(
 			Promise.resolve(this.searchService.findFiles(pattern, undefined, searchCancellation.token)),
+			searchCancellation,
+			timeoutInMs,
 			'Timeout in searching files, try a more specific search pattern'
 		);
 
--- a/src/extension/tools/node/findFilesTool.tsx
+++ b/src/extension/tools/node/findFilesTool.tsx
@@ -61,11 +61,6 @@ export class FindFilesTool implements ICopilotTool<IFindFilesToolParams> {
 		const resultsToShow = results.slice(0, maxResults);
 		// Render the prompt element with a timeout
 		const prompt = await renderPromptElementJSON(this.instantiationService, FindFilesResult, { fileResults: resultsToShow, totalResults: results.length }, options.tokenizationOptions, token);
-
-		if (prompt === undefined) {
-			throw new Error('Timeout in rendering prompt');
-		}
-
 		const result = new ExtendedLanguageModelToolResult([new LanguageModelPromptTsxPart(prompt)]);
 		const query = `\`${options.input.query}\``;
 		result.toolResultMessage = resultsToShow.length === 0 ?
--- a/src/extension/tools/node/findFilesTool.tsx
+++ b/src/extension/tools/node/findFilesTool.tsx
@@ -11,7 +11,7 @@ import { URI } from '../../../util/vs/base/common/uri';
 import * as l10n from '@vscode/l10n';
 import { ISearchService } from '../../../platform/search/common/searchService';
 import { IWorkspaceService } from '../../../platform/workspace/common/workspaceService';
-import { raceCancellation, raceTimeout } from '../../../util/vs/base/common/async';
+import { raceCancellationError, raceTimeout } from '../../../util/vs/base/common/async';
 import { CancellationToken } from '../../../util/vs/base/common/cancellation';
 import { IInstantiationService } from '../../../util/vs/platform/instantiation/common/instantiation';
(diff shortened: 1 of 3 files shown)


src/extension/tools/node/findTextInFilesTool.tsx
```diff
@@ -51,14 +51,28 @@ export class FindTextInFilesTool implements ICopilotTool<IFindTextInFilesToolPar
...
 		const isRegExp = options.input.isRegexp ?? true;
 		const queryIsValidRegex = this.isValidRegex(options.input.query);
-		let results = await this.searchAndCollectResults(options.input.query, isRegExp, patterns, maxResults, token);
+
+		// try find text with a timeout of 10s
+		// TODO: consider making the timeout configurable
+		const timeoutInMs = 10_000;
+		let results = await Promise.race([
+			this.searchAndCollectResults(options.input.query, isRegExp, patterns, maxResults, token),
+			new Promise<never>((_, reject) => setTimeout(() => reject(new Error("Timeout in searching text in files")), timeoutInMs))]);
```
roblourens: Check whether any helpers in async.ts could be reused here- `timeout`, `raceCancellation`, etc. If not that's fine, it's pretty simple.

bryanchen-d: There are`raceTimeout`, `timeout` utilities, but they just resolve to `undefined` when time's up, while in this case I think we want to throw an error similar to the cancellation case. Do you think we should make a new utility for that?

src/extension/tools/node/findFilesTool.tsx
```diff
@@ -40,14 +41,34 @@ export class FindFilesTool implements ICopilotTool<IFindFilesToolParams> {
 		// The input _should_ be a pattern matching inside a workspace, folder, but sometimes we get absolute paths, so try to resolve them
 		const pattern = inputGlobToPattern(options.input.query, this.workspaceService);
 
-		const results = await this.searchService.findFiles(pattern, undefined, token);
+		// try find files with a timeout of 10s
+		// TODO: consider making the timeout configurable
+		const timeoutInMs = 10_000;
+		const results = await raceTimeout(
+			raceCancellation(
+				Promise.resolve(this.searchService.findFiles(pattern, undefined, token)),
```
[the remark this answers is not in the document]
bryanchen-d: Unfortunately the `findFiles` returns a `Thenable<>` which doesn't fit the required `Promise<>` type, and fails compilation.

src/extension/tools/node/findTextInFilesTool.tsx
```diff
@@ -51,14 +52,41 @@ export class FindTextInFilesTool implements ICopilotTool<IFindTextInFilesToolPar
...
+				raceCancellation(
+					this.searchAndCollectResults(options.input.query, !isRegExp, patterns, maxResults, token),
+					token
+				),
+				timeoutInMs
+			);
+		}
+
+		if (results === undefined) {
+			throw new Error('Timeout in searching files');
```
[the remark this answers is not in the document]
bryanchen-d: updated.

src/extension/tools/node/findFilesTool.tsx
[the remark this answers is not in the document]
roblourens: It would make sense for the search service to return a Promise instead (this comes from our extension API but our searchService could do the `Promise.resolve`. You can do that if it makes things cleaner, you don't have to

src/extension/tools/node/findTextInFilesTool.tsx
roblourens: I see, it's fine

roblourens: Is `timeout(ms).then(() => Promise.reject(new Error("..."))` cleaner?

src/extension/tools/node/findFilesTool.tsx
```diff
@@ -40,14 +41,38 @@ export class FindFilesTool implements ICopilotTool<IFindFilesToolParams> {
 		// The input _should_ be a pattern matching inside a workspace, folder, but sometimes we get absolute paths, so try to resolve them
 		const pattern = inputGlobToPattern(options.input.query, this.workspaceService);
 
-		const results = await this.searchService.findFiles(pattern, undefined, token);
+		// try find files with a timeout of 10s
+		// TODO: consider making the timeout configurable
```
roblourens: I don't think there's a need for it to be configurable

src/extension/tools/node/findFilesTool.tsx
```diff
@@ -40,14 +41,38 @@ export class FindFilesTool implements ICopilotTool<IFindFilesToolParams> {
...
+
 		checkCancellation(token);
 
 		const maxResults = options.input.maxResults ?? 20;
 		const resultsToShow = results.slice(0, maxResults);
-		const result = new ExtendedLanguageModelToolResult([
-			new LanguageModelPromptTsxPart(
-				await renderPromptElementJSON(this.instantiationService, FindFilesResult, { fileResults: resultsToShow, totalResults: results.length }, options.tokenizationOptions, token))]);
+		// Render the prompt element with a timeout
+		const prompt = await raceTimeout(
```
roblourens: I don't think there's a need for prompt rendering to have a timeout. We do this in many places and I don't remember having an issue with it.

src/extension/tools/node/findTextInFilesTool.tsx
```diff
@@ -99,6 +131,7 @@ export class FindTextInFilesTool implements ICopilotTool<IFindTextInFilesToolPar
 			token);
 		const results: vscode.TextSearchResult2[] = [];
 		for await (const item of searchResult.results) {
+			await new Promise(r => setTimeout(r, 5000)); // yield to allow cancellation to be processed
```
roblourens: What's this? We need to wait 5s?

bryanchen-d: oops, this shouldn't be checked in, I was using it to validate the timeout case

src/extension/tools/node/findFilesTool.tsx
bryanchen-d: updated.

src/extension/tools/node/findFilesTool.tsx
bryanchen-d: updated removing the timeout guard around rendering prompt

src/extension/tools/node/findTextInFilesTool.tsx
```diff
@@ -210,8 +245,8 @@ interface IFindMatchProps extends BasePromptElementProps {
  * 1. Removes excessive extra character data from the match, e.g. avoiding
  * giant minified lines
  * 2. Wraps the match in a <match> tag
- * 3. Prioritizes lines in the middle of the match where the range lies
- */
+								* 3. Prioritizes lines in the middle of the match where the range lies
```
roblourens: Accident?

roblourens:
Please link the issue from the PR so we can keep track of what change goes with which issue. If you say `fixes microsoft/vscode#123` then the issue will be automatically closed when the PR is merged. https://github.com/microsoft/vscode/issues/262743

src/extension/tools/node/findTextInFilesTool.tsx
bryanchen-d: extracted a `raceTimeAndCancellationError` utility

src/extension/tools/node/findFilesTool.tsx
```diff
@@ -48,19 +48,10 @@ export class FindFilesTool implements ICopilotTool<IFindFilesToolParams> {
 		// also in the case of the parent token being cancelled, it will cancel this one too
 		const searchCancellation = new CancellationTokenSource(token);
```
roblourens: The `CancellationTokenSource` needs to be disposed when we're done with it

roblourens (approved):

PR #1234, opened 2025-10-06
New internal only repository telemetry request from science team

IanMatthewHuff:
This PR added in repository diff information to be logged in telemetry for only internal MSFT users on request of the science team.

This information is logged at the start of the request loop and at the end of the tool calling loop. More detailed comments at points of interest in-line. @sbatten has context on the discussion there.

### DIFF
--- a/src/extension/agents/copilotcli/node/copilotcliSessionService.ts
+++ b/src/extension/agents/copilotcli/node/copilotcliSessionService.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import type { Session, SessionManager } from '@github/copilot/sdk';
-import type { CancellationToken } from 'vscode';
+import type { CancellationToken, ChatContext, ChatRequest } from 'vscode';
 import { IEnvService } from '../../../../platform/env/common/envService';
 import { IVSCodeExtensionContext } from '../../../../platform/extContext/common/extensionContext';
 import { ILogService } from '../../../../platform/log/common/logService';
@@ -19,6 +19,8 @@ export interface ICopilotCLISession {
 	readonly label: string;
 	readonly timestamp: Date;
 }
+
+export type ExtendedChatRequest = ChatRequest & { prompt: string };
 
 export interface ICopilotCLISessionService {
 	readonly _serviceBrand: undefined;
@@ -35,6 +37,11 @@ export interface ICopilotCLISessionService {
 	// Session wrapper tracking
 	trackSessionWrapper<T extends IDisposable>(sessionId: string, wrapper: T): void;
 	findSessionWrapper<T extends IDisposable>(sessionId: string): T | undefined;
+
+	// Pending request tracking (for untitled sessions)
+	setPendingRequest(sessionId: string, request: ExtendedChatRequest, context: ChatContext): void;
+	getPendingRequest(sessionId: string): { request: ExtendedChatRequest; context: ChatContext } | undefined;
+	clearPendingRequest(sessionId: string): void;
 }
 
 export const ICopilotCLISessionService = createServiceIdentifier<ICopilotCLISessionService>('ICopilotCLISessionService');
@@ -45,6 +52,7 @@ export class CopilotCLISessionService implements ICopilotCLISessionService {
 	private _sessionManager: SessionManager | undefined;
 	private _sessionWrappers = new DisposableMap<string, IDisposable>();
 	private _sessions = new Map<string, ICopilotCLISession>();
+	private _pendingRequests = new Map<string, { request: any; context: any }>();
 
 	constructor(
 		@ILogService private readonly logService: ILogService,
@@ -222,5 +230,17 @@ export class CopilotCLISessionService implements ICopilotCLISessionService {
 
 		// Fallback to session ID
 		return `Session ${sdkSession.sessionId.slice(0, 8)}`;
+	}
+
+	public setPendingRequest(sessionId: string, request: ExtendedChatRequest, context: ChatContext): void {
+		this._pendingRequests.set(sessionId, { request, context });
+	}
+
+	public getPendingRequest(sessionId: string): { request: ExtendedChatRequest; context: ChatContext } | undefined {
+		return this._pendingRequests.get(sessionId);
+	}
+
+	public clearPendingRequest(sessionId: string): void {
+		this._pendingRequests.delete(sessionId);
 	}
 }
--- a/src/extension/chatSessions/vscode-node/chatSessions.ts
+++ b/src/extension/chatSessions/vscode-node/chatSessions.ts
@@ -58,8 +58,8 @@ export class ChatSessionsContrib extends Disposable implements IExtensionContrib
 		const copilotcliSessionItemProvider = this._register(copilotcliAgentInstaService.createInstance(CopilotCLIChatSessionItemProvider));
 		this._register(vscode.chat.registerChatSessionItemProvider(this.copilotcliSessionType, copilotcliSessionItemProvider));
 		const copilotcliAgentManager = this._register(copilotcliAgentInstaService.createInstance(CopilotCLIAgentManager));
-		const copilotcliChatSessionContentProvider = copilotcliAgentInstaService.createInstance(CopilotCLIChatSessionContentProvider);
-		const copilotcliChatSessionParticipant = new CopilotCLIChatSessionParticipant(this.copilotcliSessionType, copilotcliAgentManager, copilotcliSessionItemProvider);
+		const copilotcliChatSessionContentProvider = copilotcliAgentInstaService.createInstance(CopilotCLIChatSessionContentProvider, copilotcliAgentManager);
+		const copilotcliChatSessionParticipant = new CopilotCLIChatSessionParticipant(this.copilotcliSessionType, copilotcliAgentManager, copilotCLISessionService, copilotcliSessionItemProvider);
 		const copilotcliParticipant = vscode.chat.createChatParticipant(this.copilotcliSessionType, copilotcliChatSessionParticipant.createHandler());
 		this._register(vscode.chat.registerChatSessionContentProvider(this.copilotcliSessionType, copilotcliChatSessionContentProvider, copilotcliParticipant));
 		this._register(registerCLIChatCommands(copilotcliSessionItemProvider, copilotCLISessionService));
--- a/src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
+++ b/src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
@@ -17,8 +17,8 @@ import * as path from '../../../util/vs/base/common/path';
 import { URI } from '../../../util/vs/base/common/uri';
 import { localize } from '../../../util/vs/nls';
 import { CopilotCLIAgentManager } from '../../agents/copilotcli/node/copilotcliAgentManager';
-import { ICopilotCLISessionService } from '../../agents/copilotcli/node/copilotcliSessionService';
-import { buildChatHistoryFromEvents, parseChatMessagesToEvents } from '../../agents/copilotcli/node/copilotcliToolInvocationFormatter';
+import { ExtendedChatRequest, ICopilotCLISessionService } from '../../agents/copilotcli/node/copilotcliSessionService';
+import { buildChatHistoryFromEvents, parseChatMessagesToEvents, stripSystemReminders } from '../../agents/copilotcli/node/copilotcliToolInvocationFormatter';
 
 export class CopilotCLIChatSessionItemProvider extends Disposable implements vscode.ChatSessionItemProvider {
 	private readonly _onDidChangeChatSessionItems = this._register(new Emitter<void>());
@@ -169,6 +169,7 @@ export class CopilotCLIChatSessionItemProvider extends Disposable implements vsc
 export class CopilotCLIChatSessionContentProvider implements vscode.ChatSessionContentProvider {
 
 	constructor(
+		private readonly copilotcliAgentManager: CopilotCLIAgentManager,
 		@ICopilotCLISessionService private readonly sessionService: ICopilotCLISessionService,
 	) { }
 
@@ -180,9 +181,39 @@ export class CopilotCLIChatSessionContentProvider implements vscode.ChatSessionC
 
 		const history = existingSession ? buildChatHistoryFromEvents(events) : [];
 
+		// Check if there's a pending request for this new session
+		const pendingRequest = this.sessionService.getPendingRequest(copilotcliSessionId);
+
+		const activeResponseCallback = pendingRequest
+			? async (stream: vscode.ChatResponseStream, token: vscode.CancellationToken) => {
+				this.sessionService.clearPendingRequest(copilotcliSessionId);
+				await this.copilotcliAgentManager.handleRequest(
+					copilotcliSessionId,
+					pendingRequest.request,
+					pendingRequest.context,
+					stream,
+					token
+				);
+			}
+			: undefined;
+
+		// If there's a pending request, add it to the history as the first request turn
+		if (pendingRequest) {
+			const request = pendingRequest.request;
+			const requestTurn = new vscode.ChatRequestTurn2(
+				stripSystemReminders(request.prompt),
+				undefined,
+				[...request.references],
+				'',
+				[],
+				undefined
+			);
+			history.push(requestTurn);
+		}
+
 		return {
 			history,
-			activeResponseCallback: undefined,
+			activeResponseCallback,
 			requestHandler: undefined,
 		};
 	}
@@ -192,6 +223,7 @@ export class CopilotCLIChatSessionParticipant {
 	constructor(
 		private readonly sessionType: string,
 		private readonly copilotcliAgentManager: CopilotCLIAgentManager,
+		private readonly sessionService: ICopilotCLISessionService,
 		private readonly sessionItemProvider: CopilotCLIChatSessionItemProvider,
 	) { }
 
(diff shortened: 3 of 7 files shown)


src/extension/prompt/node/test/repoInfoTelemetry.spec.ts
```diff
@@ -0,0 +1,994 @@
...
+	ADDED_BY_US: 12,
+	ADDED_BY_THEM: 13,
+	DELETED_BY_US: 14,
+	DELETED_BY_THEM: 15,
+	BOTH_ADDED: 16,
+	BOTH_DELETED: 17,
+	BOTH_MODIFIED: 18
+} as const;
+
+suite('RepoInfoTelemetry', () => {
```
IanMatthewHuff, line 51: As opposed to the product code (which involved much more refinement and manual work) I did let Claude loose to crank a bit on the unit tests with a set of desired tests and pointing it at the other test files for usage and style. Took quite a while for the full loop, but I thought the results were good on my review.

IanMatthewHuff:
When doing a PR like this in a new repo I know that copilot will do a review, but I usually do my own copilot chat review before finalizing the PR. Result in the comment below if interested.

Prompt: 

> Could you compare the diff between my current code and upstream/main which is 742c75c. I made a number of commits to add a new way to capture git repo telemetry. This telemetry is to be used to recreate a repo at the start of agent tool calling and at the end. Can you look over my change (don't make any changes) and make a report to me. This report should detail what my changes have done. Detail specific choices / decisions that were made. Also look to see if my added code lines up with the guide in Contributing.md. If you have any particular concerns with this change you may also report those. You can start with 'git diff upstream/main --stat' and 'git diff upstream/main' to get started with the diff.

src/extension/prompt/node/chatParticipantTelemetry.ts
```diff
@@ -211,9 +214,14 @@ export class ChatTelemetryBuilder {
 		private readonly _firstTurn: boolean,
 		private readonly _request: vscode.ChatRequest,
 		@IInstantiationService private readonly instantiationService: IInstantiationService,
-	) { }
+	) {
+		this._repoInfoTelemetry = this.instantiationService.createInstance(RepoInfoTelemetry, this.baseUserTelemetry.properties.messageId);
+	}
 
 	public makeRequest(intent: IIntent, location: ChatLocation, conversation: Conversation, messages: Raw.ChatMessage[], promptTokenLength: number, references: readonly PromptReference[], endpoint: IChatEndpoint, telemetryData: readonly TelemetryData[], availableToolCount: number): InlineChatTelemetry | PanelChatTelemetry {
+		this._repoInfoTelemetry.sendBeginTelemetryIfNeeded().catch(() => {
```
IanMatthewHuff: This calling location here is one of the larger changes from the initial design that we discussed. My initial research and the jumpstart markdown doc called out the location of 'interactiveSessionMessage'

https://github.com/microsoft/vscode-copilot-chat/blob/6b5f86b82de944b3d57658bde820ad4a31e61dd0/src/extension/prompt/node/chatParticipantTelemetry.ts#L540

As the place to add the beginning of user request check. However this location is actually in the PanelChatTelemetry itself and that PanelChatTelemetry is blown away and recreated at each step of the request loop. So instead I pulled out the functionality that I had into a separate class and call it here from the builder. This catches the begin of the actual request loop.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,194 @@
...
+		const repository = gitAPI?.getRepository(repoContext.rootUri);
+		const upstreamCommit = repository?.state.HEAD?.upstream?.commit;
+		if (!upstreamCommit) {
+			return;
+		}
+
+		// Before we calculate our async diffs, sign up for file system change events
+		// Any changes during the async operations will invalidate our diff data and we send it
+		// as a failure without a diffs
+		const watcher = this._fileSystemService.createFileSystemWatcher('**/*');
```
IanMatthewHuff, line 172: Per teams discussion, if files are changed during the snapshot process just reject the event. We'll log that in telemetry so if this is happening more often than we expect we can revisit alternative ways to handle this. However this current approach seems the best way to start, as other options might have to await on some amount of data fetching. The core issue of wanting to not block, but capturing against a live git repository state feels like a trade off one way or another and the tradeoff here to be non-blocking and just reject some % of events seems right.

src/extension/prompt/node/chatParticipantTelemetry.ts
```diff
@@ -492,7 +506,6 @@ export abstract class ChatTelemetry<C extends IDocumentContext | undefined = IDo
 	protected _getTelemetryData<T extends TelemetryData>(ctor: new (...args: any[]) => T): T | undefined {
 		return <T>this._genericTelemetryData.find(d => d instanceof ctor);
 	}
-
 }
```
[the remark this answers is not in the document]
IanMatthewHuff: Fixed.

src/extension/prompt/node/chatParticipantTelemetry.ts
```diff
@@ -530,13 +544,12 @@ export class PanelChatTelemetry extends ChatTelemetry<IDocumentContext | undefin
 			promptTokenLength,
 			genericTelemetryData,
 			availableToolCount,
+			repoInfoTelemetry,
 			telemetryService
 		);
 	}
```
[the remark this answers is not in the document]
IanMatthewHuff: Fixed

src/extension/prompt/node/chatParticipantTelemetry.ts
IanMatthewHuff: Outdated, I've tweaked this.

src/extension/prompt/node/chatParticipantTelemetry.ts
```diff
@@ -211,7 +214,11 @@ export class ChatTelemetryBuilder {
 		private readonly _firstTurn: boolean,
 		private readonly _request: vscode.ChatRequest,
 		@IInstantiationService private readonly instantiationService: IInstantiationService,
-	) { }
+	) {
+		// Repo info telemetry is held here as the begin event should be sent only by the first PanelChatTelemetry instance created for a user request.
+		// and a new PanelChatTelemetry instance is created per step in the request.
+		this._repoInfoTelemetry = this.instantiationService.createInstance(RepoInfoTelemetry, this.baseUserTelemetry.properties.messageId);
```
IanMatthewHuff, line 220: This bit is a bit different from the original implementation and tripped me up a bit on my first pass where I was just adding the repo telemetry functions as a part of PanelChatTelemetry. The PanelChatTelemetry is recreated and the old one is chucked for each step in the process not per user level chat request. So if you ask the agent to do something the first PanelChatTelemetry would be trying to log the begin user repo info and then it would get thrown out for the next step PanelChatTelemetry.

So instead the class that manages this telemetry is created at the builder level which is created per user request in the chat. The PanelChatTelemetry calls the logging still at the appropriate spots (beginning and the end of the tool calls). But since the build holds the repoInfoTelemetry instance it doesn't get thrown out when the first PanelChatTelemetry is thrown out for the next one and keeps processing the begin event. It's guarded to only call the begin event one time so the subsequent PanelChatTelemetries don't re-create the begin event.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,195 @@
...
+					headCommitHash: upstreamCommit,
+					diffsJSON: undefined,
+					result: 'filesChanged',
+				};
+			}
+
+			const diffsJSON = diffs.length > 0 ? JSON.stringify(diffs) : undefined;
+
+			// Check if the diff is too big and notify that
+			if (wouldMultiplexTelemetryPropertyBeTruncated(diffsJSON)) {
```
sbatten: after discussion, the multiplexing thing doesn't apply to internal msft endpoint. so we don't need this check, its 1MB for the whole payload and a single property can be up to limit (reserve some space for the json structure and other event properties)

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,195 @@
...
+		});
+	}
+
+	/*
+	* Sends the end event telemetry
+	*/
+	public async sendEndTelemetry(): Promise<void> {
+		await this._beginTelemetryPromise;
+
+		return this._sendRepoInfoTelemetry('end').catch((error) => {
```
sbatten: I think this should use await and a proper try catch

src/extension/prompt/node/chatParticipantTelemetry.ts
```diff
@@ -552,6 +567,13 @@ export class PanelChatTelemetry extends ChatTelemetry<IDocumentContext | undefin
 		} satisfies RequestInternalPanelTelemetryProperties, {
 			turnNumber: this._conversation.turns.length,
 		} satisfies ResponseInternalPanelTelemetryMeasurements);
+
+		// Send the begin telemetry for repo info, this uses the same repo info telemetry instance held by the builder class
+		// as the begin event need to be sent only once per user request and PanelChatTelemetry is recreated per step. The class is
+		// guarded to only send one time.
+		this._repoInfoTelemetry.sendBeginTelemetryIfNeeded().catch(() => {
```
sbatten: Since the exception is caught internally, we don't need to catch here right?

src/extension/prompt/node/repoInfoTelemetry.ts
IanMatthewHuff: Oh interesting, I did see it used on a couple of calls to internalMSFT telemetry logging, but I guess it wouldn't have hurt anything in those cases. I'll remove and adjust to just respect that full payload limit.

src/extension/prompt/node/chatParticipantTelemetry.ts
IanMatthewHuff: Sorry that was a holdover from working in my usual code base, we have it flag floating promises. So I'm used to adding a catch / comment for a fire and forget operation like that. I'll update, as I see more examples of just calling them in this extension and the repoInfoTelemetry is logging the exceptions already (I'll update the functions to try catch instead of the promise returning per the other comment).

src/extension/prompt/node/repoInfoTelemetry.ts
IanMatthewHuff: I updated. Initially I was thinking of checking the entire payload size at this point, but given that we add in more shared properties later your suggestion of just going a bit more conservative (I used 90%) with the bulky diff string after it was stringified made made sense to me, so I went with that. 10% felt like more than enough room to cover the extra properties + structure.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,200 @@
...
+			if (filesChanged) {
+				return {
+					remoteUrl: githubInfo.remoteUrl,
+					headCommitHash: upstreamCommit,
+					diffsJSON: undefined,
+					result: 'filesChanged',
+				};
+			}
+
+			const diffs = (await this._gitDiffService.getChangeDiffs(repoContext.rootUri, changes)).map(diff => {
```
IanMatthewHuff, line 220: @sbatten One comment here that I was considering this morning. Given that the goal as far as I can tell is to collect examples to use, but not that we have to collect "every" example to use. I could put in some type of # of changes sanity check here to prevent a degenerate case with something like a mass formatting refactor. It looks to me like getChangeDiffs is handling things in a reasonable way (in order, so it wouldn't flood with 1000 of git calls at one time) but some type of 'skip this if our changes count is too massive' might help avoid a bad scenario.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,200 @@
...
+		const watcher = this._fileSystemService.createFileSystemWatcher('**/*');
+		let filesChanged = false;
+		const createDisposable = watcher.onDidCreate(() => filesChanged = true);
+		const changeDisposable = watcher.onDidChange(() => filesChanged = true);
+		const deleteDisposable = watcher.onDidDelete(() => filesChanged = true);
+
+		try {
+			const changes = await this._gitService.diffWith(repoContext.rootUri, '@{upstream}');
+			if (!changes || changes.length === 0) {
+				return;
```
IanMatthewHuff: Making a small tweak this morning post testing more. Instead of not sending an event if there is no change from upstream send a 'noChange' event with empty diff. Should make the 'begin' 'end' matching more clear instead of just having an 'end' event come out of nowhere with diffs if you start in a no changes state.

src/extension/prompt/node/repoInfoTelemetry.ts
sbatten: I think that makes sense, just like the rest, we can flag it to understand how often it happens and if we should ignore the rest of the session.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,205 @@
...
+
+		if (!repoContext) {
+			return;
+		}
+
+		const githubInfo = getGitHubRepoInfoFromContext(repoContext);
+		if (!githubInfo) {
+			return;
+		}
+
```
sbatten, line 147: support ADO?

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,205 @@
...
+			// Check if files changed before we send the telemetry
+			if (filesChanged) {
+				return {
+					remoteUrl: githubInfo.remoteUrl,
+					headCommitHash: upstreamCommit,
+					diffsJSON: undefined,
+					result: 'filesChanged',
+				};
+			}
```
sbatten: this one doesn't seem necessary

src/extension/prompt/node/repoInfoTelemetry.ts
IanMatthewHuff: Updated, I was a bit unsure on the best value to use here, I made it 100 as my guess was we're more just looking to avoid some extra bad cases, and it's logging a specific telemetry result (different from the diff string being too large) so we can track if it's hitting.

IanMatthewHuff: Also per your comment:

> ignore for the rest of the session

I realized that the code currently wasn't doing this. So I changed the end telemetry sending to not bother trying to calculate or send if the 'begin' event was one of the failure cases, I guess we could technically be in a 'diffTooBig' state at 'begin' and then not be after the tool runs, but we'd only have the 'end' case for that scenario, so I don't think it's useful.

src/extension/prompt/node/repoInfoTelemetry.ts
IanMatthewHuff: Agreed and updated, we already have the diff at this point.

src/extension/prompt/node/repoInfoTelemetry.ts
IanMatthewHuff: Shoot, this was a miss on my part. I had git on the brain and wasn't considering the ADO side of the request. I'll look into this now. If it's a lot to add I can ask the science folks if we can start with collecting info just from git and move forward with ADO later, since that would unblock the pipeline getting started.

IanMatthewHuff: Ohh, it looks like ADO support might be basically there already in the GitService. Checking now, I'll have to create an ADO repo quick myself to test, I've honestly not used one in years.

IanMatthewHuff: Added ado support, normalized the fetch Url before logging, and cleaned up the return types here by pulling out the base properties.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,252 @@
...
+
+type RepoInfoTelemetryProperties = {
+	remoteUrl: string | undefined;
+	repoType: 'github' | 'ado';
+	headCommitHash: string | undefined;
+	diffsJSON: string | undefined;
+	result: RepoInfoTelemetryResult;
+};
+
+type RepoInfoTelemetryMeasurements = {
```
IanMatthewHuff, line 33: Measurements now for workspaceFileCount, changedFileCount, and the size of the diffJSON.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,252 @@
...
+			const baseProperties: Omit<RepoInfoTelemetryProperties, 'diffsJSON' | 'result'> = {
+				remoteUrl: normalizedFetchUrl,
+				repoType: repoInfo.repoId.type,
+				headCommitHash: upstreamCommit,
+			};
+
+			// Workspace file index will be used to get a rough count of files in the repository
+			// We need to call initialize here to have the count, but after first initialize call
+			// further calls are no-ops so only a hit first time.
+			await this._workspaceFileIndex.initialize();
```
IanMatthewHuff, line 188: I was checking with the debugger and I do see the initialize hit for the first time here in this code as opposed to one of the other initialize locations for workspace file index. Looked at the other code paths and it seems like this might or might not be initialized at this point, so we consider that it could quite likely take the initial initialize hit here but would be be the first time.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -0,0 +1,252 @@
...
+					measurements
+				};
+			}
+
+			const diffsJSON = diffs.length > 0 ? JSON.stringify(diffs) : undefined;
+
+			// Check against our size limit to make sure our telemetry fits in the 1MB limit
+			if (diffsJSON) {
+				const diffSizeBytes = Buffer.byteLength(diffsJSON, 'utf8');
+				measurements.diffSizeBytes = diffSizeBytes;
```
IanMatthewHuff, line 243: Tracking the size of the diffJSON to get an better idea on the extent of the diffs for the science team without having to try to crack into the individual diff strings since they were interested in knowing the complexity of the diff. Also if we happen to start hitting the MAX size in telemetry we'll actually know by how much with this added.

src/platform/workspaceChunkSearch/node/nullWorkspaceFileIndex.ts
```diff
@@ -0,0 +1,56 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { GlobIncludeOptions } from '../../../util/common/glob';
+import { CancellationToken } from '../../../util/vs/base/common/cancellation';
+import { Event } from '../../../util/vs/base/common/event';
+import { URI } from '../../../util/vs/base/common/uri';
+import { IWorkspaceFileIndex } from './workspaceFileIndex';
+
+export class NullWorkspaceFileIndex implements IWorkspaceFileIndex {
```
IanMatthewHuff, line 12: Added a null service for this one as well as it helped a bunch with the unit tests.

src/extension/prompt/node/test/repoInfoTelemetry.spec.ts
```diff
@@ -0,0 +1,1543 @@
...
+		vi.spyOn(gitService, 'diffWith').mockImplementation(async () => {
+			// Simulate file creation during diff
+			mockWatcher.triggerCreate(URI.file('/test/repo/newfile.ts') as any);
+			return [{
+				uri: URI.file('/test/repo/file.ts'),
+				originalUri: URI.file('/test/repo/file.ts'),
+				renameUri: undefined,
+				status: Status.MODIFIED
+			}] as any;
+		});
```
sbatten, line 618: is this actually the create scenario? its missing the newFile.ts?

IanMatthewHuff: The way that we were using the file watcher in repoInfoTelemetry was just that if we detected any type of file change in the repository while we were collecting the diff data we'd just return 'filesChanged' and invalidate the event. For this test the triggerCreate was just simulating any type of file creation during the call to diffWith. The filename for that step could be anything, it doesn't have to line up with the URIs in the mock return value of diffWith. And the Status.MODIFIED is just for the mock return value. Test just needs to call triggerCreate on something, and then return a valid mock Change from the mock diffWith. Then verify we get 'filesChanged' sent to telemetry.

So I think the test works as is, unless I'm misunderstanding the comment?

I guess in the file deletion a couple down I do use a .DELETED status, which makes it seem like it's relevant to the test, when it's not. I can change that to modified as well quick.

IanMatthewHuff: Added a clarifying comment on those file change tests.

PR #1235, opened 2025-10-06
Fix: Exp enable completions and NES fetcher

chrmarti:
https://github.com/microsoft/vscode/issues/270004

### DIFF
--- a/src/platform/configuration/common/configurationService.ts
+++ b/src/platform/configuration/common/configurationService.ts
@@ -20,6 +20,7 @@ import { ResponseProcessor } from '../../inlineEdits/common/responseProcessor';
 import { AlternativeNotebookFormat } from '../../notebook/common/alternativeContentFormat';
 import { IExperimentationService } from '../../telemetry/common/nullExperimentationService';
 import { IValidator, vBoolean, vString } from './validator';
+import { FetcherId } from '../../networking/common/fetcherService';
 
 export const CopilotConfigPrefix = 'github.copilot';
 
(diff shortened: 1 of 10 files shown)


src/platform/configuration/common/configurationService.ts
```diff
@@ -796,6 +797,9 @@ export namespace ConfigKey {
 	export const EnableAlternateGptPrompt = defineExpSetting<boolean>('chat.alternateGptPrompt.enabled', false);
 	export const Gpt5AlternatePrompt = defineExpSetting<string>('chat.gpt5AlternatePrompt', 'default');
 	export const GrokCodeAlternatePrompt = defineExpSetting<string>('chat.grokCodeAlternatePrompt', 'default');
+
+	export const CompletionsFetcher = defineSetting<FetcherId | undefined>('chat.completionsFetcher', undefined);
```
dbaeumer: Should this not be a `defineExpSetting`. And the one below

src/platform/networking/vscode-node/fetcherServiceImpl.ts
```diff
@@ -106,6 +115,7 @@ export class FetcherService implements IFetcherService {
 						}
 						this._logService.info(`FetcherService: using ${fetcher.getUserAgentLibrary()} from now on`);
 						this._fetcher = fetcher;
+						this._knownBadFetchers = knownBadFetchers;
```
dbaeumer, line 118: Here I am unsure: if the bad fetchers are per fetch request then this assignment statement is correct. But the I would argue that we don't need `this._knownBadFetchers`. If it is a state in the fetcher then I thing we should add all `knownBadFetchers` into `this._knownBadFetchers`

chrmarti: `options.verifyJSONAndRetry` above is currently only set on the copilot token (and user info) request. If any fetchers fail we remember it here. This is incomplete because we stop at the first fetcher that succeeds, we can improve on this in the future.

PR #1237, opened 2025-10-06
feat: Add "Memory Tool" for embedding and storing workspace documents

Nishtha9:
This PR introduces a Memory Tool that embeds, indexes, and stores workspace documents to enable semantic recall.
Tool actions provided:

- Save Memory — Embed a document or selection and persist its vector, metadata (source URI, range, timestamp), and optional tags or notes.
- List Memory — Return a paginated list of stored memory entries including metadata, source references, and short excerpt previews.
- Query Memory — Run a semantic search against stored embeddings using a natural-language query; return ranked results with relevance scores and source context.
- Delete Memory — Remove stored memory entries by id, source, or filter (e.g., age, tag).
- Clear Memory — Bulk delete all stored memories for a workspace or namespace.
- Export/Import Memory — Export stored embeddings+metadata to a portable format and re-import them into another workspace or environment.

These tool actions enable semantic retrieval workflows by operating over embedded document vectors rather than keyword-only text.


Nishtha9:

bpasero:
fyi @digitarald

PR #1248, opened 2025-10-07
Update telemetry in CreateFileTool to include model and file extension details

bhavyaus:


src/extension/tools/node/createFileTool.tsx
```diff
@@ -92,6 +92,10 @@ export class CreateFileTool implements ICopilotTool<ICreateFileParams> {
 		}
 
 		const languageId = doc?.languageId ?? getLanguageForResource(uri).languageId;
+		const fileExtension = extname(uri);
+		const model = options.model ?? this._promptContext.request?.model;
+		const modelId = typeof model === 'string' ? model : model?.id;
```
connor4312: In other files we use this pattern, I'm not 100% sure the reason behind the pattern but we should stick with it for consistency

https://github.com/microsoft/vscode-copilot-chat/blob/8f7155721315874aaacfca8d1f75feedfa9aa6d6/src/extension/tools/node/newNotebookTool.tsx#L138

connor4312 (approved):

PR #1262, opened 2025-10-08
Add BYOK API key management command

zhichli:
dep of microsoft/vscode-copilot-evaluation#428

This programmatic way of managing API key via env var is a more stable way of configuring custom model in our test environment.


sandy081:
@zhichli / @lramos15 / @pwang347 

May I know the purpose of this API command `github.copilot.chat.manageBYOKAPIKey`? 
I am in the process of refactoring how BYOK models can be configured and therefore want to understand the use case behind this to see how I can fit this in the new world.

lramos15:
> @zhichli / @lramos15 / @pwang347
> 
> May I know the purpose of this API command `github.copilot.chat.manageBYOKAPIKey`? I am in the process of refactoring how BYOK models can be configured and therefore want to understand the use case behind this to see how I can fit this in the new world.

Our evaluation framework doesn't easily react with the UI of VS Code so it often runs the evaluations via commands.

So in the case of wanting to evaluate against a BYOK model we need a way to set it up completely without UI interaction and that includes setting the API key that should back it. The flow is normally
1. Write to the setting the custom endpoint config
2. use this command to set the api key
3. Use a comamnd to select the model in chat
4. Send messages to chat

PR #1266, opened 2025-10-08
Don't block extension activation on auth

roblourens:
This is hopefully the most minimal and least risky change to test this. If we keep it, then there is more cleanup that could be done in conversationFeature.ts and chatParticipants.ts.

I tested that the welcome views for things like an expired subscription still show correctly.

Now, we go to the built-in ChatSetup participant to drive extension install if needed, and for the now shorter time that the extension is activating. Then the request is forwarded to the extension, and auth happens inside any registered providers. 

Decouples the extension activation process from GitHub authentication, allowing the extension to activate without blocking on user sign-in. Removes unnecessary activation blockers and cleans up related code. Adds logging for better traceability during activation.

Fixes microsoft/vscode-copilot#3879

### DIFF
--- a/src/extension/conversation/vscode-node/chatParticipants.ts
+++ b/src/extension/conversation/vscode-node/chatParticipants.ts
@@ -10,7 +10,9 @@ import { IInteractionService } from '../../../platform/chat/common/interactionSe
 import { ConfigKey, IConfigurationService } from '../../../platform/configuration/common/configurationService';
 import { IEndpointProvider } from '../../../platform/endpoint/common/endpointProvider';
 import { IOctoKitService } from '../../../platform/github/common/githubService';
+import { ILogService } from '../../../platform/log/common/logService';
 import { IExperimentationService } from '../../../platform/telemetry/common/nullExperimentationService';
+import { DeferredPromise } from '../../../util/vs/base/common/async';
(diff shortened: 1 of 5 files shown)


src/extension/conversation/vscode-node/chatParticipants.ts
```diff
@@ -67,7 +71,22 @@ class ChatAgents implements IDisposable {
...
+		const requestBlockerDeferred = new DeferredPromise<void>();
+		this.requestBlocker = requestBlockerDeferred.p;
+		if (authenticationService.copilotToken) {
+			this.logService.debug(`ChatParticipants: Copilot token already available`);
+			requestBlockerDeferred.complete();
+		} else {
+			this._disposables.add(authenticationService.onDidAuthenticationChange(async () => {
+				const hasSession = !!authenticationService.copilotToken;
+				this.logService.debug(`ChatParticipants: onDidAuthenticationChange has token: ${hasSession}`);
+				requestBlockerDeferred.complete();
```
TylerLeonhardt, line 86: should this have been wrapped in a check of `hasSession`?

roblourens: Uhhh maybe

PR #1269, opened 2025-10-08
feat: support folder paths in get_errors tool

bryanchen-d:
Support folder paths in get_errors tool.

This is to fix https://github.com/microsoft/vscode/issues/265437

### DIFF
--- a/src/platform/test/node/simulationWorkspace.ts
+++ b/src/platform/test/node/simulationWorkspace.ts
@@ -360,6 +360,10 @@ export class SimulationWorkspace extends Disposable {
 		return this._diagnostics.get(uri) ?? [];
 	}
 
+	public getAllDiagnostics(): [vscode.Uri, vscode.Diagnostic[]][] {
+		return Array.from(this._diagnostics.entries());
+	}
+
 	public getDocument(filePathOrUri: string | vscode.Uri): IExtHostDocumentData {
 		const queryUri = typeof filePathOrUri === 'string' ? this.getUriFromFilePath(filePathOrUri) : filePathOrUri;
 		const candidateFile = this._docs.get(queryUri);
--- a/src/platform/test/node/simulationWorkspaceServices.ts
+++ b/src/platform/test/node/simulationWorkspaceServices.ts
@@ -135,7 +135,7 @@ export class SimulationLanguageDiagnosticsService extends AbstractLanguageDiagno
 	override onDidChangeDiagnostics: vscode.Event<vscode.DiagnosticChangeEvent> = this.workspace.onDidChangeDiagnostics;
 	override getDiagnostics: (resource: vscode.Uri) => vscode.Diagnostic[] = this.workspace.getDiagnostics.bind(this.workspace);
 	override getAllDiagnostics(): [vscode.Uri, vscode.Diagnostic[]][] {
-		return [];
+		return this.workspace.getAllDiagnostics();
 	}
 }
 


src/extension/tools/node/getErrorsTool.tsx
```diff
@@ -52,33 +52,81 @@ class GetErrorsTool extends Disposable implements ICopilotTool<IGetErrorsParams>
...
+			}
+
+			// find all path&range pairs and collect the ranges to further filter diagnostics
+			// if any path matches the resource without a range, take all diagnostics for that file
+			// otherwise, filter diagnostics to those intersecting one of the provided ranges
+			const ranges: Range[] = [];
+			let shouldTakeAll = false;
+			for (const path of nonNotebookPaths) {
+				// we support file or folder paths
+				if (isEqualOrParent(resource, path.uri)) {
```
roblourens, line 91: A TernarySearchTree might be an efficient way to process all the paths?

src/extension/tools/node/getErrorsTool.tsx
```diff
@@ -52,33 +52,81 @@ class GetErrorsTool extends Disposable implements ICopilotTool<IGetErrorsParams>
 		super();
 	}
 
+	getDiagnostics(paths: { uri: URI; range: Range | undefined }[]) {
```
roblourens: Can be private?

src/extension/tools/node/getErrorsTool.tsx
```diff
@@ -52,33 +52,81 @@ class GetErrorsTool extends Disposable implements ICopilotTool<IGetErrorsParams>
...
+
+			return !isNotebook;
+		});
+
+		if (nonNotebookPaths.length === 0) {
+			return results;
+		}
+
+		// for non-notebooks, we get all diagnostics and filter down
+		for (const [resource, entries] of this.languageDiagnosticsService.getAllDiagnostics()) {
```
roblourens, line 80: Let's add a test for the new code too. There are some test cases in getErrorsTool.spec.tsx but it only covers the tool output so we can add new cases to cover the tool itself too

src/extension/tools/node/getErrorsTool.tsx
bryanchen-d: Per our talk offline, this isEqualOrParent is good enough.

src/extension/tools/node/getErrorsTool.tsx
bryanchen-d: added

src/extension/tools/node/getErrorsTool.tsx
bryanchen-d: Yes. Marked private

src/extension/tools/node/getErrorsTool.tsx
```diff
@@ -52,33 +52,81 @@ class GetErrorsTool extends Disposable implements ICopilotTool<IGetErrorsParams>
...
+			}
+
+			// find all path&range pairs and collect the ranges to further filter diagnostics
+			// if any path matches the resource without a range, take all diagnostics for that file
+			// otherwise, filter diagnostics to those intersecting one of the provided ranges
+			const ranges: Range[] = [];
+			let shouldTakeAll = false;
+			for (const path of nonNotebookPaths) {
+				// we support file or folder paths
+				if (isEqualOrParent(resource, path.uri)) {
```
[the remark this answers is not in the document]
bryanchen-d: Nope, it is to check if `path.uri` is equal to or a parent of the `resource`.

src/extension/tools/node/test/getErrorsTool.spec.tsx
```diff
@@ -6,89 +6,183 @@
...
+		// Set up diagnostics service
 		diagnosticsService = new TestLanguageDiagnosticsService();
 		collection.define(ILanguageDiagnosticsService, diagnosticsService);
+
 		accessor = collection.createTestingAccessor();
+
+		// Create the tool instance
+		tool = accessor.get(IInstantiationService).createInstance(GetErrorsTool);
+		// Access the private method through reflection for testing
+		getDiagnosticsFunc = (tool as any).getDiagnostics.bind(tool);
```
roblourens: Please don't do this- see some of our recent conversations on Slack about trying to get rid of 'as any' casts. If you want to expose that method to tests, just make it public with a comment that it should only be used for tests. Or make the tests full e2e testing the public interface.

src/extension/tools/node/test/getErrorsTool.spec.tsx
```diff
@@ -6,89 +6,183 @@
...
+		const results = getDiagnosticsFunc([{ uri: tsFile1, range }]);
+
+		expect(results).toEqual([
+			{ uri: tsFile1, diagnostics: diagnosticsService.getDiagnostics(tsFile1).filter(d => d.severity <= DiagnosticSeverity.Warning && d.range.intersection(range)) }
+		]);
+	});
+
+	// Tool invocation tests
+	test('Tool invocation - with no filePaths aggregates all diagnostics and formats workspace message', async () => {
+		const result = await tool.invoke({ input: {} as any, toolInvocationToken: null! }, CancellationToken.None);
```
roblourens: Please try to avoid skipping TS, we are trying to get rid of this pattern and make sure that we only leave positive examples for Copilot to pick up on

src/extension/tools/node/test/getErrorsTool.spec.tsx
```diff
@@ -6,89 +6,183 @@
...
+		const msg = (result.toolResultMessage as any).value ?? result.toolResultMessage;
+		// 4 total (file1: 2, file2: 1, jsFile: 1) excluding the Information diagnostic
+		expect(msg).toContain('Checked workspace');
```
roblourens: It's useful to use snapshots for cases like this, if you want

roblourens (commented):
Sorry forgot to submit this review from earlier

src/extension/tools/node/test/getErrorsTool.spec.tsx
bryanchen-d: updated using snapshots

src/extension/tools/node/test/getErrorsTool.spec.tsx
bryanchen-d: Updated removing the cast to any.

src/extension/tools/node/test/getErrorsTool.spec.tsx
bryanchen-d: Got it, that makes sense. Updated the method to be public and added the comment.

PR #1286, opened 2025-10-10
rm `startDebugging` intent in favor of `startDebugging` tool

meganrogge:
fixes https://github.com/microsoft/vscode/issues/258698

@roblourens I wanted to use a confirmation here vs buttons, but it doesn't allow me to set the title of the buttons 🫠 . Is there an issue tracking that ?

Screenshot 2025-10-10 at 12 40 32 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -3169,10 +3169,7 @@
 						]
 					},
 					"github.copilot.chat.completionsFetcher": {
-						"type": [
-							"string",
-							"null"
-						],
+						"type": ["string", "null"],
 						"markdownDescription": "%github.copilot.config.completionsFetcher%",
 						"tags": [
 							"experimental",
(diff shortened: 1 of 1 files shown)


connor4312:
Is it worthwhile for the model to have a `startDebugging` tool? For the original issue of creating a launch config, it can do that with normal file editing tools, but currently the model cannot interact with debugging at all

meganrogge:
@connor4312 , I'm not sure. The `start debugging` button is kind of nice, especially for beginners, who might not find the debug panel easily. I tested with and without it and they did produce similiar results.

roblourens:
I don't really see the value of this, sorry. It seems less convenient than the agent just editing files itself. Currently we have many hyper-specific tools which are never used and we are trying to reduce these too.

Also in general, tools shouldn't be streaming markdown, that should only come from the agent. A tool can't really show a confirmation, it unfortunately doesn't really fit our API. I know that we use this pattern to plumb the chat response stream into the tool in other places, but it should really only be used in special scenarios.

meganrogge:
Ok, i'll remove the intent and not add a tool, sounds good.

connor4312:
Something we could do is show a "Start Debugging" button/affordance if we see the model changes a launch.json file. I think that would get you to a pretty close place to where you're going here, but without the addition of another tool and more predictably (it just happens on edit without an extra model step needed)

PR #1295, opened 2025-10-12
File edit tool prompt improvements - eliminate contradictions

upman:
## Problem

The prompt instructions contradict the tool examples for `insert_edit_into_file`:

- **Instructions say**: Only use `insert_edit_into_file` if `replace_string_in_file` fails
- **Examples show**: Direct usage of `insert_edit_into_file` without showing the fallback pattern

This mixed messaging causes models to use `insert_edit_into_file` as a primary tool, which is problematic because it's significantly slower on large files than `replace_string_in_file`.

## Changes

Updated tool descriptions and examples to demonstrate the correct fallback workflow:
1. First attempt: `replace_string_in_file` (faster)
2. On failure: Fall back to `insert_edit_into_file`

**Files modified:**
- `package.json`: Updated `modelDescription` and parameter descriptions for `copilot_insertEdit`
- `agentInstructions.tsx`: Updated 3 example workflows to show the fallback pattern
- `agentPrompt.spec.tsx.snap`: Updated snapshots

### DIFF
--- a/package.json
+++ b/package.json
@@ -4132,16 +4132,7 @@
 							"program"
 						]
 					}
-				},
-				"initialConfigurations": [
-					{
-						"type": "vscode-chat-replay",
-						"request": "launch",
-						"name": "Debug Chat Replay",
-						"program": "${file}",
-						"stopOnEntry": true
-					}
-				]
+				}
 			}
 		]
 	},
--- a/src/platform/endpoint/common/chatModelCapabilities.ts
+++ b/src/platform/endpoint/common/chatModelCapabilities.ts
@@ -33,7 +33,7 @@ export async function isHiddenModelA(model: LanguageModelChat | IChatEndpoint) {
 }
 
 export async function isHiddenModelB(model: LanguageModelChat | IChatEndpoint) {
-	return await getSha256Hash(model.family) === '42029ef215256f8fa9fedb53542ee6553eef76027b116f8fac5346211b1e473c';
+	return await getSha256Hash(model.family) === '4243b479ae1f345e3b3beff413f628c1c3edf2b54bc75859e736f0254231aafb';
 }
 
 /**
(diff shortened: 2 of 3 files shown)


upman:

upman:
I added a simulation test in a separate PR: https://github.com/microsoft/vscode-copilot-chat/pull/1296

package.json
```diff
@@ -652,7 +652,7 @@
...
+				"modelDescription": "Insert new code into an existing file in the workspace. Use this tool ONLY if replace_string_in_file has failed. Generate the \"explanation\" property first.\n Use this tool only once per file that needs to be modified, even if there are multiple changes for a file. Combine all the changes together. The system is very smart and can understand how to apply your edits to the files, you just need to provide minimal hints.\nAvoid repeating existing code, instead use comments to represent regions of unchanged code. Be as concise as possible. For example:\n// ...existing code...\n{ changed code }\n// ...existing code...\n{ changed code }\n// ...existing code...\n\nExample workflow - try replace_string_in_file first:\n1. First attempt with replace_string_in_file fails with error: \"Could not find the exact string in the file\"\n2. Then use insert_edit_into_file as fallback:\nclass Person {\n\t// ...existing code...\n\tage: number;\n\t// ...existing code...\n\tgetAge() {\n\treturn this.age;\n\t}\n}",
```
connor4312, line 655: insert_edit is not exclusively used with replace_string_in_file. For models with fewer capabilities it may be given as the only edit tool.

src/extension/prompts/node/agent/agentInstructions.tsx
```diff
@@ -170,7 +170,9 @@ export class DefaultAgentPrompt extends PromptElement<DefaultAgentPromptProps> {
 				changed code<br />
 				// {EXISTING_CODE_MARKER}<br />
 				<br />
-				Here is an example of how you should format an edit to an existing Person class:<br />
+				Example workflow - always try {tools[ToolName.ReplaceString] ? `${ToolName.ReplaceString}` : 'replace_string_in_file'} first, then fall back to {ToolName.EditFile} if it fails:<br />
```
connor4312, line 173: We state right above:

> Use the insert_edit_into_file tool to insert code into a file ONLY if multi_replace_string_in_file/replace_string_in_file has failed

This also would need to be conditional only if the replace string tool is actually available

upman: Makes sense

upman:
> The only flagship model that uses both replace_string and insert_edit is Gemini 2.5 Pro and that may change with future models. Other models have evolved and generally take replace_string exlusively nowadays.

I just looked at `agentIntent.ts#getAgentTools` and you're right. Most models don't have the insert_edit tool enabled. You're probably also right that all models will probably start supporting replace better.

I will close the PR :)

PR #1296, opened 2025-10-12
Add a simulation test file to make sure edit uses full file rewrite only as a fallback

upman:
- Add a simulation test that checks that the full file LLM based rewrite tool is only used as a fallback in case replace in file tool fails
- Add a large typescript file as a fixture

Related PR: https://github.com/microsoft/vscode-copilot-chat/pull/1295

### DIFF
--- a/test/e2e/editToolFallback.stest.ts
+++ b/test/e2e/editToolFallback.stest.ts
@@ -1,0 +1,90 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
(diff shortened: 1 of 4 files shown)


roblourens:
Thanks but why? I'm not sure what I'm looking at

upman:
@roblourens Sorry, I didn't tag the related PR explicitly: https://github.com/microsoft/vscode-copilot-chat/pull/1295

There is some contradiction in the tool descriptions and example for tool use that I addressed in the PR above.

Beause of this, some models like GPT4.1 end up using the `insert_edit_into_file` directly without attempting `replace_string_in_file` first. `insert_edit_into_file` is significantly slower and costs a lot more tokens.

upman:
Closing since most models seem to already not have the edit tool enabled and will probably evolve in this direction going further: https://github.com/microsoft/vscode-copilot-chat/pull/1295#issuecomment-3408064913

PR #1301, opened 2025-10-13
Allow exp of default model

lramos15:
Allows experimentation of default model via extension

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1433,7 +1433,6 @@
 			"integrity": "sha512-7cfaOQuCS27HD7DX+6ib2OrnW+b4ZBwDNnCcT0uTyidcmyWb03FnQqJybDBoCnpdxwBSfA94UAYlRCt7mV+TbA==",
 			"dev": true,
 			"license": "MIT",
-			"peer": true,
 			"dependencies": {
 				"@floating-ui/core": "^1.7.2",
 				"@floating-ui/utils": "^0.2.10"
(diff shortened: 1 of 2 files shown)


jul-stevenson:
@lramos15 do you know how these changes interact with these variables that also change the default model in agent mode? To my understanding, they've only been used for OSWE internal flights. It seems like the intention is the same, so we can switch OSWE to use the chat.defaultLanguageModel variable when they do a next iteration:

```
vscode.custommodel1
vscode.custommodel1.id
vscode.custommodel1.name
```

lramos15:
> @lramos15 do you know how these changes interact with these variables that also change the default model in agent mode? To my understanding, they've only been used for OSWE internal flights. It seems like the intention is the same, so we can switch OSWE to use the chat.defaultLanguageModel variable when they do a next iteration:
> 
> ```
> vscode.custommodel1
> vscode.custommodel1.id
> vscode.custommodel1.name
> ```

@karthiknadig Would know more about these, but I don't think so

karthiknadig:
@jul-stevenson `chat.defaultLanguageModel` overrides the one set by `vscode.custommodel1`.

jul-stevenson:
Thanks, @karthiknadig! @isidorn and I discussed moving all default model experiments to a single experiment which covers all users. Since `chat.defaultLanguageModel` overrides `vscode.custommodel1` it sounds like that variable won't even work going forward/it can be cleaned up. Does that sound right? Or is `vscode.custommodel1` doing something unique that we still need it for?

karthiknadig:
Only the default model part becomes useless. The other things it does is still useful, like acting as a proxy and allowing you to change visible name of the model in the picker.

PR #1311, opened 2025-10-13
edits: add Library to default restricted paths on macos

connor4312:


connor4312:

PR #1327, opened 2025-10-14
ensure fast new chat session swap

rebornix:


rebornix:
@rzhao271 it was poorly done via the local agent. Fixed them now. The code review from copilot is great.

PR #1343, opened 2025-10-15
Include inline completions in @vscode/chat-lib

chrmarti:
WIP for including inline completions in @vscode/chat-lib.

Initially we might be able to just reexport `getInlineCompletions` from the library because that expects objects of types the CLS also has. I have updated the extraction script to also follow `#lib/` and similar imports. 

Setup:
```
git clone -b completions-port https://github.com/microsoft/vscode-copilot-completions src/extension/completions-core
```

Extract and compile:
```
npm run extract-chat-lib
```

This currently shows compile errors. One issue appears to be that the completions code reaches back to `#bridges/` which then reaches into completions extension code. The extraction script follows all these code references and ends up extracting extension code we don't need in the library and that doesn't compile without the VS Code API. Maybe the direct references to `#bridges/` can be avoided.

Once it compiles, tests can be run:
```
(cd chat-lib && npm run build && npm run test)
```

### DIFF
--- a/src/lib/node/chatLibMain.ts
+++ b/src/lib/node/chatLibMain.ts
@@ -777,11 +777,7 @@ function setupCompletionServices(options: IInlineCompletionsProviderOptions): II
 		}
 	});
 	builder.define(ICompletionsContextProviderService, new ContextProviderStatistics());
-	try {
-		builder.define(ICompletionsPromptFactoryService, new SyncDescriptor(CompletionsPromptFactory));
-	} catch (e) {
-		console.log(e);
-	}
+	builder.define(ICompletionsPromptFactoryService, new SyncDescriptor(CompletionsPromptFactory));
 	builder.define(ICompletionsFetcherService, new class implements ICompletionsFetcherService {
 		declare _serviceBrand: undefined;
 		getImplementation(): ICompletionsFetcherService | Promise<ICompletionsFetcherService> {


src/lib/node/chatLibMain.ts
```diff
@@ -455,9 +507,327 @@ class SimpleTelemetryService implements ITelemetryService {
...
+			if (options.citationHandler) {
+				return await options.citationHandler.handleIPCodeCitation(citation);
+			}
+		}
+	});
+	builder.define(ICompletionsContextProviderService, new ContextProviderStatistics());
+	try {
+		builder.define(ICompletionsPromptFactoryService, new SyncDescriptor(CompletionsPromptFactory));
+	} catch (e) {
+		console.log(e);
```
chrmarti: Leftover console.log().

package.json
```diff
@@ -5265,6 +5223,7 @@
 		"@google/genai": "^1.22.0",
 		"@humanwhocodes/gitignore-to-minimatch": "1.0.2",
 		"@microsoft/tiktokenizer": "^1.0.10",
+		"@sinclair/typebox": "^0.34.41",
```
chrmarti, line 5226: Do we need this package in the main package.json?

src/lib/node/chatLibMain.ts
jeffreyhunter77: Thank you. I meant to take the whole try/catch block out. I've updated that.

package.json
jeffreyhunter77: Yes, it's used in a few places in the completions code, for example:

https://github.com/microsoft/vscode-copilot-chat/blob/b8345420d6e831fc3cc59f61f93eec3de4ed6896/src/extension/completions-core/vscode-node/lib/src/prompt/contextProviders/contextItemSchemas.ts#L6-L7

Things work right now because we get typebox indirectly through @vitest/snapshot. I think we should just have a direct dependency instead.

chrmarti:
Continued in https://github.com/microsoft/vscode-copilot-chat/pull/2131

PR #1268, opened 2025-10-08
Allow latest server.json schema in assisted NuGet MCP install flow

joelverhagen:
Resolve https://github.com/microsoft/vscode/issues/266097

PR #1373, opened 2025-10-16
Decouple server.json formatting from VS Code core

joelverhagen:
Improves on https://github.com/microsoft/vscode-copilot-chat/pull/1268.
Progress on https://github.com/microsoft/vscode/issues/266097.
Pair to https://github.com/microsoft/vscode/pull/271770.

This pushes the server.json -> runtime config mapping to the chat extension. Code is duplicated to decouple the components, based on @sandy081's recommendation.


src/extension/mcp/vscode-node/mapping/mcpGalleryService.ts
```diff
@@ -0,0 +1,499 @@
...
+					case 'oci':
+						return RegistryType.DOCKER;
+					case 'pypi':
+						return RegistryType.PYTHON;
+					case 'nuget':
+						return RegistryType.NUGET;
+					case 'mcpb':
+						return RegistryType.MCPB;
+					default:
+						return RegistryType.NODE;
```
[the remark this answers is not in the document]
joelverhagen: This is existing behavior in VS Code core.

src/extension/mcp/vscode-node/nuget.ts
```diff
@@ -305,31 +304,7 @@ stderr: ${installResult.stderr}`);
...
-
-		// provide empty registry metadata to enable VS Code data mapping
-		if (!manifest["_meta"]) {
-			manifest["_meta"] = {};
-		}
-
-		if (!manifest["_meta"]["io.modelcontextprotocol.registry/official"]) {
-			manifest["_meta"]["io.modelcontextprotocol.registry/official"] = {};
+			manifest["$schema"] = McpServerSchemaVersion_v2025_07_09.SCHEMA;
 		}
```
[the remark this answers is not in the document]
joelverhagen: This side effect is intended.

src/extension/mcp/vscode-node/nuget.ts
```diff
@@ -335,3 +342,900 @@ stderr: ${installResult.stderr}`);
 		return this.prepareServerJson(manifest, id, version);
 	}
 }
+
+export function mapServerJsonToMcpServer(input: unknown, registryType: RegistryType): Omit<IInstallableMcpServer, 'name'> | undefined {
```
sandy081, line 346: Since these functions are used inside this file, we can remove exporting them

src/extension/mcp/vscode-node/nuget.ts
```diff
@@ -335,3 +342,900 @@ stderr: ${installResult.stderr}`);
...
+	let data: any = input;
+
+	if (!data || typeof data !== 'object' || typeof data.$schema !== 'string') {
+		return undefined;
+	}
+
+	// starting from 2025-09-29, the server.json is wrapped in a "server" property
+	if (data.$schema !== McpServerSchemaVersion_v2025_07_09.SCHEMA) {
```
sandy081, line 354: Do you want to support all schemas or to start with just the schema that current nuget server supports? I would suggest to just support the `v0.1` schema and endpoint and given that nuget server supports that.

src/extension/mcp/vscode-node/nuget.ts
joelverhagen: I exported it so I could test it. Is there a better way to do that?

src/extension/mcp/vscode-node/nuget.ts
joelverhagen: NuGet.org doesn't support specific schema's per se (aside from a UI feature). It's more about what schemas exist in packages that have been published. There are some packages published to nuget.org with this older schema. The code is small to support it, so I felt that it was a nice thing to do.

PR #1400, opened 2025-10-17
tool can give alternative definition by model

bryanchen-d:
Pass model info into the `alternativeDefinition` method of `ICopilotTool` to allow tools to provide alternative definition based on different models passing in.

Fixes https://github.com/microsoft/vscode/issues/270663

### DIFF
--- a/src/extension/tools/node/manageTodoListToolEx.tsx
+++ b/src/extension/tools/node/manageTodoListToolEx.tsx
@@ -32,7 +32,7 @@ class ManageTodoListToolExtension implements ICopilotToolExtension<IManageTodoLi
 	alternativeDefinition(originTool: vscode.LanguageModelToolInformation, chatEndpoint: IChatEndpoint | undefined): vscode.LanguageModelToolInformation {
 		// specialize the tool definition for gpt-5 to reduce the frequency
 		const model = chatEndpoint?.model;
-		if (model === 'gpt-5' || model === 'gpt-5-codex') {
+		if (model === 'gpt-5-codex') {
 			return {
 				...originTool,
 				description: originTool.description?.replace('VERY frequently ', ''),


src/extension/tools/common/toolsRegistry.ts
```diff
@@ -49,14 +49,18 @@ export interface ICopilotTool<T> extends vscode.LanguageModelTool<T> {
...
-	 * alternative definition will only be applied within the Copilot extension,
-	 * not other extensions' usages via `vscode.lm.tools`.
+	 * can be driven by EXP for example, or customized based on the current model.
+	 * ⚠️ A tool using an alternative definition MUST still accept its default
+	 * parameters because the alternative definition will only be applied within
+	 * the Copilot extension, not other extensions' usages via `vscode.lm.tools`.
+	 *
+	 * @param modelInfo Optional information about the currently selected language model.
+	 *                  If provided, allows customizing the tool definition per model.
 	 */
```
roblourens, line 63: I said that we don't want to pass around extension api types here like `vscode.LanguageModelChat`. Should use `IChatEndpoint` or `modelFamily: string`

src/extension/tools/vscode-node/toolsService.ts
```diff
@@ -91,42 +92,54 @@ export class ToolsService extends BaseToolsService {
...
-			for (const ref of request.toolReferences) {
-				const usedTool = toolMap.get(ref.name);
-				if (usedTool?.tags.includes(`enable_other_tool_${tool.name}`)) {
-					return true;
+		return this.tools
+			.map(tool => {
+				// Apply model-specific alternative if available via alternativeDefinition
+				const owned = this._copilotTools.value.get(getToolName(tool.name) as ToolName);
+				if (owned?.alternativeDefinition) {
+					const alternative = owned.alternativeDefinition(endpoint);
```
roblourens: We call alternativeDefinition in the tools getter, would it make sense to change that to be model-specific? Like `getTools(model)`? Or should it stay generic and we only apply the model-specific stuff here?

src/extension/tools/common/toolsRegistry.ts
bryanchen-d: Updated using `IChatEndpoint`

src/extension/tools/vscode-node/toolsService.ts
bryanchen-d: Hmm, feels like tools consumers should all call `getEnabledTools`, then it is better the tools getter stay generic - also consider there is a cache - I wonder if we should not call `alternativeDefinition` in the tools getter as we will call it later in getEnabledTools.

pwang347:
@bryanchen-d Just an idea, would it make sense to just expose tool-specific options? e.g., for removing `VERY frequently ` we can have a feature flag like `tools.manage_todo_list.removeVeryFrequentlyClause`. And then we have a way for models to pass this as an option. Thinking from the perspective that things should be easier for even external folks to make contributions, I think centralizing the per-provider/model logic would actually be ideal.

bryanchen-d:
> @bryanchen-d Just an idea, would it make sense to just expose tool-specific options? e.g., for removing `VERY frequently ` we can have a feature flag like `tools.manage_todo_list.removeVeryFrequentlyClause`. And then we have a way for models to pass this as an option. Thinking from the perspective that things should be easier for even external folks to make contributions, I think centralizing the per-provider/model logic would actually be ideal.

Not sure if I follow, for models to pass options into tools, we already have `inputSchema` field to tell models what parameters each tool can support. 

This scenario is for passing the tool definition to models before models use the tool.

Basically the challenge here is that the to-do list tool is defined in vscode while we want to modify the description in the extension based on model. Ideally we should centralize the tool definition in the extension.

pwang347:
> > @bryanchen-d Just an idea, would it make sense to just expose tool-specific options? e.g., for removing `VERY frequently ` we can have a feature flag like `tools.manage_todo_list.removeVeryFrequentlyClause`. And then we have a way for models to pass this as an option. Thinking from the perspective that things should be easier for even external folks to make contributions, I think centralizing the per-provider/model logic would actually be ideal.
> 
> Not sure if I follow, for models to pass options into tools, we already have `inputSchema` field to tell models what parameters each tool can support.
> 
> This scenario is for passing the tool definition to models before models use the tool.
> 
> Basically the challenge here is that the to-do list tool is defined in vscode while we want to modify the description in the extension based on model. Ideally we should centralize the tool definition in the extension.

@bhavyaus has been refactoring code to centralize prompts belonging to a specific provider (e.g., see https://github.com/microsoft/vscode-copilot-chat/blob/main/src/extension/prompts/node/agent/openAIPrompts.tsx). Wondering if we could also control this behaviour from that file - but I think it could probably be done as a followup since we still have a lot of areas with these model conditionals sprinkled in several other places like the base instructions.

Played around a bit with the idea here: https://github.com/microsoft/vscode-copilot-chat/pull/1681/files#diff-982afe7aa9d84aad97affd4ffbc0cab2bdda7e39338a69ec36931d413d9d3736

bryanchen-d:
> > > @bryanchen-d Just an idea, would it make sense to just expose tool-specific options? e.g., for removing `VERY frequently ` we can have a feature flag like `tools.manage_todo_list.removeVeryFrequentlyClause`. And then we have a way for models to pass this as an option. Thinking from the perspective that things should be easier for even external folks to make contributions, I think centralizing the per-provider/model logic would actually be ideal.
> > 
> > 
> > Not sure if I follow, for models to pass options into tools, we already have `inputSchema` field to tell models what parameters each tool can support.
> > This scenario is for passing the tool definition to models before models use the tool.
> > Basically the challenge here is that the to-do list tool is defined in vscode while we want to modify the description in the extension based on model. Ideally we should centralize the tool definition in the extension.
> 
> @bhavyaus has been refactoring code to centralize prompts belonging to a specific provider (e.g., see https://github.com/microsoft/vscode-copilot-chat/blob/main/src/extension/prompts/node/agent/openAIPrompts.tsx). Wondering if we could also control this behaviour from that file - but I think it could probably be done as a followup since we still have a lot of areas with these model conditionals sprinkled in several other places like the base instructions.
> 
> Played around a bit with the idea here: https://github.com/microsoft/vscode-copilot-chat/pull/1681/files#diff-982afe7aa9d84aad97affd4ffbc0cab2bdda7e39338a69ec36931d413d9d3736

Hmm, then does the whole prompt become :
```
{system prompt generated given specific model provider}
{system prompt generated given specific model}
     { specialized instruction for individual tool}
...
{ tool definition}
...
{ user input}
```
?
Then the tool definitions will be in two different places, if we can have the tool definition generated given specific model, we can have it in one place to not confuse the models:
```
{system prompt generated given specific model provider}
{system prompt generated given specific model}
...
{ tool definition generated given specific model}
...
{ user input}
```

src/extension/tools/node/manageTodoListToolEx.tsx
```diff
@@ -0,0 +1,45 @@
...
+ */
+class ManageTodoListToolExtension implements ICopilotToolExtension<IManageTodoListParams> {
+	static readonly toolName = ToolName.CoreManageTodoList;
+	constructor(
+		@ILogService readonly _logService: ILogService
+	) { }
+
+	alternativeDefinition(originTool: vscode.LanguageModelToolInformation, chatEndpoint: IChatEndpoint | undefined): vscode.LanguageModelToolInformation {
+		// specialize the tool definition for gpt-5 to reduce the frequency
+		if (chatEndpoint?.model === 'gpt-5') {
```
roblourens: And gpt-5-codex

roblourens (approved):

PR #1401, opened 2025-10-17
Add experimental reranker service

sbatten:
Adds an experimental reranking endpoint to be fulfilled via ExP

### DIFF
--- a/src/extension/workspaceSemanticSearch/node/semanticSearchTextSearchProvider.ts
+++ b/src/extension/workspaceSemanticSearch/node/semanticSearchTextSearchProvider.ts
@@ -307,6 +307,7 @@ export class SemanticSearchTextSearchProvider implements vscode.AITextSearchProv
 					});
 				} catch (ex) {
 					// ignore rerank errors
+					this._logService.error(`SemanticSearchTextSearchProvider::provideAITextSearchResults rerank failed. error=${ex}`);
 				}
 			} else {
 				this.reportTelemetry();


src/platform/workspaceChunkSearch/common/rerankerService.ts
```diff
@@ -0,0 +1,113 @@
...
+	private get _endpoint(): string | undefined {
+		return this._expService.getTreatmentVariable<string>('rerankEndpointUrl')?.trim();
+	}
+
+	get isAvailable(): boolean {
+		return !!this._endpoint;
+	}
+
+	async rerank(query: string, documents: readonly FileChunkAndScore[], token: CancellationToken): Promise<readonly FileChunkAndScore[]> {
+		if (!documents.length || !this.isAvailable || !this._endpoint) {
```
mjbvz, line 64: Should we warn if you call this service and it's not available?

src/platform/workspaceChunkSearch/node/workspaceChunkSearchService.ts
```diff
@@ -417,6 +419,23 @@ class WorkspaceChunkSearchServiceImpl extends Disposable implements IWorkspaceCh
 					chunks: filteredChunks
 				}
 			};
+
+			// If explicit rerank is enabled, use the remote reranker
+			if (options.enableRerank && this._rerankerService.isAvailable) {
```
mjbvz, line 425: Just checking if we really want to do this in the full workspace case too

mjbvz (dismissed):

src/platform/workspaceChunkSearch/node/workspaceChunkSearchService.ts
sbatten: I'm not sure what @rebornix does in this case today. I think if you explicitly pass "enable rerank" maybe. Or maybe you shouldn't even get the "full workspace" case but be treated as if you want ranked semantic chunks.

src/platform/workspaceChunkSearch/common/rerankerService.ts
sbatten: maybe we throw?

src/platform/workspaceChunkSearch/node/workspaceChunkSearchService.ts
sbatten: @rebornix thoughts on this?

PR #1406, opened 2025-10-17
tools: make mcp tool calls actually run in parallel

connor4312:
Closes https://github.com/microsoft/vscode/issues/271989

We didn't actually parallize calls in general because we need to include
prompt sizing information through our APIs, and so need to render each
'flexGrow' group sequentially. However, we know that MCP's don't consume
any of this information today, so we can do this calls in parallel safely.

Implementation isn't the prettiest but it does the job.

### DIFF
--- a/src/extension/prompts/node/panel/toolCalling.tsx
+++ b/src/extension/prompts/node/panel/toolCalling.tsx
@@ -24,7 +24,8 @@ import { CancellationToken } from '../../../../util/vs/base/common/cancellation'
 import { toErrorMessage } from '../../../../util/vs/base/common/errorMessage';
 import { isCancellationError } from '../../../../util/vs/base/common/errors';
 import { URI, UriComponents } from '../../../../util/vs/base/common/uri';
-import { LanguageModelDataPart, LanguageModelDataPart2, LanguageModelPartAudience, LanguageModelPromptTsxPart, LanguageModelTextPart, LanguageModelTextPart2, LanguageModelToolResult } from '../../../../vscodeTypes';
+import { IInstantiationService, ServicesAccessor } from '../../../../util/vs/platform/instantiation/common/instantiation';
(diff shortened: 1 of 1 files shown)


amunger (approved):

src/extension/prompts/node/panel/toolCalling.tsx
```diff
@@ -233,14 +233,81 @@ class ToolResultElement extends PromptElement<ToolResultElementProps, void> {
...
 			}
-			this.sendToolCallTelemetry(outcome, validation);
+			sendToolCallTelemetry(props, outcome, validation, endpointProvider, telemetryService);
+		}
+
+		return { toolResult, isCancelled, extraMetadata };
+	}
+
+	let call: IToolResultElementActualProps['call'];
+	if (tool?.source instanceof LanguageModelToolMCPSource) {
```
roblourens, line 248: Which tools can use this? Ones that won't use 'flex' in a prompt-tsx tool result?

connor4312: Yes. Extension-based tools can respect the tokenBudget, but we know MCP's do not.

roblourens: How do you feel about allow-listing some tools by name so I can do things like run subagents in parallel?

connor4312: Sounds reasonable to me. We could also look and reworking the rendering a bit. Like it'd probably be fine to give every tool call in the same request the same budget (same flexGrow value). Maybe that's actually preferable to the change I made here.

roblourens: I probably can't do this for now because we don't have a way to group the parallel subagent tool calls in UI

PR #1408, opened 2025-10-17
Fix issue with internal git repo telemetry collection

IanMatthewHuff:
There was an issue that we ran into when looking at the git repo data that I added. In particular there were a lot of files that were listed as having diffs, but inside the diffJSON package the actual diff string was "".

I investigated and the issue was pretty clear. The files changed were calculated looking at the upstreamCommit ref that we calculate, but the actual individual diff string calculations in getChangeDiff were looking at a diff against HEAD, not against that computed ref. When testing this would work with staged, un-staged, or new files. But it fails if you have local commits that are not pushed upstream. In this case it correctly registers the files as changed, but doesn't pick up the diff from getDiffChanges.

Also per request from science team I added a mapping to convert the file operation (MODIFIED, DELETED, ect..) from the enum value to the readable string.

Scenarios that I tested manually with this (first scenario here is the easiest way to see the repo):

- [x] Edit a file test.ts, stage the change, commit change, then run the agent (Previous to this change you would see in the 'begin' git diff an entry for test.ts, but the diff string would be empty, now it's populated with the committed change)
- [x] Edit a file test.ts, stage and commit change, make a different edit to the file locally and don't stage, then run the agent (Should see both the committed change and the local change in the diff string for test.ts)
- [x] Test python files. The science team was seeing almost all cs / ts files in the data and not python. It think this is just an artifact of heavier usage of CS / TS for internals and we just have an initial bit of data. The file type shouldn't matter. But I did check this to make sure. py files showed up fine in the testing.
- [x] Create a new branch on the repo and don't push it upstream. Make a local change and check that it shows in the diff. This was previous issue we had with calculating the upstream so I checked this scenario again.
- [x] Create a new branch on the repo and don't push it upstream. Make a change, commit it, then check to see if it shows in the diff (basically the first scenario on a new branch that's not been pushed upstream)
- [x] Check the new string status mapping. Checked that a change with modified and deleted files showed the correct status strings for each.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
 	"icon": "assets/copilot.png",
 	"pricing": "Trial",
 	"engines": {
-		"vscode": "^1.106.0-20251022",
+		"vscode": "^1.106.0-20251023",
 		"npm": ">=9.0.0",
 		"node": ">=22.14.0"
 	},
--- a/src/extension/completions-core/vscode-node/extension/src/textDocumentManager.ts
+++ b/src/extension/completions-core/vscode-node/extension/src/textDocumentManager.ts
@@ -3,20 +3,22 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as vscode from 'vscode';
+import { window, workspace } from 'vscode';
 import { Context } from '../../lib/src/context';
 import { detectLanguage } from '../../lib/src/language/languageDetection';
 import { CopilotTextDocument, INotebookCell, INotebookDocument, ITextDocument } from '../../lib/src/textDocument';
 import { TextDocumentManager, WorkspaceFoldersChangeEvent } from '../../lib/src/textDocumentManager';
 import { transformEvent } from '../../lib/src/util/event';
 import { normalizeUri } from '../../lib/src/util/uri';
-import * as vscode from 'vscode';
-import { window, workspace } from 'vscode';
 
 // List of document URI schemes that avoid ghost text suggestions
 const ignoreUriSchemes = new Set([
 	'output', // vscode output pane (important: avoids infinite log loop)
 	'search-editor', // search results virtual document
 	'comment', // very little context available and suggestions are often bad
+	'git', // virtual file tracked by git
+	'chat-editing-snapshot-text-model', // VS Code Chat temporary editing snapshot
 ]);
 
 export function wrapDoc(_ctx: Context, doc: vscode.TextDocument): ITextDocument | undefined {
--- a/src/extension/completions-core/vscode-node/lib/src/snippy/compute.ts
+++ b/src/extension/completions-core/vscode-node/lib/src/snippy/compute.ts
@@ -7,7 +7,7 @@ const SnippyLexemeRegex = new RegExp('[_\\p{L}\\p{Nd}]+|====+|----+|####+|////+|
 // This should be kept in sync with snippy at /pkg/fingerprint/settings.go#L108
 export const MinTokenLength = 65;
 
-function lexemeLength(text: string) {
+export function lexemeLength(text: string) {
 	let i = 0;
 	let m: RegExpExecArray | null;
 	SnippyLexemeRegex.lastIndex = 0;
--- a/src/extension/completions-core/vscode-node/lib/src/snippy/test/compute.test.ts
+++ b/src/extension/completions-core/vscode-node/lib/src/snippy/test/compute.test.ts
@@ -1,0 +1,45 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import * as assert from 'assert';
+import * as SnippyCompute from '../../snippy/compute';
+
+const testMatchSource =
+	'function calculateDaysBetweenDates(begin, end) {\n    var oneDay = 24*60*60*1000; // hours*minutes*seconds*milliseconds\n    var firstDate = new Date(begin);\n    var secondDate = new Date(end);\n\n    return Math.round(Math.abs((firstDate.getTime() - secondDate.getTime())/(oneDay)));\n}';
+
+suite('Compute', function () {
+	const testCases = [
+		{ input: 'const', expected: 1 },
+		{ input: 'const foo = "bar";', expected: 7 },
+		{
+			input: `for (var i = 1; i <= 100; i++) {
+                if (i % 15 == 0) {
+                    console.log("FizzBuzz");
+                } else if (i % 3 == 0) {
+                    console.log("Fizz");
+                } else if (i % 5 == 0) {
+                    console.log("Buzz");
+                } else {
+                    console.log(i);
+                }
+                }`,
+			expected: 65,
+		},
+	];
+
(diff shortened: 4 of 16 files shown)


src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -12,6 +12,30 @@ import { ILogService } from '../../../platform/log/common/logService';
 import { ITelemetryService } from '../../../platform/telemetry/common/telemetry';
 import { IWorkspaceFileIndex } from '../../../platform/workspaceChunkSearch/node/workspaceFileIndex';
 
+// Create a mapping for the git status enum to put the actual status string in telemetry
```
IanMatthewHuff, line 16: This mapping was a smaller request from the science team I added on. I put a comment here to this effect, but I think this mapping is safe since the const enum is part of the git extension public API and while it might be added to, it wouldn't be changed or reordered.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -217,12 +241,12 @@ export class RepoInfoTelemetry {
 				};
 			}
 
-			const diffs = (await this._gitDiffService.getChangeDiffs(repoContext.rootUri, changes)).map(diff => {
+			const diffs = (await this._gitDiffService.getWorkingTreeDiffsFromRef(repoContext.rootUri, changes, upstreamCommit)).map(diff => {
```
IanMatthewHuff, line 267: The per file diff calculation here now takes the upstreamCommit we calculated earlier.

src/extension/prompt/vscode-node/gitDiffService.ts
```diff
@@ -30,6 +30,46 @@ export class GitDiffService implements IGitDiffService {
 		return repositoryOrUri;
 	}
 
+	// Get the diff between the current state of the repository and the specified ref for each of the provided changes
+	async getWorkingTreeDiffsFromRef(repositoryOrUri: Repository | Uri, changes: Change[], ref: string): Promise<Diff[]> {
```
IanMatthewHuff, line 34: When I initially was fixing this I made passing a ref optional into getChangeDiffs and it would handle both cases. But looking at how the index operations were handled in that function I felt like this function was doing a fundamentally different operation from getChangeDiffs, and while they were similar in some ways, it would be confusing to mix the two, so I added this as a separate function on the service.

src/extension/prompt/vscode-node/gitDiffService.ts
```diff
@@ -30,6 +30,46 @@ export class GitDiffService implements IGitDiffService {
...
+		for (const change of changes) {
+			if (await this._ignoreService.isCopilotIgnored(change.uri)) {
+				this._logService.debug(`[GitDiffService] Ignoring change due to content exclusion rule based on uri: ${change.uri.toString()}`);
+				continue;
+			}
+
+			let diff: string;
+			if (change.status === 7 /* UNTRACKED */) {
+				// For untracked files, generate a patch showing all content as additions
+				diff = await this._getUntrackedChangePatch(repository, change.uri);
```
sbatten, line 53: Were you able to test with changes added to the index? That seems to have special handling in the other function, so I just want to confirm it works.

IanMatthewHuff: @sbatten This was a good call to check. I had checked with committed and with unstaged in particular but not with index changes on the new function. There was a tiny bit of a rabbit hole here when I investigated, so I'll summarize.

I checked some staged scenarios with modified files and while I got the correct diff to recreate the file each time, I was a bit surprised that I didn't see INDEX_MODIFIED status in scenarios where I expected it. But in the git extension I saw that if we are using diffWith on a ref we'll generally just be getting a MODIFIED, since diffWith is not using --cached, so it's purposefully ignoring staged vs unstaged in the local repository versus the target ref. It's just going to return the diff from the target ref on modified files for everything between the current state and that ref and won't return an INDEX_MODIFIED.

To test I did the following rotation (just using shorthand here to summarize), which all looked good to me it always had MODIFIED and the full diff of all current changes.  

fileA => changeA (unstaged) => diff MODIFIED (changeA) // So the diff showed MODIFIED status with changeA in the file diff 
fileA => changeA (staged) => diff MODIFIED (changeA) 
fileA => changeA (committed) => diff MODIFIED (changeA) 
fileA => changeA (committed) + changeB (unstaged) => diff MODIFIED (changeA + changeB)
fileA => changeA (committed) + changeB (staged) => diff MODIFIED (changeA + changeB) 
fileA => changeA (committed) + changeB (committed) => diff MODIFIED (changeA + changeB) 
fileA => changeA (committed) + changeB (committed) + changeC (unstaged) => diff MODIFIED (changeA + changeB + changeC)  

However your comment here did lead me to check and find a bug in one scenario. One of the places that we do currently see an INDEX_* status is when a new file is added, then it shows up as INDEX_ADDED. In this case it returns the new file content via the getWorkingTreeDiffsFromRef, which is correct. But in the case where we have a new file and that new file is in the repository and not ignored, but is untracked we were missing that file. That file won't show up at all in diffWith against the ref. Instead it will be in either untrackedChanges or workingTreeChanges with status UNTRACKED (which it ends up in can change based on config). So after the code where we get our set of file changes from diffWith I add in any changes (by file name) that are not already picked up by diffWith, this pulls in those untracked files. So tested this out with github and ADO repos.
 
When a new file is untracked is shows up with the file contents and status: UNTRACKED. Then when it's added with git add it shows up with the file contents as INDEX_ADDED. In both cases it has the correct file contents. If you modify the file after adding it without pushing the add to upstream, you get INDEX_ADDED still with the modified file contents.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -192,7 +217,29 @@ export class RepoInfoTelemetry {
...
 
-			const changes = await this._gitService.diffWith(repoContext.rootUri, upstreamCommit);
+			// Combine our diff against the upstream commit with untracked changes, and working tree changes
+			// A change like a new untracked file could end up in either the untracked or working tree changes and won't be in the diffWith.
+			const diffChanges = await this._gitService.diffWith(repoContext.rootUri, upstreamCommit) ?? [];
+
+			const changeMap = new Map<string, Change>();
+
+			// Prority to the diffWith changes, then working tree changes, then untracked changes.
+			for (const change of diffChanges) {
```
IanMatthewHuff, line 227: This area has the change to pick up the missing UNTRACKED files. According to the git extension it looks like they could show up in either workingTreeChanges or untrackedChanges depending on config. Anything that shows up already via diffWith gets preferred as that's the same diffWith command we'll run to get the actual diff string.

src/extension/prompt/node/test/repoInfoTelemetry.spec.ts
```diff
@@ -1041,10 +1041,128 @@ suite('RepoInfoTelemetry', () => {
 
 		const diffs = JSON.parse(call[1].diffsJSON);
 		assert.strictEqual(diffs.length, 1);
-		assert.strictEqual(diffs[0].status, Status.INDEX_RENAMED);
+		assert.strictEqual(diffs[0].status, 'INDEX_RENAMED');
 		assert.ok(diffs[0].renameUri);
 	});
 
+	test('should include untracked files from both workingTreeChanges and untrackedChanges', async () => {
```
IanMatthewHuff, line 1048: New test to check that we pick up the untracked files.

roblourens (approved):

PR #1419, opened 2025-10-19
update cli/cloud session setting

rebornix:


src/platform/configuration/common/configurationService.ts
```diff
@@ -742,7 +742,8 @@ export namespace ConfigKey {
 
 		export const EnableClaudeCodeAgent = defineSetting<boolean | string | undefined>('chat.advanced.claudeCode.enabled', false);
 		export const ClaudeCodeDebugEnabled = defineSetting<boolean>('chat.advanced.claudeCode.debug', false);
-		export const CopilotCLIEnabled = defineSetting<boolean | undefined>('chat.advanced.copilotCLI.enabled', false);
+		export const CopilotCLIEnabled = defineSetting<boolean | undefined>('chat.advanced.copilotCLI.enabled', { defaultValue: false, teamDefaultValue: true });
+		export const CopilotCloudEnabled = defineSetting<boolean | undefined>('chat.advanced.copilotCodingAgent.enabled', false);
```
joshspicer, line 746:
```suggestion
		export const CopilotCloudEnabled = defineSetting<boolean | undefined>('chat.advanced.copilotCloud.enabled', false);
```
why not?

rebornix: that's the setting we shared with folks to try, since it's going away next week, I didn't try to get the name right

PR #1428, opened 2025-10-20
add exp filter for fcv1 copilot token flag

ulugbekna:


src/platform/telemetry/vscode-node/microsoftExperimentationService.ts
```diff
@@ -146,6 +146,17 @@ class DevDeviceIdFilterProvider implements IExperimentationFilterProvider {
 	}
 }
 
+class Fcv1FilterProvider implements IExperimentationFilterProvider {
```
benibenj: Maybe do not create a new filter provider and instead add the filter in `GithubAccountFilterProvider`

PR #1438, opened 2025-10-20
Copilot Cloud Agent: Custom Agents

joshspicer:
related https://github.com/microsoft/vscode/issues/270855
spec: https://github.com/github/sweagentd/blob/main/internal/customagent/SPEC.md#custom-agent-config-endpoints

Wire in custom agents to cloud agent

image

### Limitations

No API (currently) exists to query which custom agent was previously set. Will address when api updated.
https://github.com/microsoft/vscode-copilot-chat/blob/11b5b0b92b55ba0c44113a078042df4a174663ea/src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts#L214-L218


joshspicer:

PR #1440, opened 2025-10-20
Pass chat references as attachments to @github/copilot  sdk

DonJayamanne:
For #270852
For microsoft/vscode-internalbacklog#5982

### DIFF
--- a/src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
+++ b/src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
@@ -4,15 +4,13 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from 'vscode';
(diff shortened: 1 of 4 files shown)


src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
```diff
@@ -194,28 +192,24 @@ export class CopilotCLIChatSessionParticipant {
 	}
 
 	private async handleRequest(request: vscode.ChatRequest, context: vscode.ChatContext, stream: vscode.ChatResponseStream, token: vscode.CancellationToken): Promise<vscode.ChatResult | void> {
-		// Resolve the prompt with references before processing
-		const resolvedPrompt = this.resolvePrompt(request);
-		const processedRequest: ExtendedChatRequest = { ...request, prompt: resolvedPrompt };
```
DonJayamanne, line 199: I moved all of this code out, so as to avoid changing the request object, then we parse/extract stuff from request and build prompt closer to the SDK layer.

PR #1464, opened 2025-10-21
nes: convert trigger-on-active-editor-change setting to be last-edit-time-based

ulugbekna:
it now respects last edit time and doesn't trigger after N seconds


Tyriar (approved):

ulugbekna:

PR #1479, opened 2025-10-21
Prevent plan from implementing

digitarald:
Testing shows promise, even declining specific requests to implement:

image

image


pierceboggan (approved):
LGTM

PR #1480, opened 2025-10-21
Adopt resources for identifying sessions

mjbvz:
Counterpart of https://github.com/microsoft/vscode/pull/272013

@osortega Can you take a look at the copilot parts of this change? I don't understand all the ways the session id formatted, so it would be better if you handled this part


joshspicer:
working on the merge conflict

rebornix (approved):

PR #1481, opened 2025-10-21
Support for opening session on different workspace

osortega:


src/extension/chatSessions/vscode/chatSessionsUriHandler.ts
```diff
@@ -35,15 +44,52 @@ export class ChatSessionsUriHandler implements CustomUriHandler {
...
+		// Check if there's a pending session to open
+		const pendingSession = this._extensionContext.globalState.get<{
+			type: string;
+			id: string;
+			url: string;
+			branch: string;
+			timestamp: number;
+		}>(PENDING_CHAT_SESSION_STORAGE_KEY);
+
+		if (pendingSession) {
```
alexr00: I think you should also check if the open workspace actually matches the one that you saved.

PR #1485, opened 2025-10-22
wire up multifilediff part

joshspicer:
part of https://github.com/microsoft/vscode/issues/271424

This wires up the `multiDiffPart` and fetches metadata like `+/-` from GH.  

Clicking on individual elements is currently disabled.  

Screenshot 2025-10-21 at 6 51 53 PM


rebornix:
I would find it misleading if the part says there are changes but opening the diff editor shows no change. A uri/document provider with a special scheme (naming it unique to avoid conflict with ghpri) might be better.

joshspicer:
Right I don't think it should be left like this.  At the minimum something like https://github.com/microsoft/vscode/pull/272605, or wiring it up so that GHPR (if installed, or prompted to install) can compute the diffs

joshspicer:
depends on https://github.com/microsoft/vscode/pull/272605

PR #1495, opened 2025-10-22
Fix applyPatch-tool emitting incorrect edits

jrieken:
Started to use applyPatch in stests (`joh/slow-meerkat`) which resulted in corrected files because of bogous edits. The culprit is that the applyPatch-tool computes multiple text edits and then dispatches them to the stream via individual calls. The semantics of that is that  call `N` knows new offsets after applying edits from the `N-1` call. Tho, the calls aren't made like that by the tool. Instead, it should dispatch its edit in a single call


connor4312:

PR #1517, opened 2025-10-22
Add resolved model

lramos15:
Initial pass of adding the server side resolved model to our logging

Also add exp for auto mode hint


src/platform/endpoint/common/automodeService.ts
```diff
@@ -168,9 +170,11 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 			headers['Copilot-Session-Token'] = sessionToken;
 		}
 
+		const autoModeHint = this._expService.getTreatmentVariable<string>('copilotchat.autoModelHint') || 'auto';
```
isidorn, line 173: small typo. Variable should be called `autoModelHint` not `autoModeHint`
@lramos15

lramos15: We call it mode in some places too. idk why 😅 . but good point

PR #1534, opened 2025-10-23
Change CCA chat sessions type

osortega:
Using copilot-swe-agent type to maintain compatibility


Tyriar (approved):

rebornix:
let's not merge this since it doesn't work as expected.

PR #1536, opened 2025-10-23
Saving the plan should not automatically send the prompt

pierceboggan:
This is necessary in circumstances where you want to alter the model used for the plan OR give feeback on open questions presented in the plan first.

### DIFF
--- a/assets/agents/Plan.agent.md
+++ b/assets/agents/Plan.agent.md
@@ -8,7 +8,6 @@ handoffs:
   - label: Open in Editor
     agent: agent
     prompt: Save the plan
-    send: true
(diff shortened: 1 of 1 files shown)


digitarald (approved):

OrenMe:
@pierceboggan maybe allow also setting model for when switching to another mode
I know that customer agent modes have ability to define model but if we want to have advanced workflows this can be very useful

OrenMe:
Or maybe even allow handoffs to remain in same mode but delegate to runSubagent 
A good example is writing the plan to file, who wants this to take the main context window quata?

PR #1549, opened 2025-10-23
InlineChatIntent

jrieken:


src/extension/tools/node/editFileToolResult.tsx
```diff
@@ -76,7 +76,10 @@ export class EditFileResult extends PromptElement<IEditFileResultProps> {
 				healedEdits.push({ file: filePath, healing: file.healed });
 			}
 
-			const diagnostics = !this.testContext.isInSimulationTests && this.configurationService.getExperimentBasedConfig(ConfigKey.AutoFixDiagnostics, this.experimentationService) && !(file.isNotebook)
+			const diagnostics = (this.props.diagnosticsTimeout === undefined || this.props.diagnosticsTimeout >= 0)
```
jrieken, line 79: fyi @roblourens - I know there is a discussion around this feature and I don't wanna get to involved in that, but for inline chat we really don't need to wait for diagnostics because the tool result won't be fed back to the LM

PR #1555, opened 2025-10-23
Avoid waiting for shell integration

DonJayamanne:
For microsoft/vscode#270852
For microsoft/vscode-internalbacklog#5982


src/extension/chatSessions/vscode-node/copilotCLITerminalIntegration.ts
```diff
@@ -185,7 +185,7 @@ ELECTRON_RUN_AS_NODE=1 "${process.execPath}" "${path.join(storageLocation, COPIL
 			await new Promise<void>(resolve => this._register(disposableTimeout(resolve, 500))); // Wait a bit to ensure the terminal is ready
```
anthonykim1, line 185: Does removing this break anything? 

If it was to avoid potentially a dup looking command in the terminal, I think we should be in better position with new `terminal.integrated.shellIntegration.timeout` and some improvements that came along the way. (default value -1 for the setting should be enough)

PR #1582, opened 2025-10-24
Add new collaborators in PR cache check

kycutler:

PR #1583, opened 2025-10-24
use live collaborators list

joaomoreno:
alternative to #1582


ulugbekna:
since we have Lad's approval, thought I would try merging to unblock Kyle's PR

kycutler:
Yeah I think I don't show up on the collaborator list for some reason. When I query it directly, I see 30 contributors whereas the old list had nearly 60 even without the new folks. Not sure how that list is configured either.

PR #1595, opened 2025-10-24
Fix for deeplink opening on already open workspace

osortega:
This fixes an issue with deeplink where we would fail to open a chat session if you already have a vscode window open with the workspace


rebornix:

rebornix:

osortega:

PR #1596, opened 2025-10-24
handle problem_statement undefined from capi

joshspicer:


Tyriar (dismissed):

joshspicer:
same goal as https://github.com/microsoft/vscode-copilot-chat/commit/0f78648d8b8e5e1657f9f751c6bb5f95bdcc6660

@mjbvz we're going to go with this approach as we fear `jobInfo` may also be able to be undefined on old sessions

rebornix (approved):

PR #1600, opened 2025-10-24
Rename placeholder view to match the real one

roblourens:


roblourens:

joshspicer (approved):

PR #1601, opened 2025-10-24
move possibly async code into async function

amunger:
https://github.com/microsoft/vscode-jupyter/issues/17119

having the possibly-async line in the middle of the invoke function seemed to break the rendering.


amunger:
nevermind, still got a repro

PR #1650, opened 2025-10-27
Add VSC Model B prompt variants

AbdelrahmanAbouelenin:
### Summary
Implements two distinct VSC model custom prompt variants (A and B).

### Implementation
- Added `VSC_MODEL_HASHES_A` (6 hashes) and `VSC_MODEL_HASHES_B` (2 hashes) in `chatModelCapabilities.ts`
- Implemented `isVSCModelA()` and `isVSCModelB()` for SHA256-based detection
- Dual resolver pattern: `VSCModelPromptResolverA` and `VSCModelPromptResolverB`


src/extension/prompts/node/agent/vscModelPrompts.tsx
```diff
@@ -146,16 +147,164 @@ class VSCModelPrompt extends PromptElement<DefaultAgentPromptProps> {
...
+					</> :
+					<>
+						Don't try to edit an existing file without reading it first, so you can make changes properly.<br />
+						Use the {ToolName.EditFile} tool to edit files. When editing files, group your changes by file.<br />
+						NEVER show the changes to the user, just call the tool, and the edits will be applied and shown to the user.<br />
+						NEVER print a codeblock that represents a change to a file, use {ToolName.EditFile} instead.<br />
+					</>}
+				<GenericEditingTips {...this.props} />
+			</Tag>}
+			<NotebookInstructions {...this.props} />
```
karthiknadig, line 330: @bhavyaus the use of `this.props` here, isn't that duplicating the instructions twice?

src/extension/prompts/node/agent/vscModelPrompts.tsx
```diff
@@ -146,16 +147,164 @@ class VSCModelPrompt extends PromptElement<DefaultAgentPromptProps> {
...
+	}
+
+	resolvePrompt(endpoint: IChatEndpoint): PromptConstructor | undefined {
+		return VSCModelPromptA;
+	}
+}
 
-	static matchesModel(endpoint: IChatEndpoint): Promise<boolean> | boolean {
-		return isVSCModel(endpoint);
+class VSCModelPromptResolverB implements IAgentPrompt {
```
bhavyaus, line 369: You don't need to register another resolver. You could just update the matches model  to return for modelA or ModelB

And then resolve the right prompt in the resolveModel.

src/extension/prompts/node/agent/vscModelPrompts.tsx
bhavyaus: The props are used to perform checks to see if any instructions should be shown for the endpoint.  No duplucation.

src/extension/prompts/node/agent/vscModelPrompts.tsx
AbdelrahmanAbouelenin: can you elaborate?  we can check in match func we can return true for modelA or ModelB but we will need to recheck in resolveprompt and we cannot do that since it's a sync func while isVSCModelA and isVSCModelB are async funcs.
however, we can have a state var that gets set in matchmodels and then we can use this value in resolveprompt, does that sounds good?

bhavyaus: This is fine for now. Resolving comment

bhavyaus (dismissed):

karthiknadig (dismissed):

karthiknadig (approved):

roblourens (approved):

PR #1661, opened 2025-10-27
Hide custom open ai compatible from stable

lramos15:


linonetwo:
Please revert this. As copilot have weekly rate limit, we want to use our own stable providers and switch between multiple of providers. **While keep access to previous chat context**.

If copilot could provide stable service, then this is not a problem. But at this point, block us away from both pro+ subscription and our custom provider is not reasonable.

PR #1680, opened 2025-10-28
Pull xAI BYOK models directly from API

pwang347:
Additional models are now available for selection in BYOK.

image

### DIFF
--- a/src/extension/byok/vscode-node/xAIProvider.ts
+++ b/src/extension/byok/vscode-node/xAIProvider.ts
@@ -82,10 +82,11 @@ export class XAIBYOKLMProvider extends BaseOpenAICompatibleLMProvider {
 				throw new Error('Invalid response format from xAI API');
 			}
 			this._logService.trace(`Fetched ${data.models.length} language models from xAI`);
-			const modelList: BYOKKnownModels = this._knownModels ? { ...this._knownModels } : {};
+			const modelList: BYOKKnownModels = {};
 			for (const model of data.models) {
 				if (this._knownModels && this._knownModels[model.id]) {
-					continue; // Skip known models so that we can override these correctly from CDN
+					modelList[model.id] = this._knownModels[model.id];
+					continue;
 				}
 
 				// Add new model with reasonable defaults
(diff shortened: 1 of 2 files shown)


src/extension/byok/vscode-node/xAIProvider.ts
```diff
@@ -32,4 +51,69 @@ export class XAIBYOKLMProvider extends BaseOpenAICompatibleLMProvider {
...
+					'Content-Type': 'application/json'
+				}
+			});
+
+			const data = await response.json() as XAIListLanguageModelsAPIResponse;
+			if (!data.models || !Array.isArray(data.models)) {
+				throw new Error('Invalid response format from xAI API');
+			}
+			this._logService.trace(`Fetched ${data.models.length} language models from xAI`);
+			const modelList: BYOKKnownModels = this._knownModels ? { ...this._knownModels } : {};
```
lramos15: Doesn't this code have the risk of a model that is in our CDN but not available in your API key shwoing up in the list? Since we always apply our CDN list? 

The logic should be we iterate through data.model and take the capabilities from the CDN if present else sensible defaults. That way xAI is the source of truth since they know best

src/extension/prompts/node/agent/xAIPrompts.tsx
```diff
@@ -121,6 +123,10 @@ class DefaultGrokCodeFastAgentPrompt extends PromptElement<DefaultAgentPromptPro
 class XAIPromptResolver implements IAgentPrompt {
 	static readonly familyPrefixes = ['grok-code'];
 
+	static async matchesModel(endpoint: IChatEndpoint): Promise<boolean> {
+		return endpoint.family === BYOK_PROVIDER_NAME;
```
lramos15: What is this used for? It seems unrelated to the rest of the PR

src/extension/byok/vscode-node/xAIProvider.ts
pwang347: fair point, updated!

src/extension/prompts/node/agent/xAIPrompts.tsx
pwang347: I can move this to a separate PR, this just enables the xAI prompts for the BYOK models as well. I tested that the behaviour for models like Grok 4 and other BYOK are relatively the same after the change and nothing is broken.

PR #1684, opened 2025-10-28
Remove All Tools

24anisha:
To debug whether tool disabling logic is working (it's not)


src/extension/tools/vscode-node/toolsService.ts
```diff
@@ -69,7 +75,11 @@ export class ToolsService extends BaseToolsService {
 		@ILogService logService: ILogService
 	) {
 		super(logService);
-		this._copilotTools = new Lazy(() => new Map(ToolRegistry.getTools().map(t => [t.toolName, instantiationService.createInstance(t)] as const)));
+		this._copilotTools = new Lazy(() => {
+			const filteredTools = ToolRegistry.getTools()
+				.filter(t => !ToolsService.DisabledTools.includes(t.toolName));
```
connor4312, line 78: This won't do anything to disable the tool and might calls other issues. The `tools` array on this class is the point of truth for what tools are available

PR #1717, opened 2025-10-30
first cut - migrate settings marked with INTERNAL and adopt to advanced tag

sandy081:


src/platform/configuration/common/configurationService.ts
```diff
@@ -589,16 +640,86 @@ export namespace ConfigKey {
...
 	/**
-	 * Internal and debugging settings that should be hidden from users.
-	 *
-	 * Features should only be in this list temporarily, moving on to experimental to be accessible to early adopters.
-	*/
-	export namespace Internal {
+	 * Advanced experimental settings that are available for all users to configure.
+	 * These should only be in this list temporarily. Either should be made stable or removed.
+	 */
+	export namespace AdvancedExperimental {
```
ulugbekna, line 646: I think having such precise and many namespaces makes it painful for developers to refer to settings. Can we just move previously internal settings from Internal to outer namespace along with other public settings?

package.json
```diff
@@ -3338,6 +3338,469 @@
...
+						"markdownDescription": "%github.copilot.config.projectLabels.expanded%",
+						"tags": [
+							"advanced",
+							"experimental",
+							"onExp"
+						]
+					},
+					"github.copilot.chat.projectLabels.chat": {
+						"type": "boolean",
+						"default": false,
```
ulugbekna, line 3355: Having to duplicate defaults here and in configurationService.ts will very likely result in a bug at some point in future

src/platform/configuration/common/configurationService.ts
sandy081: IMO this gives better distinction and easy to manage them going forward just like Internal, especially we have lot of chat settings.

package.json
sandy081: That's true. I have the same feeling. But I do not want to touch that as it was there before and do not want to break.

bpasero (approved):

alexdima (dismissed):
Nice!

PR #1746, opened 2025-10-31
Safely destructure custom agents response to handle null/undefined. Fix https://github.com/microsoft/vscode/issues/274437

tamuratak:
Safely destructure custom agents response to handle null/undefined. Fix https://github.com/microsoft/vscode/issues/274437


joshspicer:
+1 `getCustomAgentsWithToken` expects to return `GetCustomAgentsResponse` but the function it returns (`makeGitHubAPIRequest`) returns `any`, allowing this through.  

Please consider moving the check into `getCustomAgentsWithToken`, perhaps now returning `GetCustomAgentsResponse|undefined`

PR #1752, opened 2025-11-02
Remove Copilot CLI Agent manager and introduce ICopilotCLISession interface

DonJayamanne:
For microsoft/vscode#270852
For microsoft/vscode-internalbacklog#5982

PR #1753, opened 2025-11-02
Refactor creation of work trees

DonJayamanne:


src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
```diff
@@ -35,12 +35,7 @@ export class CopilotCLIWorktreeManager {
 	constructor(
 		@IVSCodeExtensionContext private readonly extensionContext: IVSCodeExtensionContext) { }
 
-	async createWorktreeIfNeeded(sessionId: string, stream: vscode.ChatResponseStream): Promise<string | undefined> {
-		const isolationEnabled = this._sessionIsolation.get(sessionId) ?? false;
-		if (!isolationEnabled) {
-			return undefined;
-		}
-
+	async createWorktree(stream: vscode.ChatResponseStream): Promise<string | undefined> {
```
DonJayamanne, line 38: @osortega 
I think getting the caller to decide whether to call this method or not is an improvement, as opposed to giving more responsibility to this method.
Callers already have all of the information using the public methods in this class.

I believe that will also improve readability as sometimes we pass in `untitled id` and other cases we pass in the `real session id` (what confuses me is the fact that here we MUST pass in the old `untitled id` & not the `real session id`).

I'd like to remove that ambiguity from this class/code.

DonJayamanne: Please ignore the other commits, thats from another PR https://github.com/microsoft/vscode-copilot-chat/pull/1752/files

It was while trying to work on that I ran into this and was confused what session id had to be passed into the method.

PR #1775, opened 2025-11-03
Linkify line numbers to enable precise line navigation from chat responses

vijayupadya:
When chat response mentions a file with a line hint, for example:

utils/logger.ts (line 42)
in lines 10–12 of server.py
config.json, lines 5 through 9
at line 7 in README.md

this change now upgrades that file reference so clicking it opens the file directly at the specified line (first line of a range for now). This works whether the line annotation appears after the filename or before it.

image

### DIFF
--- a/src/extension/linkify/common/filePathLinkifier.ts
+++ b/src/extension/linkify/common/filePathLinkifier.ts
@@ -79,7 +79,13 @@ export class FilePathLinkifier implements IContributedLinkifier {
 			pathText ??= match.groups?.['inlineCodePath'] ?? match.groups?.['plainTextPath'] ?? '';
 
 			parts.push(this.resolvePathText(pathText, context)
-				.then(uri => uri ? new LinkifyLocationAnchor(uri) : matched));
+				.then(uri => {
+					if (!uri) {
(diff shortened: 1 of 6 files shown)


src/extension/linkify/common/filePathLinkifier.ts
```diff
@@ -47,7 +47,7 @@ export class FilePathLinkifier implements IContributedLinkifier {
 	) { }
 
 	async linkify(text: string, context: LinkifierContext, token: CancellationToken): Promise<LinkifiedText> {
-		const parts: Array<Promise<LinkifiedPart> | LinkifiedPart> = [];
+		const parts: LinkifiedPart[] = [];
```
mjbvz, line 50: We have an array of promises because we want to resolve all the links in parallel instead of sequentially. I'd prefer keeping it like that

mjbvz (commented):
I think we should re-evaluate if we can get models to generate proper links that we can easily parse. Earlier models were not great at this and we still likely need to be a little pragmatic, but I'm really not a fan of extending the link handler to try parsing sentences like this

PR #1788, opened 2025-11-04
update api to avoid warnings in builtin

justschen:
moves over to use finalized api


mjbvz (approved):
vscode.d.ts changes look good. It's pulling in a few other other finalized apis from this iteration too

PR #1803, opened 2025-11-05
Upstream model generated file and line linkification

vijayupadya:
Fixes https://github.com/microsoft/vscode/issues/276767

**Motivation**

Move file path linkification upstream to the model so it emits canonical, machine‑parsable Markdown links ([path](path#Lstart[-end])), reducing reliance on heuristic fallback and enabling precise line anchors. Linkification falls back to current component if model does not handle it.

Also ensure consistent behavior across all model families (OpenAI/GPT‑5, Anthropic/Claude, etc.) while eliminating duplicated prompt text.

This is example response with line number linkification:

image

### DIFF
--- a/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-arctic-fox.spec.snap
+++ b/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-arctic-fox.spec.snap
@@ -47,7 +47,8 @@ You are producing text that will be rendered as markdown by the VS Code UI. Foll
 - Markdown text. Use structure only when it helps scanability.
 - Headers: optional; short Title Case (1-3 words) wrapped in **…**; no blank line before the first bullet; add only if they truly help.
 - Bullets: use - ; merge related points; keep to one line when possible; 4-6 per list ordered by importance; keep phrasing consistent.
-- Monospace: backticks for commands/paths/env vars/code ids and inline examples; use for literal keyword bullets; never combine with **.
+- Monospace: backticks for commands, env vars, and code identifiers; never combine with **.
+- File path and line number formatting rules are defined in the fileLinkification section below.
 - Code samples or multi-line snippets should be wrapped in fenced code blocks; include an info string as often as possible.
 - Structure: group related bullets; order sections general → specific → supporting; for subsections, start with a bolded keyword bullet, then items; match complexity to the task.
 - Tone: collaborative, concise, factual; present tense, active voice; self-contained; no "above/below"; parallel wording.
@@ -55,10 +56,42 @@ You are producing text that will be rendered as markdown by the VS Code UI. Foll
 - Adaptation: code explanations → precise, structured with code refs; simple tasks → lead with outcome; big changes → logical walkthrough + rationale + next actions; casual one-offs → plain sentences, no headers/bullets.
 </final_answer_structure_and_style_guidelines>
 <special_formatting>
-When referring to a filename or symbol in the user's workspace, wrap it in backticks.
-<example>
-The class `Person` is in `src/models/person.ts`.
-</example>
+Use proper Markdown formatting: - Wrap symbol names (classes, methods, variables) in backticks: `MyClass`, `handleClick()`
+- When mentioning files or line numbers, always follow the rules in fileLinkification section below:<fileLinkification>
+When mentioning files or line numbers, always convert them to markdown links using workspace-relative paths and 1-based line numbers.
+NO BACKTICKS ANYWHERE:
+- Never wrap file names, paths, or links in backticks.
+- Never use inline-code formatting for any file reference.
+
+REQUIRED FORMATS:
+- File: [path/file.ts](path/file.ts)
+- Line: [file.ts](file.ts#L10)
+- Range: [file.ts](file.ts#L10-L12)
+
+PATH RULES:
+- Without line numbers: Display text must match the target path.
+- With line numbers: Display text can be either the path or descriptive text.
+- Use '/' only; strip drive letters and external folders.
+- Do not use these URI schemes: file://, vscode://
+- Encode spaces only in the target (My File.md → My%20File.md).
+- Non-contiguous lines require separate links. NEVER use comma-separated line references like #L10-L12, L20.
+- Valid formats: [file.ts](file.ts#L10) or [file.ts#L10] only. Invalid: ([file.ts#L10]) or [file.ts](file.ts)#L10
+
+USAGE EXAMPLES:
+- With path as display: The handler is in [src/handler.ts](src/handler.ts#L10).
+- With descriptive text: The [widget initialization](src/widget.ts#L321) runs on startup.
+- Bullet list: [Init widget](src/widget.ts#L321)
+- File only: See [src/config.ts](src/config.ts) for settings.
+
+FORBIDDEN (NEVER OUTPUT):
+- Inline code: `file.ts`, `src/file.ts`, `L86`.
+- Plain text file names: file.ts, chatService.ts.
+- References without links when mentioning specific file locations.
+- Specific line citations without links ("Line 86", "at line 86", "on line 25").
+- Combining multiple line references in one link: [file.ts#L10-L12, L20](file.ts#L10-L12, L20)
+
+
+</fileLinkification>
 
 </special_formatting>
 
--- a/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-opus-4.5.spec.snap
+++ b/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-opus-4.5.spec.snap
@@ -43,12 +43,42 @@ Important Reminder: Avoid referencing Notebook Cell Ids in user messages. Use ce
 Important Reminder: Markdown cells cannot be executed
 </notebookInstructions>
 <outputFormatting>
-Use proper Markdown formatting in your answers. When referring to a filename or symbol in the user's workspace, wrap it in backticks.
-<example>
-The class `Person` is in `src/models/person.ts`.
-The function `calculateTotal` is defined in `lib/utils/math.ts`.
-You can find the configuration in `config/app.config.json`.
-</example>
+Use proper Markdown formatting. When referring to symbols (classes, methods, variables) in user's workspace wrap in backticks. For file paths and line number rules, see fileLinkification section
+<fileLinkification>
+When mentioning files or line numbers, always convert them to markdown links using workspace-relative paths and 1-based line numbers.
+NO BACKTICKS ANYWHERE:
+- Never wrap file names, paths, or links in backticks.
+- Never use inline-code formatting for any file reference.
+
+REQUIRED FORMATS:
+- File: [path/file.ts](path/file.ts)
+- Line: [file.ts](file.ts#L10)
+- Range: [file.ts](file.ts#L10-L12)
+
+PATH RULES:
+- Without line numbers: Display text must match the target path.
+- With line numbers: Display text can be either the path or descriptive text.
+- Use '/' only; strip drive letters and external folders.
+- Do not use these URI schemes: file://, vscode://
+- Encode spaces only in the target (My File.md → My%20File.md).
+- Non-contiguous lines require separate links. NEVER use comma-separated line references like #L10-L12, L20.
+- Valid formats: [file.ts](file.ts#L10) or [file.ts#L10] only. Invalid: ([file.ts#L10]) or [file.ts](file.ts)#L10
+
+USAGE EXAMPLES:
+- With path as display: The handler is in [src/handler.ts](src/handler.ts#L10).
+- With descriptive text: The [widget initialization](src/widget.ts#L321) runs on startup.
+- Bullet list: [Init widget](src/widget.ts#L321)
+- File only: See [src/config.ts](src/config.ts) for settings.
+
+FORBIDDEN (NEVER OUTPUT):
+- Inline code: `file.ts`, `src/file.ts`, `L86`.
+- Plain text file names: file.ts, chatService.ts.
+- References without links when mentioning specific file locations.
+- Specific line citations without links ("Line 86", "at line 86", "on line 25").
+- Combining multiple line references in one link: [file.ts#L10-L12, L20](file.ts#L10-L12, L20)
+
+
+</fileLinkification>
 
 </outputFormatting>
 
--- a/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_tools-arctic-fox.spec.snap
+++ b/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_tools-arctic-fox.spec.snap
@@ -47,7 +47,8 @@ You are producing text that will be rendered as markdown by the VS Code UI. Foll
 - Markdown text. Use structure only when it helps scanability.
 - Headers: optional; short Title Case (1-3 words) wrapped in **…**; no blank line before the first bullet; add only if they truly help.
 - Bullets: use - ; merge related points; keep to one line when possible; 4-6 per list ordered by importance; keep phrasing consistent.
-- Monospace: backticks for commands/paths/env vars/code ids and inline examples; use for literal keyword bullets; never combine with **.
+- Monospace: backticks for commands, env vars, and code identifiers; never combine with **.
+- File path and line number formatting rules are defined in the fileLinkification section below.
 - Code samples or multi-line snippets should be wrapped in fenced code blocks; include an info string as often as possible.
 - Structure: group related bullets; order sections general → specific → supporting; for subsections, start with a bolded keyword bullet, then items; match complexity to the task.
 - Tone: collaborative, concise, factual; present tense, active voice; self-contained; no "above/below"; parallel wording.
@@ -55,10 +56,42 @@ You are producing text that will be rendered as markdown by the VS Code UI. Foll
 - Adaptation: code explanations → precise, structured with code refs; simple tasks → lead with outcome; big changes → logical walkthrough + rationale + next actions; casual one-offs → plain sentences, no headers/bullets.
 </final_answer_structure_and_style_guidelines>
 <special_formatting>
-When referring to a filename or symbol in the user's workspace, wrap it in backticks.
-<example>
-The class `Person` is in `src/models/person.ts`.
-</example>
+Use proper Markdown formatting: - Wrap symbol names (classes, methods, variables) in backticks: `MyClass`, `handleClick()`
+- When mentioning files or line numbers, always follow the rules in fileLinkification section below:<fileLinkification>
+When mentioning files or line numbers, always convert them to markdown links using workspace-relative paths and 1-based line numbers.
+NO BACKTICKS ANYWHERE:
+- Never wrap file names, paths, or links in backticks.
+- Never use inline-code formatting for any file reference.
+
+REQUIRED FORMATS:
+- File: [path/file.ts](path/file.ts)
+- Line: [file.ts](file.ts#L10)
+- Range: [file.ts](file.ts#L10-L12)
+
+PATH RULES:
+- Without line numbers: Display text must match the target path.
+- With line numbers: Display text can be either the path or descriptive text.
+- Use '/' only; strip drive letters and external folders.
+- Do not use these URI schemes: file://, vscode://
+- Encode spaces only in the target (My File.md → My%20File.md).
+- Non-contiguous lines require separate links. NEVER use comma-separated line references like #L10-L12, L20.
+- Valid formats: [file.ts](file.ts#L10) or [file.ts#L10] only. Invalid: ([file.ts#L10]) or [file.ts](file.ts)#L10
+
+USAGE EXAMPLES:
+- With path as display: The handler is in [src/handler.ts](src/handler.ts#L10).
+- With descriptive text: The [widget initialization](src/widget.ts#L321) runs on startup.
+- Bullet list: [Init widget](src/widget.ts#L321)
+- File only: See [src/config.ts](src/config.ts) for settings.
+
+FORBIDDEN (NEVER OUTPUT):
+- Inline code: `file.ts`, `src/file.ts`, `L86`.
+- Plain text file names: file.ts, chatService.ts.
+- References without links when mentioning specific file locations.
+- Specific line citations without links ("Line 86", "at line 86", "on line 25").
+- Combining multiple line references in one link: [file.ts#L10-L12, L20](file.ts#L10-L12, L20)
+
+
+</fileLinkification>
 
 </special_formatting>
 
--- a/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_tools-claude-opus-4.5.spec.snap
+++ b/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_tools-claude-opus-4.5.spec.snap
@@ -45,12 +45,42 @@ Important Reminder: Avoid referencing Notebook Cell Ids in user messages. Use ce
 Important Reminder: Markdown cells cannot be executed
 </notebookInstructions>
 <outputFormatting>
-Use proper Markdown formatting in your answers. When referring to a filename or symbol in the user's workspace, wrap it in backticks.
-<example>
-The class `Person` is in `src/models/person.ts`.
-The function `calculateTotal` is defined in `lib/utils/math.ts`.
-You can find the configuration in `config/app.config.json`.
-</example>
+Use proper Markdown formatting. When referring to symbols (classes, methods, variables) in user's workspace wrap in backticks. For file paths and line number rules, see fileLinkification section
+<fileLinkification>
+When mentioning files or line numbers, always convert them to markdown links using workspace-relative paths and 1-based line numbers.
+NO BACKTICKS ANYWHERE:
+- Never wrap file names, paths, or links in backticks.
+- Never use inline-code formatting for any file reference.
+
+REQUIRED FORMATS:
+- File: [path/file.ts](path/file.ts)
+- Line: [file.ts](file.ts#L10)
+- Range: [file.ts](file.ts#L10-L12)
+
+PATH RULES:
+- Without line numbers: Display text must match the target path.
+- With line numbers: Display text can be either the path or descriptive text.
+- Use '/' only; strip drive letters and external folders.
+- Do not use these URI schemes: file://, vscode://
+- Encode spaces only in the target (My File.md → My%20File.md).
+- Non-contiguous lines require separate links. NEVER use comma-separated line references like #L10-L12, L20.
+- Valid formats: [file.ts](file.ts#L10) or [file.ts#L10] only. Invalid: ([file.ts#L10]) or [file.ts](file.ts)#L10
+
+USAGE EXAMPLES:
+- With path as display: The handler is in [src/handler.ts](src/handler.ts#L10).
+- With descriptive text: The [widget initialization](src/widget.ts#L321) runs on startup.
+- Bullet list: [Init widget](src/widget.ts#L321)
+- File only: See [src/config.ts](src/config.ts) for settings.
+
+FORBIDDEN (NEVER OUTPUT):
+- Inline code: `file.ts`, `src/file.ts`, `L86`.
+- Plain text file names: file.ts, chatService.ts.
+- References without links when mentioning specific file locations.
+- Specific line citations without links ("Line 86", "at line 86", "on line 25").
+- Combining multiple line references in one link: [file.ts#L10-L12, L20](file.ts#L10-L12, L20)
+
+
+</fileLinkification>
 
 </outputFormatting>
 
--- a/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-cache_BPs-arctic-fox.spec.snap
+++ b/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-cache_BPs-arctic-fox.spec.snap
@@ -47,7 +47,8 @@ You are producing text that will be rendered as markdown by the VS Code UI. Foll
 - Markdown text. Use structure only when it helps scanability.
 - Headers: optional; short Title Case (1-3 words) wrapped in **…**; no blank line before the first bullet; add only if they truly help.
 - Bullets: use - ; merge related points; keep to one line when possible; 4-6 per list ordered by importance; keep phrasing consistent.
-- Monospace: backticks for commands/paths/env vars/code ids and inline examples; use for literal keyword bullets; never combine with **.
+- Monospace: backticks for commands, env vars, and code identifiers; never combine with **.
+- File path and line number formatting rules are defined in the fileLinkification section below.
 - Code samples or multi-line snippets should be wrapped in fenced code blocks; include an info string as often as possible.
 - Structure: group related bullets; order sections general → specific → supporting; for subsections, start with a bolded keyword bullet, then items; match complexity to the task.
 - Tone: collaborative, concise, factual; present tense, active voice; self-contained; no "above/below"; parallel wording.
@@ -55,10 +56,42 @@ You are producing text that will be rendered as markdown by the VS Code UI. Foll
 - Adaptation: code explanations → precise, structured with code refs; simple tasks → lead with outcome; big changes → logical walkthrough + rationale + next actions; casual one-offs → plain sentences, no headers/bullets.
 </final_answer_structure_and_style_guidelines>
 <special_formatting>
-When referring to a filename or symbol in the user's workspace, wrap it in backticks.
-<example>
-The class `Person` is in `src/models/person.ts`.
-</example>
+Use proper Markdown formatting: - Wrap symbol names (classes, methods, variables) in backticks: `MyClass`, `handleClick()`
+- When mentioning files or line numbers, always follow the rules in fileLinkification section below:<fileLinkification>
+When mentioning files or line numbers, always convert them to markdown links using workspace-relative paths and 1-based line numbers.
+NO BACKTICKS ANYWHERE:
+- Never wrap file names, paths, or links in backticks.
+- Never use inline-code formatting for any file reference.
+
+REQUIRED FORMATS:
+- File: [path/file.ts](path/file.ts)
+- Line: [file.ts](file.ts#L10)
+- Range: [file.ts](file.ts#L10-L12)
+
+PATH RULES:
+- Without line numbers: Display text must match the target path.
+- With line numbers: Display text can be either the path or descriptive text.
+- Use '/' only; strip drive letters and external folders.
+- Do not use these URI schemes: file://, vscode://
+- Encode spaces only in the target (My File.md → My%20File.md).
+- Non-contiguous lines require separate links. NEVER use comma-separated line references like #L10-L12, L20.
+- Valid formats: [file.ts](file.ts#L10) or [file.ts#L10] only. Invalid: ([file.ts#L10]) or [file.ts](file.ts)#L10
+
+USAGE EXAMPLES:
+- With path as display: The handler is in [src/handler.ts](src/handler.ts#L10).
+- With descriptive text: The [widget initialization](src/widget.ts#L321) runs on startup.
+- Bullet list: [Init widget](src/widget.ts#L321)
+- File only: See [src/config.ts](src/config.ts) for settings.
+
+FORBIDDEN (NEVER OUTPUT):
+- Inline code: `file.ts`, `src/file.ts`, `L86`.
+- Plain text file names: file.ts, chatService.ts.
+- References without links when mentioning specific file locations.
+- Specific line citations without links ("Line 86", "at line 86", "on line 25").
+- Combining multiple line references in one link: [file.ts#L10-L12, L20](file.ts#L10-L12, L20)
+
+
+</fileLinkification>
 
 </special_formatting>
 
--- a/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-cache_BPs-claude-opus-4.5.spec.snap
+++ b/src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-cache_BPs-claude-opus-4.5.spec.snap
@@ -32,12 +32,42 @@ You don't currently have any tools available for running terminal commands. If t
 Tools can be disabled by the user. You may see tools used previously in the conversation that are not currently available. Be careful to only use the tools that are currently available to you.
 </toolUseInstructions>
 <outputFormatting>
-Use proper Markdown formatting in your answers. When referring to a filename or symbol in the user's workspace, wrap it in backticks.
-<example>
-The class `Person` is in `src/models/person.ts`.
-The function `calculateTotal` is defined in `lib/utils/math.ts`.
-You can find the configuration in `config/app.config.json`.
-</example>
+Use proper Markdown formatting. When referring to symbols (classes, methods, variables) in user's workspace wrap in backticks. For file paths and line number rules, see fileLinkification section
+<fileLinkification>
+When mentioning files or line numbers, always convert them to markdown links using workspace-relative paths and 1-based line numbers.
+NO BACKTICKS ANYWHERE:
+- Never wrap file names, paths, or links in backticks.
+- Never use inline-code formatting for any file reference.
+
+REQUIRED FORMATS:
+- File: [path/file.ts](path/file.ts)
+- Line: [file.ts](file.ts#L10)
+- Range: [file.ts](file.ts#L10-L12)
+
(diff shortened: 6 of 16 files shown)


src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,40 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { PromptElement } from '@vscode/prompt-tsx';
+import { Tag } from '../base/tag';
+
+export class FileLinkificationInstructions extends PromptElement<{}> {
+	render() {
+		return <Tag name='file_linkification'>
```
mjbvz: Just a gut feeling but this seems pretty verbose and I'm not sure how well models will follow the good/bad method written this way

src/extension/prompts/node/agent/openAIPrompts.tsx
```diff
@@ -313,6 +315,7 @@ class DefaultGpt5AgentPrompt extends PromptElement<DefaultAgentPromptProps> {
 				For casual greetings, acknowledgements, or other one-off conversational messages that are not delivering substantive information or structured results, respond naturally without section headers or bullet formatting.<br />
 				<br />
 				When referring to a filename or symbol in the user's workspace, wrap it in backticks.<br />
+				<FileLinkificationInstructions />
```
mjbvz: I don't think this should break up the example

src/extension/prompts/node/agent/openAIPrompts.tsx
mjbvz: We should revisit these instructions too as they duplicate the linkfication stuff. I'd expect to have some base set of linkfication rules and then a (hopefully) small set of additional customizations for each model

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,96 @@
...
+import { CancellationToken } from '../../../util/vs/base/common/cancellation';
+import { Location, Position, Range, Uri } from '../../../vscodeTypes';
+import { coalesceParts, LinkifiedPart, LinkifiedText, LinkifyLocationAnchor } from './linkifiedText';
+import { IContributedLinkifier, LinkifierContext } from './linkifyService';
+
+// Matches markdown links where the text is a path and optional #L anchor is present
+// Example: [src/file.ts](src/file.ts#L10-12) or [src/file.ts](src/file.ts)
+const modelLinkRe = /\[(?<text>[^\]\n]+)\]\((?<target>[^\s)]+)\)/gu;
+
+export class ModelFilePathLinkifier implements IContributedLinkifier {
```
mjbvz, line 19: Let's break up the existing `FilePathLinkifier` into the real links functionality and the inline code functionality. Maybe just delete the real links stuff from `FilePathLinkifier` and keep this class. I'd like to avoid the duplication though and make it so we just have one place that handles the markdown file links

vijayupadya: updated

src/extension/prompts/node/agent/openAIPrompts.tsx
vijayupadya: updated

src/extension/prompts/node/agent/openAIPrompts.tsx
vijayupadya: Fixed

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
vijayupadya: Fixed

src/extension/prompts/node/agent/anthropicPrompts.tsx
```diff
@@ -200,6 +202,7 @@ class Claude45DefaultPrompt extends PromptElement<DefaultAgentPromptProps> {
 					The function `calculateTotal` is defined in `lib/utils/math.ts`.<br />
 					You can find the configuration in `config/app.config.json`.
 				</Tag>
+				<FileLinkificationInstructions />
```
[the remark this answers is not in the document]
vijayupadya: expected

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,239 @@
...
+			const resolved = await this.resolveTarget(parsed.targetPath, workspaceFolders, parsed.preserveDirectorySlash);
+			if (!resolved) {
+				parts.push(original);
+				continue;
+			}
+
+			const basePath = getWorkspaceFileDisplayPath(this.workspaceService, resolved);
+			const anchorRange = this.parseAnchor(parsed.anchor);
+			if (parsed.anchor && !anchorRange) {
+				parts.push(original);
```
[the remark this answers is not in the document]
vijayupadya: fixed

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,31 @@
...
+			ALWAYS convert file paths to markdown links with 1-based line numbers:<br />
+			`[path/to/file.ts](path/to/file.ts)` - whole file<br />
+			`[path/to/file.ts](path/to/file.ts#L10)` - single line<br />
+			`[path/to/file.ts](path/to/file.ts#L10-12)` - line range<br />
```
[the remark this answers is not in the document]
vijayupadya: fixed

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,31 @@
...
+			Examples:<br />
+			❌ `The function is in exampleScript.ts at line 25.`<br />
+			✓ `The function is in [exampleScript.ts](exampleScript.ts#L25).`<br />
+			Critical rules:<br />
+			- Bracket text = exact file path only (no backticks, no `#L...`, no descriptive text like "line 549")<br />
```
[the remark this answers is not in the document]
vijayupadya: updated

src/extension/linkify/common/filePathLinkifier.ts
```diff
@@ -35,8 +32,8 @@ const pathMatchRe = new RegExp(
  * Linkifies file paths in responses. This includes:
  *
  * ```
- * [file.md](file.md)
  * `file.md`
+ * foo.ts
  * ```
  */
 export class FilePathLinkifier implements IContributedLinkifier {
```
mjbvz, line 39: Do we still need this class at all? Maybe just delete it and call  new `ModelFilePathLinkifier` `FilePathLinkifier` instead

It seems like the new linkifier should handle everything that the current one does. If there are any gaps we can fix those instead of keeping the fallback around

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,239 @@
...
+import { FileType } from '../../../platform/filesystem/common/fileTypes';
+import { getWorkspaceFileDisplayPath, IWorkspaceService } from '../../../platform/workspace/common/workspaceService';
+import { CancellationToken } from '../../../util/vs/base/common/cancellation';
+import { Location, Position, Range, Uri } from '../../../vscodeTypes';
+import { coalesceParts, LinkifiedPart, LinkifiedText, LinkifyLocationAnchor } from './linkifiedText';
+import { IContributedLinkifier, LinkifierContext } from './linkifyService';
+
+// Matches markdown links where the text is a path and optional #L anchor is present
+// Example: [src/file.ts](src/file.ts#L10-12) or [src/file.ts](src/file.ts)
+const modelLinkRe = /\[(?<text>[^\]\n]+)\]\((?<target>[^\s)]+)\)/gu;
```
mjbvz, line 17: Make sure this can support the case where the path text is code too as this is fairly common and models have likely been trained on a lot of it:

~~~
[`file.ts`](file.ts)
~~~

Especially true if we get rid of the old linkifier

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,239 @@
...
+				continue;
+			}
+
+			const workspaceFolders = this.workspaceService.getWorkspaceFolders();
+			if (!this.canLinkify(parsed, workspaceFolders)) {
+				parts.push(original);
+				continue;
+			}
+
+			const resolved = await this.resolveTarget(parsed.targetPath, workspaceFolders, parsed.preserveDirectorySlash);
```
mjbvz: Let's try to run these type of calls in parallel instead of awaiting each iteration

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,36 @@
...
+export class FileLinkificationInstructions extends PromptElement<{}> {
+	render() {
+		return <Tag name='file_linkification'>
+			ALWAYS convert file paths to markdown links with 1-based line numbers whenever you cite specific code locations. Use links inside normal sentences, not as the entire answer.<br />
+			Format examples:<br />
+			- `The handler lives in [path/to/file.ts](path/to/file.ts#L10).` (single line)<br />
+			- `See [path/to/file.ts](path/to/file.ts#L10-12) for the range.`<br />
+			- `Configuration is defined in [path/to/file.ts](path/to/file.ts).` (whole file)<br />
+			When you need a bullet list of references, write a short description before the link, for example:<br />
+			- Await chat view: [path/to/chatQuick.ts](path/to/chatQuick.ts#L142)<br />
```
mjbvz: Is this the syntax we want or should we support:

```
[Await chat view](path/to/chatQuick.ts#L142)
```

That feels more natural if we can get the model to generate these links properly

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,36 @@
...
+			- Show widget: [path/to/chatQuick.ts](path/to/chatQuick.ts#L321)<br />
+			Examples:<br />
+			❌ `The function is in exampleScript.ts at line 25.`<br />
+			✓ `The function is in [exampleScript.ts](exampleScript.ts#L25).`<br />
```
mjbvz: Is the `❌` how we provided counter examples to the model? Seems like it's mostly used in our docs, not in the actual prompts

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { PromptElement } from '@vscode/prompt-tsx';
+import { Tag } from '../base/tag';
+
+export class FileLinkificationInstructions extends PromptElement<{}> {
+	render() {
+		return <Tag name='file_linkification'>
+			ALWAYS convert file paths to markdown links with 1-based line numbers whenever you cite specific code locations. Use links inside normal sentences, not as the entire answer.<br />
+			Format examples:<br />
+			- `The handler lives in [path/to/file.ts](path/to/file.ts#L10).` (single line)<br />
```
roblourens: What's the advantage of this over a path in backticks? We're asking the model to repeat the path twice, it could be long.

roblourens: In every other path scenario, we ask it to use absolute paths. How do you handle relative paths within multiroot workspaces?

roblourens: And right now I think we support linkifying many different patterns, does this mean we would no longer support those patterns? (I strongly support simplifying the patterns we handle)

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,36 @@
...
+			- `See [path/to/file.ts](path/to/file.ts#L10-12) for the range.`<br />
+			- `Configuration is defined in [path/to/file.ts](path/to/file.ts).` (whole file)<br />
+			When you need a bullet list of references, write a short description before the link, for example:<br />
```
roblourens: Why?

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,36 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { PromptElement } from '@vscode/prompt-tsx';
+import { Tag } from '../base/tag';
+
+export class FileLinkificationInstructions extends PromptElement<{}> {
+	render() {
+		return <Tag name='file_linkification'>
```
roblourens: This is a lot, I just don't think all these rules and examples are necessary for a modern model. What you're asking for is extremely simple, I would expect that a decent model can pick it up with just a couple sentences.

We could do it in a scientific way and write some evals to test all models and how far we can trim the rules while still getting good results.

roblourens (dismissed):

src/extension/linkify/common/modelFilePathLinkifier.ts
vijayupadya: Yes, works. Added unit test to validate this.

src/extension/linkify/common/modelFilePathLinkifier.ts
vijayupadya: Valid point. Updated

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
vijayupadya: Agree, updated

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
vijayupadya: Updated

src/extension/linkify/common/filePathLinkifier.ts
vijayupadya: Right now ModelFilePathLinkifier handles structured markdown links with anchors [text](path#L10)text, while FilePathLinkifier handles inline code and plain text paths.
They're registered with clear separation of concerns, where FilePathLinkifier acts as a fallback for informal references.

That said, I'm happy to explore merging them if you think that would be better! Let me know your preference and I can make it work either way.

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
vijayupadya: Valid point, I started with simple few sentence prompts, but I got inconsistent results across models. Within the model also it was inconsistent between different files. I will try pruning it offline separately (with evals) and simplify without loosing the intent.

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
vijayupadya: model was generating responses like below, without any description on what those lines represent.

chatListRenderer.ts#857-857
chatListRenderer.ts#1001-1001
chatListRenderer.ts#1477-1477
chatListRenderer.ts#1486-1486

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
vijayupadya: For multi root workspace, currently the first match is selected.

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
roblourens: That's interesting. I wonder if all the rules about links caused it to hyper-focus on links specifically and forget what is was actually trying to communicate 😅

src/extension/linkify/common/filePathLinkifier.ts
mjbvz: My hope is that the  model will handle creating proper links now. The inline code and plain text path logic was added because previous models were no good at generating links consistently 

Worth testing removing the old linkifier to and seeing if the model still generates paths in plaintext /  inline code. If it does, maybe we need to tweak the instructions a bit

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,239 @@
...
+	private tryCreateFileUri(path: string): Uri | undefined {
+		try {
+			return Uri.file(path);
+		} catch {
+			return undefined;
+		}
+	}
+
+	private toVsUri(folder: Uri): Uri {
+		return Uri.parse(folder.toString());
```
mjbvz: The Uri -> string -> Uri flow seems strange. Why is it needed?

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,239 @@
...
+	private async resolveTarget(targetPath: string, workspaceFolders: readonly Uri[], preserveDirectorySlash: boolean): Promise<Uri | undefined> {
+		if (!workspaceFolders.length) {
+			return undefined;
+		}
+
+		const folderUris = workspaceFolders.map(folder => this.toVsUri(folder));
+
+		if (this.isAbsolutePath(targetPath)) {
+			const absoluteUri = this.tryCreateFileUri(targetPath);
+			if (!absoluteUri) {
```
mjbvz: This should look at the workspace folders to decide what type of path to create instead of always trying to create `file:` uris. Especially in remote cases, `file` may not be right

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,239 @@
...
+		} catch {
+			return undefined;
+		}
+	}
+
+	private toVsUri(folder: Uri): Uri {
+		return Uri.parse(folder.toString());
+	}
+
+	private isEqualOrParentFs(target: Uri, folder: Uri): boolean {
```
mjbvz: We should have a helper for this: `isEqualOrParent`?

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,239 @@
...
+		const targetFs = this.normalizeFsPath(target);
+		const folderFs = this.normalizeFsPath(folder);
+		return targetFs === folderFs || targetFs.startsWith(folderFs.endsWith('/') ? folderFs : `${folderFs}/`);
+	}
+
+	private normalizeFsPath(resource: Uri): string {
+		// Convert Windows backslashes to forward slashes and remove duplicate separators for stable comparisons.
+		return resource.fsPath.replace(/\\/g, '/').replace(/\/+/g, '/').toLowerCase();
```
mjbvz: Doing this manually also seems fishy. If you want paths with just `/`, you a use `.path` instead of `.fsPath`. Also you can use `normalizePath` to normalize uris

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,239 @@
...
+
+	private parseAnchor(anchor: string | undefined): { readonly range: Range; readonly startLine: string; readonly endLine: string | undefined } | undefined {
+		// Ensure the anchor follows the #L123, #L123-456, or #L123-L456 format before parsing it.
+		if (!anchor || !/^L\d+(?:-L?\d+)?$/.test(anchor)) {
+			return undefined;
+		}
+
+		// Capture the start (and optional end) line numbers from the anchor.
+		// Support both L123-456 and L123-L456 formats
+		const match = /^L(\d+)(?:-L?(\d+))?$/.exec(anchor);
```
mjbvz, line 216: I'd skip the `.test` above since it duplicates the regex and just try parsing directly

src/extension/linkify/common/modelFilePathLinkifier.ts
```diff
@@ -0,0 +1,239 @@
...
+	private canLinkify(parsed: { readonly text: string; readonly targetPath: string; readonly anchor: string | undefined }, workspaceFolders: readonly Uri[]): boolean {
+		const { text, targetPath, anchor } = parsed;
+		const textMatchesBase = targetPath === text;
+		const textIsFilename = !text.includes('/') && targetPath.endsWith(`/${text}`);
+		const descriptiveWithAnchor = !!anchor; // Allow any descriptive text when anchor is present
+
+		return Boolean(workspaceFolders.length) && (textMatchesBase || textIsFilename || descriptiveWithAnchor);
+	}
+
+	private async resolveTarget(targetPath: string, workspaceFolders: readonly Uri[], preserveDirectorySlash: boolean): Promise<Uri | undefined> {
```
mjbvz: Worth passing along the cancellation token in this case too and checking it in the calls to `tryState`

vijayupadya: updated

src/extension/linkify/common/modelFilePathLinkifier.ts
vijayupadya: updated

src/extension/linkify/common/modelFilePathLinkifier.ts
vijayupadya: fixed

src/extension/linkify/common/modelFilePathLinkifier.ts
vijayupadya: fixed

src/extension/linkify/common/modelFilePathLinkifier.ts
vijayupadya: fixed

src/extension/prompts/node/agent/anthropicPrompts.tsx
```diff
@@ -194,12 +195,12 @@ class Claude45DefaultPrompt extends PromptElement<DefaultAgentPromptProps> {
...
 			<NotebookInstructions {...this.props} />
 			<Tag name='outputFormatting'>
-				Use proper Markdown formatting in your answers. When referring to a filename or symbol in the user's workspace, wrap it in backticks.<br />
+				Use proper Markdown formatting. Backtick code identifiers (classes, methods, variables). For file paths and code locations, follow `FileLinkificationInstructions` (markdown links with line anchors; never backticks).<br />
 				<Tag name='example'>
-					The class `Person` is in `src/models/person.ts`.<br />
-					The function `calculateTotal` is defined in `lib/utils/math.ts`.<br />
-					You can find the configuration in `config/app.config.json`.
+					Identifiers in backticks: `Person`, `calculateTotal`, `AppConfig`.<br />
+					File path and line anchor formatting rules are defined in `FileLinkificationInstructions` below.
```
bhavyaus: You can remove this line and just inlcude the examples inside the FileLinkificationInstructions?

src/extension/prompts/node/agent/anthropicPrompts.tsx
```diff
@@ -194,12 +195,12 @@ class Claude45DefaultPrompt extends PromptElement<DefaultAgentPromptProps> {
 			{this.props.availableTools && <McpToolInstructions tools={this.props.availableTools} />}
 			<NotebookInstructions {...this.props} />
 			<Tag name='outputFormatting'>
-				Use proper Markdown formatting in your answers. When referring to a filename or symbol in the user's workspace, wrap it in backticks.<br />
+				Use proper Markdown formatting. Backtick code identifiers (classes, methods, variables). For file paths and code locations, follow `FileLinkificationInstructions` (markdown links with line anchors; never backticks).<br />
```
bhavyaus: The previous phrasing, “wrap in backticks,” was more proper than using “backtick” as a verb here and might be less confusing.

bhavyaus: Also do we have anyway to test this besides manual testing? Might be good to update our benchmarks if not.

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,35 @@
...
+			Format examples:<br />
+			- `The handler lives in [path/to/file.ts](path/to/file.ts#L10).` (single line)<br />
+			- `See [path/to/file.ts](path/to/file.ts#L10-L12) for the range.`<br />
+			- `Configuration is defined in [path/to/file.ts](path/to/file.ts).` (whole file)<br />
+			- `The widget renderer attaches anchors ([src/renderer.ts](src/renderer.ts#L42-L48)).` (in parentheses)<br />
+			When you need a bullet list of references with line numbers, you can use descriptive text:<br />
+			- [Await chat view](path/to/chatQuick.ts#L142)<br />
+			- [Show widget](path/to/chatQuick.ts#L321)<br />
+			NEVER cite file paths as plain text when referring to specific locations. For example, instead of saying `The function is in exampleScript.ts at line 25.`, say `The function is in [exampleScript.ts](exampleScript.ts#L25).`<br />
+			Critical rules:<br />
```
bhavyaus: I don't think this makes much of an impact with most models if they are not included as reminders closer to the user message

src/extension/prompts/node/agent/anthropicPrompts.tsx
vijayupadya: updated

src/extension/prompts/node/agent/anthropicPrompts.tsx
vijayupadya: updated

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
vijayupadya: In my testing, I have found this prompt to make a difference. Without this, line # always appear separate from the file path reference, for e.g. `line 25 in file.ts `or `file.ts at line 25.`

src/extension/linkify/common/modelFilePathLinkifier.ts
vijayupadya: updated

src/extension/linkify/common/filePathLinkifier.ts
mjbvz: Still interested to see if we can avoid having two linkifier paths and just prefer using the model

src/extension/prompts/node/agent/openai/gpt51CodexPrompt.tsx
```diff
@@ -69,7 +70,8 @@ class Gpt51CodexPrompt extends PromptElement<DefaultAgentPromptProps> {
 				- Markdown text. Use structure only when it helps scanability.<br />
 				- Headers: optional; short Title Case (1-3 words) wrapped in **…**; no blank line before the first bullet; add only if they truly help.<br />
 				- Bullets: use - ; merge related points; keep to one line when possible; 4-6 per list ordered by importance; keep phrasing consistent.<br />
-				- Monospace: backticks for commands/paths/env vars/code ids and inline examples; use for literal keyword bullets; never combine with **.<br />
+				- Monospace: backticks for commands, env vars, and code identifiers; never combine with **.<br />
+				- File path + line anchor link rules are defined in `FileLinkificationInstructions` below.<br />
```
mjbvz: Does the model actually see the name `FileLinkificationInstructions`? Should we just include the instructions here?

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
```diff
@@ -74,12 +74,34 @@ Important Reminder: Avoid referencing Notebook Cell Ids in user messages. Use ce
...
-Use proper Markdown formatting in your answers. When referring to a filename or symbol in the user's workspace, wrap it in backticks.
+Use proper Markdown formatting. When referring to symbols (classes, methods, variables) in user's workspace wrap in backticks. For file paths and code locations, follow `FileLinkificationInstructions` (markdown links with line anchors; never backticks).
 <example>
-The class `Person` is in `src/models/person.ts`.
-The function `calculateTotal` is defined in `lib/utils/math.ts`.
-You can find the configuration in `config/app.config.json`.
+File path and line anchor formatting rules are defined in `FileLinkificationInstructions` below.
 </example>
+<file_linkification>
+ALWAYS convert file paths to markdown links with 1-based line numbers whenever you cite specific code locations. Use links inside normal sentences, not as the entire answer.
```
mjbvz: Not sure if it would help models or not but may be worth seeing if we should clarify that this is meant for workspace files, not just any arbitrary file path (such as files in examples)

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
```diff
@@ -74,12 +74,34 @@ Important Reminder: Avoid referencing Notebook Cell Ids in user messages. Use ce
...
+- `The handler lives in [path/to/file.ts](path/to/file.ts#L10).` (single line)
+- `See [path/to/file.ts](path/to/file.ts#L10-L12) for the range.`
+- `Configuration is defined in [path/to/file.ts](path/to/file.ts).` (whole file)
+- `The widget renderer attaches anchors ([src/renderer.ts](src/renderer.ts#L42-L48)).` (in parentheses)
+When you need a bullet list of references with line numbers, you can use descriptive text:
+- [Await chat view](path/to/chatQuick.ts#L142)
+- [Show widget](path/to/chatQuick.ts#L321)
+NEVER cite file paths as plain text when referring to specific locations. For example, instead of saying `The function is in exampleScript.ts at line 25.`, say `The function is in [exampleScript.ts](exampleScript.ts#L25).`
+Critical rules:
+- Link text must be the exact file path (no backticks, no `#L` in the visible text, no extra wording). Keep the `#L` anchor in the **link target**, e.g. `[src/file.ts](src/file.ts#L25)`.
```
mjbvz: This instructions seems to go against the example above: `[Await chat view](path/to/chatQuick.ts#L142)`

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
```diff
@@ -74,12 +74,34 @@ Important Reminder: Avoid referencing Notebook Cell Ids in user messages. Use ce
...
+- Always include both brackets **and** parentheses. `[src/file.ts](src/file.ts#L25)` is valid; `[src/file.ts#L25]` is not.
```
mjbvz: Worth seeing if this is needed. If it is, maybe say something like "make sure it's a valid markdown link" instead

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
```diff
@@ -74,12 +74,34 @@ Important Reminder: Avoid referencing Notebook Cell Ids in user messages. Use ce
...
+- Path format: Strip drive letters and workspace parent folders - use only path after workspace root
+- Transform `c:/Repos/workspace/src/file.ts` → `[src/file.ts](src/file.ts)`
+- Always use forward slashes `/`, never backslashes `/`
+- Do not use URIs like file://, vscode://, or https://.
```
mjbvz: The `https` part makes me worried it won't generate web links when it should

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
mjbvz: Also worth calling out the paths should be relative to the workspace

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
```diff
@@ -74,12 +74,34 @@ Important Reminder: Avoid referencing Notebook Cell Ids in user messages. Use ce
...
-You can find the configuration in `config/app.config.json`.
+File path and line anchor formatting rules are defined in `FileLinkificationInstructions` below.
 </example>
+<file_linkification>
+ALWAYS convert file paths to markdown links with 1-based line numbers whenever you cite specific code locations. Use links inside normal sentences, not as the entire answer.
+Format examples:
+- `The handler lives in [path/to/file.ts](path/to/file.ts#L10).` (single line)
+- `See [path/to/file.ts](path/to/file.ts#L10-L12) for the range.`
+- `Configuration is defined in [path/to/file.ts](path/to/file.ts).` (whole file)
+- `The widget renderer attaches anchors ([src/renderer.ts](src/renderer.ts#L42-L48)).` (in parentheses)
```
mjbvz: Why is the parens case special?

vijayupadya: Not needed, removed.

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
vijayupadya: Removed https

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
vijayupadya: This is needed. I simplified text a little.

Common mistakes I've seen models make:

[src/file.ts#L25] - Missing parentheses (not a link at all, just literal text)
[src/file.ts] - Missing target completely
[[src/file.ts]](src/file.ts) - Double brackets
[src/file.ts](src/file.ts)#L25 - Anchor outside the parentheses

Also, models misplace the line anchor, without explicit instruction, models often do:

[src/file.ts#L25](src/file.ts) - Line number in display text 
[src/file.ts](src/file.ts)#L25 - Anchor after closing paren 
[src/file.ts L25](src/file.ts#L25) - Extra text in display

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
vijayupadya: Simplified text

src/extension/prompts/node/agent/test/__snapshots__/agentPrompts-all_non_edit_tools-claude-sonnet-4.5.spec.snap
vijayupadya: Updated

src/extension/prompts/node/agent/openai/gpt51CodexPrompt.tsx
vijayupadya: I updated it with the 'fileLinkification' tag, which it is able to find.

src/extension/prompts/node/agent/openai/defaultOpenAIPrompt.tsx
```diff
@@ -99,15 +100,15 @@ export class DefaultOpenAIAgentPrompt extends PromptElement<DefaultAgentPromptPr
 			{this.props.availableTools && <McpToolInstructions tools={this.props.availableTools} />}
 			<NotebookInstructions {...this.props} />
 			<Tag name='outputFormatting'>
-				Use proper Markdown formatting in your answers. When referring to a filename or symbol in the user's workspace, wrap it in backticks.<br />
+				Use proper Markdown formatting. Backtick code identifiers (classes, functions, variables, commands). For file paths or specific code locations, do NOT use backticks—convert them to markdown links with line anchors; file path and line anchor link rules are defined in `FileLinkificationInstructions` below. Avoid duplicating those examples here.<br />
 				<Tag name='example'>
```
mjbvz: Should the `FileLinkificationInstructions` be included in the output formatting section similar to `MathIntegrationRules`? Wondering we can avoid the `File path link examples live in fileLinkification section below.` section

src/extension/prompts/node/agent/openai/gpt5CodexPrompt.tsx
```diff
@@ -69,18 +70,14 @@ class CodexStyleGpt5CodexPrompt extends PromptElement<DefaultAgentPromptProps> {
...
 			- Tone: collaborative, concise, factual; present tense, active voice; self-contained; no "above/below"; parallel wording.<br />
 			- Don'ts: no nested bullets/hierarchies; no ANSI codes; don't cram unrelated keywords; keep keyword lists short—wrap/reformat if long; avoid naming formatting styles in answers.<br />
 			- Adaptation: code explanations → precise, structured with code refs; simple tasks → lead with outcome; big changes → logical walkthrough + rationale + next actions; casual one-offs → plain sentences, no headers/bullets.<br />
-			- File References: When referencing files in your response, always follow the below rules:<br />
-			* Use inline code to make file paths clickable.<br />
-			* Each reference should have a stand alone path. Even if it's the same file.<br />
-			* Accepted: absolute, workspace-relative, a/ or b/ diff prefixes, or bare filename/suffix.<br />
-			* Do not use URIs like file://, vscode://, or https://.<br />
-			* Examples: src/app.ts, C:\repo\project\main.rs<br />
+			- File path and line anchor link rules are defined in fileLinkification section below.<br />
```
mjbvz: I don't think we need this sentence if we include the rules right after?

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { PromptElement } from '@vscode/prompt-tsx';
+import { Tag } from '../base/tag';
+
+export class FileLinkificationInstructions extends PromptElement<{}> {
+	render() {
+		return <Tag name='fileLinkification'>
+			ALWAYS convert file paths to markdown links with 1-based line numbers whenever you cite specific code locations in the workspace. Paths should be relative to workspace root.<br />
+			<br />
+			**Inline references:** Use the file path as link text within sentences:<br />
```
roblourens: When I tested this with 5.1-codex, I got a markdown link in backticks:

```
`[src/vs/workbench/contrib/chat/common/chatService.ts](src/vs/workbench/contrib/chat/common/chatService.ts)` is the central TypeScript definition file for the chat feature’s shared contracts
```

I think you aren't looking for the markdown link to be wrapped in backticks, but it's not clear from the examples because they are wrapped in backticks, so I would suggest removing those.

roblourens: And then I pulled with the other commit you added, and I got a path that wasn't formatted as a link at all

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
roblourens: So that means that it's not possible for the model to precisely refer to a certain file like package.json which would exist at the same place in both repos? I also worry that it might try to dedupe by including the name of the workspace, which is something I've seen it try to do with search tools, even though you don't ask for that.

roblourens (commented):
I see the anchor rendered as this 
Image

when the model returned 

```
The base implementation is in [SearchService](src/vs/workbench/services/search/common/searchService.ts#L28) which:
```

This was sonnet 4.5

src/extension/prompts/node/agent/fileLinkificationInstructions.tsx
vijayupadya: I made few changes and made prompts consistent across models. In my local testing, the above scenario worked.

vijayupadya:
I have made the changes so now multi workspace paths show up correctly.
It now generates responses like below:

Both [app1/package.json](app1/package.json#L12) and [app2/package.json](app2/package.json#L12) are adding the dependency for react version ^18.2.0.

src/extension/prompts/node/agent/openai/defaultOpenAIPrompt.tsx
vijayupadya: updated

src/extension/prompts/node/agent/openai/gpt5CodexPrompt.tsx
vijayupadya: These are added for additional re-enforcement as I see without this, models are not consistent in honoring the rules.

mjbvz:
Looks like a bunch of conflicts crept in. Trying to resolve now and then will need re-review

PR #1809, opened 2025-11-05
Clear env variable contributions when loading chat ext

DonJayamanne:
Fixes https://github.com/microsoft/vscode/issues/275692

### DIFF
--- a/src/platform/terminal/vscode/terminalServiceImpl.ts
+++ b/src/platform/terminal/vscode/terminalServiceImpl.ts
@@ -22,9 +22,12 @@ export class TerminalServiceImpl extends Disposable implements ITerminalService 
 		@IVSCodeExtensionContext private readonly context: IVSCodeExtensionContext,
 	) {
 		super();
-		// As env variables are persisted, clear them on startup
-		// This way any old contributions from previous versions/sessions are cleared out.
-		this.context.environmentVariableCollection.clear();
+		// This used to be setup in the past for Copilot CLI auth in terminals.
+		// It was only ever shipped in the VSCode insiders and never got into stable.
+		// So this is only required for users who had insiders installed before it was removed.
+		// Safe to remove this after a few months or so (https://github.com/microsoft/vscode/issues/275692).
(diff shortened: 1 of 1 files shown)


src/platform/terminal/vscode/terminalServiceImpl.ts
```diff
@@ -22,6 +22,9 @@ export class TerminalServiceImpl extends Disposable implements ITerminalService
 		@IVSCodeExtensionContext private readonly context: IVSCodeExtensionContext,
 	) {
 		super();
+		// As env variables are persisted, clear them on startup
+		// This way any old contributions from previous versions/sessions are cleared out.
+		this.context.environmentVariableCollection.clear();
```
Tyriar: Is that going to break the copilot-debug one?

DonJayamanne: No, that gets updated when extension activates.
So those env variables are always setup when ext loads
verified by looking at the terminal env contributions

DonJayamanne: We only have two contributions, copilot cli and debug PATH contributions
And both are setup in activation phase.

Tyriar: If this runs after that, it'll clear that out. The right way to delete ones that used to be there permanently is to call delete on the variable name and keep it there.

Clearing the whole collection is only good if you're setting up everything immediately after (not async),otherwise you could get flickering of warning, could be races, etc.

Tyriar: A diff of the environment variable mutators is calculated and applies to the terminal's on next microtask I believe.

DonJayamanne: > If this runs after that, it'll clear that out.

This is the constructor, so this will always run first 100% of the time, the changes are made in this class via instance methods.

> The right way to delete ones that used to be there permanently is to call delete on the variable name and keep it there.

Will do this.

PR #1817, opened 2025-11-06
permissive gh auth

joshspicer:
closes https://github.com/microsoft/vscode/issues/275641


TylerLeonhardt (dismissed):
This is fine for this release but we should really revisit the UX around this step up

mjbvz (approved):

PR #1868, opened 2025-11-08
Support includeIgnoredFiles option in search results

bryanchen-d:
Fixes https://github.com/microsoft/vscode/issues/270857

Introduce a new option to include files that are normally ignored according to .gitignore and other exclusion settings in search results.

### DIFF
--- a/src/extension/tools/node/findTextInFilesTool.tsx
+++ b/src/extension/tools/node/findTextInFilesTool.tsx
@@ -98,9 +98,9 @@ export class FindTextInFilesTool implements ICopilotTool<IFindTextInFilesToolPar
 				}
 			}
 
-			noMatchInstructions = `Your search pattern might be excluded completedly by either the search.exclude settings or .*ignore files.
-			If you believe that it should have results, you can check into the .*ignore files and the exclude setting (here are some excluded patterns for reference:[${excludePaths.join(';')}] separated by ';').
-			Then if you want to include those files you can call the tool again by setting "includeIgnoredFiles" to true.`;
+			noMatchInstructions = `Your search pattern might be excluded completely by either the search.exclude settings or .*ignore files.
+If you believe that it should have results, you can check into the .*ignore files and the exclude setting (here are some excluded patterns for reference:[${excludePaths.join(',')}]).
+Then if you want to include those files you can call the tool again by setting "includeIgnoredFiles" to true.`;
 		}
 
 		const prompt = await renderPromptElementJSON(this.instantiationService,


src/extension/tools/node/findTextInFilesTool.tsx
```diff
@@ -126,16 +128,20 @@ export class FindTextInFilesTool implements ICopilotTool<IFindTextInFilesToolPar
 		}
 	}
 
-	private async searchAndCollectResults(query: string, isRegExp: boolean, patterns: vscode.GlobPattern[] | undefined, maxResults: number, token: CancellationToken): Promise<vscode.TextSearchResult2[]> {
+	private async searchAndCollectResults(query: string, isRegExp: boolean, patterns: vscode.GlobPattern[] | undefined, maxResults: number, includeIgnoredFiles: boolean | undefined, token: CancellationToken): Promise<vscode.TextSearchResult2[]> {
+		const findOptions = {
```
roblourens: If you write this as `const findOptions: vscode.FindTextInFilesOptions2` instead of casting, then you will get an error if the object includes a key that isn't a correct prop in that object.

package.json
```diff
@@ -290,6 +290,10 @@
 						"maxResults": {
 							"type": "number",
 							"description": "The maximum number of results to return. Do not use this unless necessary, it can slow things down. By default, only some matches are returned. If you use this and don't see what you're looking for, you can try again with a more specific query or a larger maxResults."
+						},
+						"includeIgnoredFiles": {
+							"type": "boolean",
+							"description": "Whether to include files that would normally be ignored according to .gitignore, other ignore files and `files.exclude` and `search.exclude` settings."
```
roblourens: Include a message like `Warning: using this may cause the search to be slower. Only set it when you want to search in ignored folders like node_modules or build outputs.`

roblourens: Does this work? Did you see it used correctly in some scenarios?

bryanchen-d: Hmm, sometimes it did pass in `includeIgnoredFiles = true` but today quite often it doesn't pass in like that. Let me try some tweaks.

src/extension/tools/node/findTextInFilesTool.tsx
bryanchen-d: You are right. Updated.

package.json
roblourens: This would be a great one to have a vscbench eval for

bryanchen-d: I updated it adding instructions about the option to include ignored files in the search. Also I tested it with a new eval for the scenario. That new eval is in PR: https://github.com/microsoft/vscode-copilot-evaluation/pull/873

src/extension/tools/node/findTextInFilesTool.tsx
```diff
@@ -72,17 +77,35 @@ export class FindTextInFilesTool implements ICopilotTool<IFindTextInFilesToolPar
...
+			const excludePaths: string[] = [];
+			if (excludeSettings) {
+				for (const [path, isExcluded] of Object.entries(excludeSettings)) {
+					if (isExcluded) {
+						excludePaths.push(path);
+					}
+				}
+			}
+
+			noMatchInstructions = `Your search pattern might be excluded completedly by either the search.exclude settings or .*ignore files.
```
roblourens:
```suggestion
			noMatchInstructions = `Your search pattern might be excluded completely by either the search.exclude settings or .*ignore files.
```

src/extension/tools/node/findTextInFilesTool.tsx
```diff
@@ -72,17 +77,35 @@ export class FindTextInFilesTool implements ICopilotTool<IFindTextInFilesToolPar
...
+			If you believe that it should have results, you can check into the .*ignore files and the exclude setting (here are some excluded patterns for reference:[${excludePaths.join(';')}] separated by ';').
```
roblourens: Does this seem necessary, or can it figure out how to fall back to use this parameter without the extra hint?

roblourens: If so, how about just a comma-separated list? Feels like you shouldn't have to explain how the list is formatted

roblourens: These deep indents will be in the string

bryanchen-d: The extra hint will make it more stable from my tests. Let me try shorten it by using comma.

src/extension/tools/node/findTextInFilesTool.tsx
bryanchen-d: updated.

src/extension/tools/node/findTextInFilesTool.tsx
bryanchen-d: Updated and validated with eval.

roblourens (approved):

PR #1874, opened 2025-11-08
Copilot CLI - publish statistics for session using git worktree

lszomoru:


src/extension/agents/copilotcli/node/copilotcliSession.ts
```diff
@@ -169,6 +171,12 @@ export class CopilotCLISession extends DisposableStore implements ICopilotCLISes
 			await this._sdkSession.send({ prompt, attachments, abortController });
 			this.logService.trace(`[CopilotCLISession] Invoking session (completed) ${this.sessionId}`);
 
+			if (this._options.workingDirectory) {
```
DonJayamanne: just because we have workingDirectory doesn't mean its in a git worktree, nor does it mean its git SCM, also feels wrong to stage changes just because we run something in chat.

Is this meant to be done only for git worktrees?
If, so, we might need some other way to determine whether this session is dealing with git work trees

src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
```diff
@@ -154,36 +155,50 @@ export class CopilotCLIChatSessionItemProvider extends Disposable implements vsc
 
 	public async provideChatSessionItems(token: vscode.CancellationToken): Promise<vscode.ChatSessionItem[]> {
 		const sessions = await this.copilotcliSessionService.getAllSessions(token);
-		const diskSessions = sessions.map(session => this._toChatSessionItem(session));
+		const diskSessions: vscode.ChatSessionItem[] = [];
+		for (const session of sessions) {
```
DonJayamanne: Promise.all(... would be be, as we'd be processing & awaiting on all promises in parallel as opposed to processing each session one at at time sequentially.

DonJayamanne (changes requested):
Question about work tree & Promise.all

src/extension/agents/copilotcli/node/copilotcliSession.ts
lszomoru: Great point! Adjusted the code. Please take a look.

DonJayamanne (dismissed):

PR #1876, opened 2025-11-08
Fix subagent request marked as user-initiated

roblourens:
microsoft/vscode#276305 in main


sr-tream:
Why its managed on client-side? Its potential abuse-point

PR #1923, opened 2025-11-11
update problem matcher for use with tsgo

meganrogge:
Before:

Screenshot 2025-11-11 at 11 21 01 AM

After:

Screenshot 2025-11-11 at 11 19 15 AM


meganrogge:
Will update here instead https://github.com/microsoft/vscode/blob/68ea87416d88addc9a09bfd3186fed1a3c9a2003/extensions/typescript-language-features/package.json#L1788

PR #1925, opened 2025-11-11
Search Subagent Tool

24anisha:
Added a search subagent tool that will eventually be used to call the fine-tuned swe-grep style model


src/extension/tools/node/searchSubagentTool.ts
```diff
@@ -0,0 +1,218 @@
...
+import { Conversation, Turn } from '../../prompt/common/conversation';
+import { IBuildPromptContext } from '../../prompt/common/intents';
+import { SubagentToolCallingLoop } from '../../prompt/node/subagentLoop';
+import { SearchSubagentPrompt } from '../../prompts/node/agent/searchSubagentPrompt';
+import { PromptElementCtor } from '../../prompts/node/base/promptElement';
+import { ToolName } from '../common/toolNames';
+import { CopilotToolMode, ICopilotTool, ToolRegistry } from '../common/toolsRegistry';
+
+// Local function to render PromptTsx parts to strings (simplified to avoid vscode-node import restrictions)
+function renderToolResultToStringNoBudget(part: LanguageModelPromptTsxPart): string {
```
24anisha: I had to create this function because I couldn't import it from vscode-node into node. Is this an alright way to handle it?

src/extension/tools/node/searchSubagentTool.ts
```diff
@@ -0,0 +1,218 @@
...
+					description: options.input.description,
+					time: new Date().toISOString(),
+					responseType: loopResult.response.type,
+					success: loopResult.response.type === ChatFetchResponseType.Success
+				},
+				conversation: rounds,
+				finalResponse: loopResult.round.response
+			};
+
+			fs.writeFileSync(trajectoryFile, JSON.stringify(trajectory, null, 2), 'utf-8');
```
24anisha: this code section works when I'm locally debugging vscode-copilot-chat, and writes a file in the correct format to my machine at the home directory. however, the file doesn't show up if I run an msbench benchmark. @karthiknadig you mentioned to yufan that files written out should appear in the results. how do we resolve this?

src/extension/tools/node/searchSubagentTool.ts
zhichli: No, duplicating isn't ideal. The issue is VS Code's layering rule: `vscode-node/` is more specific than `node/`, so `node/` cannot import from it (imports only flow downward: `vscode-node` → `node` → `common`).

Maybe move `requestLoggerToolResult.tsx` into `src/extension/prompt/node/` since it has no VS Code runtime dependencies. Then both folders can import it properly without duplication.

src/extension/tools/node/searchSubagentTool.ts
zhichli: Regarding subagent tool call capturing, I’ve created a [POC commit](https://github.com/microsoft/vscode-copilot-chat/commit/3b2a339e85b0aa43b0cac5f0a4cdcd39a19cad4e) on top of [Anisha's branch](https://github.com/24anisha/vscode-copilot-chat/commits/anisha/search_subagent/) to show the recommended approach for capturing subagent tool calls in the exported log, which our eval relies on.

PR #1953, opened 2025-11-12
get rid of task as any

meganrogge:


vscode.proposed.taskExecutionTerminal.d.ts
```diff
@@ -0,0 +1,15 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// #234440
+declare module 'vscode' {
+
+	export interface TaskExecution {
```
Tyriar: I don't see any other proposed .d.ts' in the root directory, this needs to move somewhere?

Tyriar (dismissed):

vscode.proposed.taskExecutionTerminal.d.ts
meganrogge: I included that to show that even when it's provided in the repo, that property is still not recognized. I don't think there's a way to get rid of `as any` here

src/platform/tasks/vscode/tasksService.ts
```diff
@@ -32,7 +32,7 @@ export class TasksService extends DisposableStore implements ITasksService {
 	) {
 		super();
 		this.add(vscode.tasks.onDidStartTask(e => {
-			const terminal: vscode.Terminal | undefined = e.execution.terminal;
+			const terminal: vscode.Terminal | undefined = (e.execution as unknown as { terminal?: vscode.Terminal }).terminal;
```
Tyriar, line 35: Better, but still not great 😅 

Proposed .d.ts look like they live inside `src/extension/`, was that the problem?

PR #1960, opened 2025-11-12
Remove explicit 'any'

bryanchen-d:
Part of https://github.com/microsoft/vscode/issues/276879

### DIFF
--- a/src/extension/tools/common/toolsRegistry.ts
+++ b/src/extension/tools/common/toolsRegistry.ts
@@ -67,35 +67,33 @@ export interface ICopilotToolExtension<T> {
 export interface ICopilotTool<T> extends vscode.LanguageModelTool<T>, ICopilotToolExtension<T> {
 }
 
-export type CopilotToolParamsBase = {} | void;
-
-export interface ICopilotToolCtor<T = CopilotToolParamsBase> {
+export interface ICopilotToolCtor {
 	readonly toolName: ToolName;
-	new(...args: never[]): ICopilotTool<T>;
+	new(...args: never[]): ICopilotTool<unknown>;
 }
 
-export interface ICopilotToolExtensionCtor<T = CopilotToolParamsBase> {
+export interface ICopilotToolExtensionCtor {
 	readonly toolName: ToolName;
-	new(...args: never[]): ICopilotToolExtension<T>;
+	new(...args: never[]): ICopilotToolExtension<unknown>;
 }
 
 export const ToolRegistry = new class {
-	private _tools: Array<ICopilotToolCtor<CopilotToolParamsBase>> = [];
-	private _toolExtensions: Array<ICopilotToolExtensionCtor<CopilotToolParamsBase>> = [];
+	private _tools: Array<ICopilotToolCtor> = [];
+	private _toolExtensions: Array<ICopilotToolExtensionCtor> = [];
 
-	public registerTool<T extends CopilotToolParamsBase>(tool: ICopilotToolCtor<T>) {
+	public registerTool(tool: ICopilotToolCtor) {
 		this._tools.push(tool);
 	}
 
-	public getTools(): readonly ICopilotToolCtor<CopilotToolParamsBase>[] {
+	public getTools(): readonly ICopilotToolCtor[] {
 		return this._tools;
 	}
 
-	public registerToolExtension<T extends CopilotToolParamsBase>(tool: ICopilotToolExtensionCtor<T>) {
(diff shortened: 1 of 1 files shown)


src/extension/tools/common/toolsRegistry.ts
```diff
@@ -67,34 +67,35 @@ export interface ICopilotToolExtension<T> {
...
 }
 
 export const ToolRegistry = new class {
-	private _tools: ICopilotToolCtor[] = [];
-	private _toolExtensions: ICopilotToolExtensionCtor[] = [];
+	private _tools: Array<ICopilotToolCtor<CopilotToolParamsBase>> = [];
+	private _toolExtensions: Array<ICopilotToolExtensionCtor<CopilotToolParamsBase>> = [];
 
-	public registerTool(tool: ICopilotToolCtor) {
+	public registerTool<T extends CopilotToolParamsBase>(tool: { readonly toolName: ToolName; new(...args: any[]): ICopilotTool<T> }) {
```
roblourens: This adds an `any` :)

src/extension/tools/common/toolsRegistry.ts
```diff
@@ -67,34 +67,35 @@ export interface ICopilotToolExtension<T> {
...
 		this._tools.push(tool);
 	}
 
-	public getTools(): readonly ICopilotToolCtor[] {
+	public getTools(): readonly ICopilotToolCtor<CopilotToolParamsBase>[] {
 		return this._tools;
 	}
 
-	public registerToolExtension(tool: ICopilotToolExtensionCtor) {
+	public registerToolExtension<T extends CopilotToolParamsBase>(tool: { readonly toolName: ToolName; new(...args: any[]): ICopilotToolExtension<T> }) {
```
roblourens: `any`

roblourens: Why inline this type instead of using the `ICopilotToolExtensionCtor` interface?

src/extension/tools/common/toolsRegistry.ts
```diff
@@ -67,34 +67,35 @@ export interface ICopilotToolExtension<T> {
 export interface ICopilotTool<T> extends vscode.LanguageModelTool<T>, ICopilotToolExtension<T> {
 }
 
+export type CopilotToolParamsBase = {} | void;
```
roblourens: This is the type of the tool args? Should it just be `unknown`?

src/extension/tools/common/toolsRegistry.ts
bryanchen-d: updated.

src/extension/tools/common/toolsRegistry.ts
bryanchen-d: Oops, I overlooked that. Thanks!

src/extension/tools/common/toolsRegistry.ts
bryanchen-d: I see currently all tools' args are either a object or void for now, unless we want to support more types?

roblourens: I'm just wondering, do we actually get something by having a type for the tool args at this level? Since we generally just invoke tools with 'unknown' args anyway

bryanchen-d: I am fine with `unknown` though, let me try that.

bryanchen-d: updated using unknown instead.

src/extension/tools/common/toolsRegistry.ts
```diff
@@ -67,20 +67,19 @@ export interface ICopilotToolExtension<T> {
 export interface ICopilotTool<T> extends vscode.LanguageModelTool<T>, ICopilotToolExtension<T> {
 }
 
-
 export interface ICopilotToolCtor {
 	readonly toolName: ToolName;
-	new(...args: any[]): ICopilotTool<any>;
+	new(...args: never[]): ICopilotTool<unknown>;
```
roblourens, line 72: Why 'never'? I also ran into an issue trying to fix constructor arg types recently, did `unknown` not work?

bryanchen-d: `unknown` does not work. The typecheck failed at  calling `ToolRegistry.registerTool()`:
```
Argument of type 'typeof InstallExtensionTool' is not assignable to parameter of type 'ICopilotToolCtor'.
  Types of parameters '_commandService' and 'args' are incompatible.
    Type 'unknown' is not assignable to type 'IRunCommandExecutionService'.ts(2345)
   ```

PR #1974, opened 2025-11-13
Adding new vsc hidden models  

AbdelrahmanAbouelenin:
## Add Family-Based Custom Prompt Selection

Adds a three-layer matching system for custom agent prompts, enabling prompt selection based on model ID, hidden family, finally family name prefix (fallback).

### Priority Order
* Model ID hash match (highest priority - hidden models)
* Family name hash match (hidden family names)
* Family prefix string match (public family names - existing behavior)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2213,6 +2213,11 @@
 				"icon": "$(export)"
 			},
 			{
+				"command": "github.copilot.chat.showAsChatSession",
+				"title": "Show as chat session",
+				"icon": "$(chat-sparkle)"
+			},
+			{
 				"command": "github.copilot.debug.collectWorkspaceIndexDiagnostics",
 				"title": "%github.copilot.command.collectWorkspaceIndexDiagnostics%",
 				"category": "Developer"
@@ -3368,9 +3373,21 @@
 					"when": "resourceLangId == chatReplay"
 				},
 				{
+					"command": "github.copilot.chat.showAsChatSession",
+					"group": "navigation@9",
+					"when": "resourceLangId == chatReplay"
+				},
+				{
 					"command": "github.copilot.chat.applyCopilotCLIAgentSessionChanges",
 					"group": "navigation@1",
 					"when": "resourceScheme == copilotcli-worktree-changes"
+				}
+			],
+			"explorer/context": [
+				{
+					"command": "github.copilot.chat.showAsChatSession",
+					"when": "resourceFilename === 'benchRun.chatReplay.json' || resourceFilename === 'chat-export-logs.json'",
+					"group": "2_copilot@1"
 				}
 			],
 			"editor/context": [
@@ -3631,6 +3648,10 @@
 				},
 				{
 					"command": "github.copilot.chat.applyCopilotCLIAgentSessionChanges",
+					"when": "false"
+				},
+				{
+					"command": "github.copilot.chat.showAsChatSession",
 					"when": "false"
 				}
 			],
--- a/src/extension/agents/copilotcli/common/copilotCLITools.ts
+++ b/src/extension/agents/copilotcli/common/copilotCLITools.ts
@@ -27,7 +27,7 @@ interface ViewTool {
 }
 
 interface EditTool {
-	toolName: 'edit';
+	toolName: 'edit' | 'str_replace';
 	arguments: {
 		path: string;
 		old_str?: string;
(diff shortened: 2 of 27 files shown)


src/extension/prompts/node/agent/agentPrompt.tsx
```diff
@@ -321,7 +321,8 @@ export class AgentUserMessage extends PromptElement<AgentUserMessageProps> {
 			this.logService.trace('Re-rendering historical user message');
 		}
 
-		const shouldIncludePreamble = await isVSCModelA(this.props.endpoint);
+		// Check if the model matches VSC Model A (checks both model ID and family)
+		const shouldIncludePreamble = await isVSCModelA(this.props.endpoint) || await isVSCModelA(this.props.endpoint.family);
```
karthiknadig, line 325: `endpoint` already has the `family` field, should it be explicitly passed? that seems unnecessary.

src/extension/prompts/node/agent/promptRegistry.ts
```diff
@@ -42,14 +42,17 @@ export const PromptRegistry = new class {
 	async getPrompt(
 		endpoint: IChatEndpoint
 	): Promise<IAgentPromptCtor | undefined> {
-
-		for (const prompt of this.promptsWithMatcher) {
-			const matches = await prompt.matchesModel(endpoint);
-			if (matches) {
-				return prompt;
+		// Try matching by model ID first (pass full endpoint), then by family name (pass string)
+		for (const key of [endpoint, endpoint.family]) {
```
karthiknadig, line 46: This is not needed if `matchesModel` takes `endpoint` and in `isVSCModelA` checks both id and family.

src/platform/endpoint/common/chatModelCapabilities.ts
```diff
@@ -19,18 +19,31 @@ const VSC_MODEL_HASHES_A = [
 	'1d28f8e6e5af58c60e9a52385314a3c7bc61f7226e1444e31fe60c58c30e8235',
 	'3104045f9b69dbb7a3d76cc8a0aa89eb05e10677c4dd914655ea87f4be000f4e',
 	'b576d46942ee2c45ecd979cbbcb62688ae3171a07ac83f53b783787f345e3dd7',
-	'b46570bfd230db11a82d5463c160b9830195def7086519ca319c41037b991820',
+	'6b0f165d0590bf8d508540a796b4fda77bf6a0a4ed4e8524d5451b1913100a95',
```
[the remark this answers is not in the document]
karthiknadig: @AbdelrahmanAbouelenin Please review the multiple hash usage.

src/extension/prompts/node/agent/promptRegistry.ts
```diff
@@ -15,14 +15,14 @@ export interface IAgentPrompt {
...
-	matchesModel?(endpoint: IChatEndpoint): Promise<boolean> | boolean;
+	matchesModel?(model: IChatEndpoint | string): Promise<boolean> | boolean;
 	new(...args: any[]): IAgentPrompt;
 }
 
 export type AgentPromptClass = IAgentPromptCtor & (new (...args: any[]) => IAgentPrompt);
 
 type PromptWithMatcher = IAgentPromptCtor & {
-	matchesModel: (endpoint: IChatEndpoint) => Promise<boolean> | boolean;
+	matchesModel: (model: IChatEndpoint | string) => Promise<boolean> | boolean;
```
bhavyaus, line 25: Why is this change needed?

src/extension/prompts/node/agent/promptRegistry.ts
bhavyaus: And endpoint also has the model id. So your implementation of matchesModel should be able to handle this and these changes to promptRegitry are not neeeded.

AbdelrahmanAbouelenin: I wanted to make sure to we model id is checked first across all registry/hashes regardless of order since model id matching should be top priority.
for example, if promptA, promptB and promptC is registered and endpoint family is registered in prompt A, we should check modelID if it get match in promptB and promptC.
if we checked model ID and family we will get a match and return prompt even though the id can be promptB registry.

src/extension/prompts/node/agent/promptRegistry.ts
AbdelrahmanAbouelenin: so we can pass model family name or endpoint and afterward check for model ID.

src/extension/prompts/node/agent/promptRegistry.ts
AbdelrahmanAbouelenin: if you think this is inefficient, we can check for both (ID and family) inside isVSCModelA/B.
but priority will be based on the prompt order in the registry.

PR #1995, opened 2025-11-13
Align tool names

pwang347:
Update names as part of https://github.com/microsoft/vscode/pull/277047

### DIFF
--- a/package.json
+++ b/package.json
@@ -813,7 +813,7 @@
 				"name": "copilot_editNotebook",
 				"toolReferenceName": "editNotebook",
 				"icon": "$(pencil)",
-				"displayName": "%copilot.tools.editNotebook.name%",
+				"displayName": "%copilot.tools.editNotebook.description%",
 				"userDescription": "Edit a notebook file in the workspace",
 				"modelDescription": "This is a tool for editing an existing Notebook file in the workspace. Generate the \"explanation\" property first.\nThe system is very smart and can understand how to apply your edits to the notebooks.\nWhen updating the content of an existing cell, ensure newCode preserves whitespace and indentation exactly and does NOT include any code markers such as (...existing code...).",
 				"tags": [
@@ -1222,8 +1222,7 @@
 			},
 			{
 				"name": "launch",
-				"icon": "$(rocket)",
-				"description": "Launch and run code, binaries or tests in the workspace",
+				"description": "",
 				"tools": [
 					"runNotebookCell",
 					"testFailure"
@@ -1231,6 +1230,7 @@
 			},
 			{
 				"name": "vscode",
+				"description": "",
 				"tools": [
 					"getProjectSetupInfo",
 					"installExtension",
--- a/package.nls.json
+++ b/package.nls.json
@@ -306,6 +306,7 @@
 	"copilot.tools.getDocInfo.description": "For a symbol like a class or function, find the information about how to document it",
 	"copilot.tools.changes.description": "Get diffs of changed files",
 	"copilot.tools.newJupyterNotebook.description": "Create a new Jupyter Notebook",
+	"copilot.tools.editNotebook.description": "Edit a notebook file in the workspace",
 	"copilot.tools.runNotebookCell.description": "Trigger the execution of a cell in a notebook file",
 	"copilot.tools.getNotebookCellOutput.description": "Read the output of a previously executed cell",
 	"copilot.tools.fetchWebPage.description": "Fetch the main content from a web page. You should include the URL of the page you want to fetch.",


package.json
```diff
@@ -1136,45 +1193,59 @@
 				"description": "%copilot.toolSet.editing.description%",
 				"icon": "$(pencil)",
 				"tools": [
-					"createFile",
 					"createDirectory",
+					"createFile",
```
pwang347, line 1192: these are sorted alphabetically, not sure if there was intent behind the original order?

package.json
```diff
@@ -189,9 +189,11 @@
 			{
 				"name": "copilot_listCodeUsages",
 				"toolReferenceName": "usages",
+				"legacyToolReferenceFullNames": [
+					"usages"
```
roblourens, line 193: This needs to be set even when it's the same?

package.json
roblourens: I don't think the order was expected to matter

package.json
```diff
@@ -1136,45 +1185,56 @@
...
 					"listDirectory",
-					"readFile",
-					"codebase",
-					"searchResults"
+					"searchResults",
+					"textSearch"
 				]
 			},
 			{
-				"name": "new",
```
roblourens, line 1170: This toolset is driving a specific scenario for `#new` which helps the user set up a new workspace from scratch. I've never been crazy about how it works but we should either keep it or talk through it with Bhavya. And is `runVscodeCommand` missing from agent mode with this change?

package.json
pwang347: Yeah it looks a bit odd, but because the original reference was top-level the fully qualified name is just the `toolReferenceName` whereas for others it's `toolsetName/toolReferenceName`. This is making sure that existing references to `usages` should show a deprecation squiggle but should otherwise still resolve even after we've added this to a new toolset.

package.json
pwang347: Got it, yes let me chat with Bhavya to make sure we're on the same page.

> is runVscodeCommand missing from agent mode with this change?

I renamed it to `vscode/runCommand` to avoid redundancy, but maybe we should keep it as is?

package.json
```diff
@@ -1136,45 +1185,56 @@
...
-				"icon": "$(notebook)",
+				"name": "read",
+				"description": "%copilot.toolSet.read.description%",
+				"icon": "$(eye)",
 				"tools": [
-					"runCell",
-					"getNotebookSummary",
-					"readNotebookCellOutput"
+					"readFile",
+					"usages"
```
aeschli: Shouldn't usages go to the `search` tools set?

package.json
aeschli: I like the change to move 'runVscodeCommand' under 'vscode' as 'vscode/runCommand'

aeschli: Maybe '#new' should better be a slash-command `/new`. As a slash command it could make sure that it has all tools needed. And give extra context on how to create things. FYI @bhavyaus

package.json
```diff
@@ -1136,45 +1185,56 @@
...
-					"installExtension"
+					"installExtension",
+					"newWorkspace",
+					"openSimpleBrowser",
+					"problems",
+					"runNotebookCell",
+					"runCommand",
+					"testFailure",
+					"vscodeAPI"
+				]
```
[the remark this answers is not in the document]
aeschli: Need to decide what to do with the notebook tools. All under 'runNotebooks' or all under vscode?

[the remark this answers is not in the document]
pwang347: GH splits up the notebook read/edit into read/edit
image

But maybe we can still have a separate notebook toolset with the same tools included?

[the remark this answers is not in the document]
pwang347: discussed this one offline, will keep it as is for now

roblourens (dismissed):

package.json
```diff
@@ -330,6 +330,7 @@
 					"search/readFile"
 				],
 				"displayName": "%copilot.tools.readFile.name%",
+				"userDescription": "Read the contents of a file. You must specify the line range you're interested in. Line numbers are 1-indexed. If the file contents returned are insufficient for your task, you may call this tool again to retrieve more content. Prefer reading larger ranges over doing many small reads.",
```
roblourens, line 333: userDescription should be localized, and should be oriented at the user, not the model

package.json
```diff
@@ -813,6 +814,7 @@
 				"toolReferenceName": "editNotebook",
 				"icon": "$(pencil)",
 				"displayName": "%copilot.tools.editNotebook.name%",
+				"userDescription": "Edit a notebook file in the workspace",
```
roblourens, line 817: Should be localized

pwang347: Thought I committed this, but guess not. Fix here: https://github.com/microsoft/vscode-copilot-chat/pull/2147

Thanks!

PR #2010, opened 2025-11-14
Add client_deviceid to environment service

benibenj:
```Copilot Generated Description:``` Introduce a new property `devDeviceId` to the environment service interface and its implementations, and include it in telemetry properties as `client_deviceid`.

https://github.com/microsoft/vscode/issues/277473


benibenj:
Yes, the candidates have all gone through the discussion today in Zurich

PR #2051, opened 2025-11-17
Support GitHub MCP in-box

TylerLeonhardt:
Fixes https://github.com/microsoft/vscode/issues/254836

This allows the GitHub MCP Server to be readily available in-box.

To start, you must enable it with:
```
"github.copilot.chat.githubMcpServer.enabled": true
```

After that you should see it in your tool picker and if you have already signed in to Copilot Chat with the bigger token (to do codesearch, CCA, etc) no further auth should be required.

Oh, and if you're using GHE.com, that works too.

There's some configuration settings:
```
  "github.copilot.chat.githubMcpServer.toolsets": ["default"],
  "github.copilot.chat.githubMcpServer.readonly": false,
  "github.copilot.chat.githubMcpServer.lockdown": false,
```
which align with https://github.com/github/github-mcp-server/blob/main/docs/remote-server.md#optional-headers

Oh, and if you enable this, `#githubRepo` is disabled since they are redundant.

### Next steps

1. We don't want to enable this by default because it influences token usage... but we want it to be discoverable. Since we are disabled by default, we can solve this in another PR. Ideas: allow me to "uncheck" an MCP Server by default, make this a single virtual tool somehow
2. With GitHub being available from the MCP Registry, what should we do about the conflict there?
3. Should there be alignment with what scopes we enable in Copilot CLI?

### Blocker

GitHub MCP is rolling out upscoping ...which is what we need to really support this. They've laid a foundation down with the `search_orgs` tool which will say you need `read:orgs` to query it if you give it a token that does not have that but:
1. the `resource_metadata` is currently wrong (small fix)
2. it tells you the scope you're missing but it should instead return that scope added to the scopes that your token already had

So both those things, along with rolling it out to all tools, should be done before this could be enabled by default.


karthiknadig (dismissed):

DonJayamanne (dismissed):

TylerLeonhardt:
@copilot resolve conflicts

aeschli (approved):

PR #2065, opened 2025-11-18
Implement supportedEndoints for BYOK

roblourens:


koureasstavros:
I do not understand, is this one commit makes chat support the responses endpoint for Azure Provider for example?

lramos15:
> I do not understand, is this one commit makes chat support the responses endpoint for Azure Provider for example?

Azure already supports the responses endpoint. Just do make sure to include the fully qualified url with `/responses`. This PR makes it so we can specify some models as responses only i.e. GPT 5 Codex via our remote model configs

koureasstavros:
> > I do not understand, is this one commit makes chat support the responses endpoint for Azure Provider for example?
> 
> Azure already supports the responses endpoint. Just do make sure to include the fully qualified url with `/responses`. This PR makes it so we can specify some models as responses only i.e. GPT 5 Codex via our remote model configs

Yes, alright so you took the deployment name from key of the dictionary from the settings.json.
This works but it might be a little bit misleading as the completions models can work as the deployment name is into the url and it's not mandatory to put it in the model-id.

PR #2069, opened 2025-11-18
Re-check editor preview flag for auto

lramos15:


joshspicer:
verified

## false
Screenshot 2025-11-18 at 4 42 41 PM

## true
Screenshot 2025-11-18 at 4 45 32 PM

PR #2070, opened 2025-11-18
Add EntraId authentication support for Azure specific endpoints

eleanorjboyd:
fixes https://github.com/microsoft/vscode/issues/270234

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -169,7 +169,7 @@
 	"github.copilot.config.codeGeneration.instruction.file": "A path to a file that will be added to Copilot requests that generate code. Optionally, you can specify a language for the instruction.",
 	"github.copilot.config.azureAuthType": "Authentication method for Azure OpenAI models. Entra ID is recommended for enterprise security and uses your Azure credentials.",
 	"github.copilot.config.azureAuthType.entraId": "Use Entra ID (Azure AD) authentication with your Microsoft account credentials",
-	"github.copilot.config.azureAuthType.apiKey": "Use API key authentication",
+	"github.copilot.config.azureAuthType.apiKey": "Use API key authentication. Not recommended as this is less secure than token based authentication.",
 	"github.copilot.config.testGeneration.instructions": "A set of instructions that will be added to Copilot requests that generate tests.\nInstructions can come from: \n- a file in the workspace: `{ \"file\": \"fileName\" }`\n- text in natural language: `{ \"text\": \"Use underscore for field names.\" }`\n\nNote: Keep your instructions short and precise. Poor instructions can degrade Copilot's quality and performance.",
 	"github.copilot.config.testGeneration.instructions.deprecated": "Use instructions files instead. See https://aka.ms/vscode-ghcp-custom-instructions for more information.",
 	"github.copilot.config.experimental.testGeneration.instruction.text": "A text instruction that will be added to Copilot requests that generate tests. Optionally, you can specify a language for the instruction.",
--- a/src/extension/byok/vscode-node/azureProvider.ts
+++ b/src/extension/byok/vscode-node/azureProvider.ts
@@ -83,7 +83,7 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
 					await vscode.authentication.getSession(
 						AzureAuthMode.MICROSOFT_AUTH_PROVIDER,
 						[AzureAuthMode.COGNITIVE_SERVICES_SCOPE],
-						{ createIfNone: true, silent: false }
+						{ createIfNone: true }
 					);
 				} catch (error) {
 					// If sign-in fails, don't show models in picker
@@ -123,8 +123,8 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
 			const modelInfo = await this.getModelInfo(model.id, undefined, {
 				maxInputTokens: model.maxInputTokens,
 				maxOutputTokens: model.maxOutputTokens,
-				toolCalling: !!model.capabilities?.toolCalling || false,
-				vision: !!model.capabilities?.imageInput || false,
+				toolCalling: !!model.capabilities?.toolCalling,
+				vision: !!model.capabilities?.imageInput,
 				name: model.name,
 				url: model.url,
 				thinking: model.thinking,
--- a/src/util/common/test/shims/vscodeTypesShim.ts
+++ b/src/util/common/test/shims/vscodeTypesShim.ts
@@ -121,8 +121,8 @@ const shim = {
 	FileType,
 	ChatSessionStatus,
 	authentication: {
-		getSession: () => Promise.reject(new Error('authentication.getSession not mocked in test'))
-	} as any
+		getSession: (providerId: string, scopes: readonly string[], options?: { createIfNone?: boolean; silent?: boolean; clearSessionPreference?: boolean; forceNewSession?: boolean }) => Promise.reject(new Error('authentication.getSession not mocked in test'))
+	}
 };
 
 export = shim;
--- a/src/extension/byok/vscode-node/test/azureProvider.spec.ts
+++ b/src/extension/byok/vscode-node/test/azureProvider.spec.ts
@@ -125,22 +125,6 @@ describe('AzureBYOKModelProvider', () => {
 			provider = instaService.createInstance(AzureBYOKModelProvider, mockByokStorageService);
 		});
 
-		it('should return all models when authentication succeeds in non-silent mode', async () => {
-			const mockSession = { accessToken: 'test-token', account: { id: 'test', label: 'test' }, scopes: [], id: 'test' };
-			vi.spyOn(vscode.authentication, 'getSession').mockResolvedValue(mockSession);
-
-			const models = await provider['getModelsWithCredentials'](false);
-
-			expect(vscode.authentication.getSession).toHaveBeenCalledWith(
-				AzureAuthMode.MICROSOFT_AUTH_PROVIDER,
-				[AzureAuthMode.COGNITIVE_SERVICES_SCOPE],
-				{ createIfNone: true, silent: false }
-			);
-			expect(Object.keys(models)).toHaveLength(2);
-			expect(models['gpt-4']).toBeDefined();
-			expect(models['gpt-35-turbo']).toBeDefined();
-		});
-
 		it('should return all models without prompting authentication in silent mode', async () => {
 			const getSessionSpy = vi.spyOn(vscode.authentication, 'getSession');
 
@@ -162,7 +146,7 @@ describe('AzureBYOKModelProvider', () => {
 			expect(vscode.authentication.getSession).toHaveBeenCalledWith(
 				AzureAuthMode.MICROSOFT_AUTH_PROVIDER,
 				[AzureAuthMode.COGNITIVE_SERVICES_SCOPE],
-				{ createIfNone: true, silent: false }
+				{ createIfNone: true }
 			);
 		});
 


src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -65,4 +69,81 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
...
+			});
+
+			const openAIChatEndpoint = this._instantiationService.createInstance(
+				AzureOpenAIEndpoint,
+				modelInfo,
+				session.accessToken,  // Pass Entra ID token
+				model.url
+			);
+
+			return this._lmWrapper.provideLanguageModelResponse(
```
eleanorjboyd, line 142: @Advitya17 any return codes we should handle that aren't "normal" https responses? We catch the classic 404, 403 etc ones

package.json
```diff
@@ -3087,6 +3012,18 @@
 						],
 						"markdownDescription": "%github.copilot.config.virtualTools.threshold%"
 					},
+					"github.copilot.chat.azureAuthType": {
```
eleanorjboyd, line 3215: @Advitya17 thoughts on the setting here? that it would be a per user basis since most users would want to auth one way

src/extension/byok/vscode-node/azureProvider.ts
Advitya17: Technically AI Foundry deployments may also have 5xx Internal Server Errors, so could look into those as well. Presuming this functionality is supposed to be general-purpose towards AI Foundry (LLM) deployments.

src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -65,4 +69,89 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
...
+				session = await authentication.getSession(
+					'microsoft',
+					['https://cognitiveservices.azure.com/.default'],
+					{ createIfNone: true }
+				);
+			} catch (err) {
+				const errorMessage = 'Azure authentication failed. Please sign in to your Microsoft account and try again.';
+				this._logService.error('[AzureBYOKModelProvider] Authentication failed during chat request:', err);
+				throw new Error(errorMessage);
+			}
```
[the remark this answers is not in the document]
Advitya17: @eleanorjboyd this could be interesting to look into, just as an improvement (so, non-blocking comment)

package.json
Advitya17: Interesting experience, sounds good to me. Also great to see default will be Entra ID, especially for lots of Azure/Microsoft users for whom this default is typically desirable and has less friction!

src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -65,4 +69,89 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
 	protected override resolveUrl(modelId: string, url: string): string {
 		return resolveAzureUrl(modelId, url);
 	}
+
+	protected override async getModelsWithCredentials(silent: boolean): Promise<BYOKKnownModels> {
+		// Check user's authentication preference from settings github.copilot.chat.azureAuthType (default: 'entraId')
+		const authType = this._configurationService.getConfig(ConfigKey.AzureAuthType);
+
+		if (authType === 'entraId') {
```
Advitya17: ideally store these constant strings in an enum class that stores EntraID key/s key-based auth options

src/extension/byok/vscode-node/azureProvider.ts
Advitya17: Also 400 Bad Requests (such as max output token limit greater than allowed, invalid model param values, or input prompt > input prompt token limit), and 429 Too Many API Requests (due to rate limits) are quite common to see too!

src/platform/configuration/common/configurationService.ts
```diff
@@ -856,6 +856,7 @@ export namespace ConfigKey {
 	export const CurrentEditorAgentContext = defineSetting<boolean>('chat.agent.currentEditorContext.enabled', ConfigType.Simple, true);
 	/** BYOK  */
 	export const OllamaEndpoint = defineSetting<string>('chat.byok.ollamaEndpoint', ConfigType.Simple, 'http://localhost:11434');
+	export const AzureAuthType = defineSetting<'entraId' | 'apiKey'>('chat.azureAuthType', ConfigType.Simple, 'entraId');
```
Advitya17: Enum vars for these 2 options ideally

src/extension/byok/vscode-node/azureProvider.ts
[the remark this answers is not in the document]
eleanorjboyd: the current design was @TylerLeonhardt suggestion based on his experience with this auth stuff - I think it was to just make sure there is nothing stale 

cc @TylerLeonhardt

src/extension/byok/vscode-node/azureProvider.ts
eleanorjboyd: investigated to makes sure these are handled by the parent and I think all these mentioned are- thanks for confirming what to look for!

src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -108,11 +109,18 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
...
-			);
+			let session;
+			try {
+				session = await authentication.getSession(
+					'microsoft',
+					['https://cognitiveservices.azure.com/.default'],
+					{ createIfNone: true }
+				);
+			} catch (err) {
+				const errorMessage = 'Azure authentication failed. Please sign in to your Microsoft account and try again.';
```
lramos15: What does this look like? Is it a bad experience?

src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -65,4 +69,89 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
...
+						AzureAuthMode.MICROSOFT_AUTH_PROVIDER,
+						[AzureAuthMode.COGNITIVE_SERVICES_SCOPE],
+						{ createIfNone: true }
+					);
+				} catch (error) {
+					// If sign-in fails, don't show models in picker
+					this._logService.error('[AzureBYOKModelProvider] Authentication failed during Entra ID sign-in:', error);
+					return {};
+				}
+			}
```
TylerLeonhardt, line 93: there's a `silent: true` flag in addition to a `createIfNone` btw in case you wanna try to get this state silently.

src/extension/byok/vscode-node/azureProvider.ts
eleanorjboyd: Yeah good catch, gave it a try with a few options and I think not catching the error here and instead just letting the parent catch it and display it looks better as it avoids a really long stack trace mapping to this file and function. This is what it looks like now that ive updated it.

Image

src/extension/byok/vscode-node/azureProvider.ts
eleanorjboyd: I feel like it should pop up but open to input from others

TylerLeonhardt: I mean for when the `silent` variable is true.

TylerLeonhardt: Oh I see this is only used to prompt.

src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -65,4 +70,87 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
...
+		progress: Progress<LanguageModelResponsePart2>,
+		token: CancellationToken
+	): Promise<void> {
+		const authType = this._configurationService.getConfig(ConfigKey.AzureAuthType);
+
+		if (authType === AzureAuthMode.EntraId) {
+			const session = await vscode.authentication.getSession(
+				AzureAuthMode.MICROSOFT_AUTH_PROVIDER,
+				[AzureAuthMode.COGNITIVE_SERVICES_SCOPE],
+				{ createIfNone: true }
```
TylerLeonhardt: `session` is always defined when you do `createIfNone: true`. The types should say `session` is an AuthSession.

package.json
```diff
@@ -3210,6 +3210,18 @@
...
+						"type": "string",
+						"enum": [
+							"entraId",
+							"apiKey"
+						],
+						"default": "entraId",
+						"tags": [
+							"experimental"
+						],
+						"markdownDescription": "Authentication method for Azure OpenAI models. Use `entraId` for Entra ID (Azure AD) authentication or `apiKey` for API key authentication. Entra ID is recommended for enterprise security and uses your Azure credentials."
```
TylerLeonhardt: localize this

package.json
```diff
@@ -3210,6 +3210,18 @@
 						],
 						"markdownDescription": "%github.copilot.config.virtualTools.threshold%"
 					},
+					"github.copilot.chat.azureAuthType": {
+						"type": "string",
+						"enum": [
```
TylerLeonhardt, line 3217: should add enumDescriptions too and localize them

src/extension/byok/vscode-node/azureProvider.ts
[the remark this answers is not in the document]
TylerLeonhardt: we cache at the auth provider layer. Caching shouldn't be done at that extension level. This is the right way to call this.

src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -110,17 +110,16 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
...
 		if (authType === AzureAuthMode.EntraId) {
-			const session = await vscode.authentication.getSession(
+			// Session is guaranteed to be defined when createIfNone: true
+			const session: vscode.AuthenticationSession = await vscode.authentication.getSession(
 				AzureAuthMode.MICROSOFT_AUTH_PROVIDER,
 				[AzureAuthMode.COGNITIVE_SERVICES_SCOPE],
-				{ createIfNone: true }
+				{
+					createIfNone: true,
+					silent: false
```
TylerLeonhardt, line 119: redundant

src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -110,17 +110,16 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
 		const authType = this._configurationService.getConfig(ConfigKey.AzureAuthType);
 
 		if (authType === AzureAuthMode.EntraId) {
-			const session = await vscode.authentication.getSession(
+			// Session is guaranteed to be defined when createIfNone: true
```
TylerLeonhardt, line 113: the typing for getSession should say this already.

package.nls.json
```diff
@@ -167,6 +167,9 @@
 	"github.copilot.config.codeGeneration.useInstructionFiles": "Controls whether code instructions from `.github/copilot-instructions.md` are added to Copilot requests.\n\nNote: Keep your instructions short and precise. Poor instructions can degrade Copilot's quality and performance. [Learn more](https://aka.ms/github-copilot-custom-instructions) about customizing Copilot.",
 	"github.copilot.config.codeGeneration.instruction.text": "A text instruction that will be added to Copilot requests that generate code. Optionally, you can specify a language for the instruction.",
 	"github.copilot.config.codeGeneration.instruction.file": "A path to a file that will be added to Copilot requests that generate code. Optionally, you can specify a language for the instruction.",
+	"github.copilot.config.azureAuthType": "Authentication method for Azure OpenAI models. Entra ID is recommended for enterprise security and uses your Azure credentials.",
+	"github.copilot.config.azureAuthType.entraId": "Use Entra ID (Azure AD) authentication with your Microsoft account credentials",
+	"github.copilot.config.azureAuthType.apiKey": "Use API key authentication",
```
TylerLeonhardt: consider mentioning that api key auth is deprecated if AI Foundry says so. If they don't say so, then never mind

src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -65,4 +70,86 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
...
+					createIfNone: true,
+					silent: false
+				}
+			);
+
+			const modelInfo = await this.getModelInfo(model.id, undefined, {
+				maxInputTokens: model.maxInputTokens,
+				maxOutputTokens: model.maxOutputTokens,
+				toolCalling: !!model.capabilities?.toolCalling || false,
+				vision: !!model.capabilities?.imageInput || false,
```
TylerLeonhardt: you have the `!!` in front so the `|| false` is redundant

src/util/common/test/shims/vscodeTypesShim.ts
```diff
@@ -118,7 +119,10 @@ const shim: typeof vscodeTypes = {
 	SnippetString,
 	SnippetTextEdit,
 	FileType,
-	ChatSessionStatus
+	ChatSessionStatus,
+	authentication: {
+		getSession: () => Promise.reject(new Error('authentication.getSession not mocked in test'))
+	} as any
```
TylerLeonhardt: no as any pls

eleanorjboyd: oops sorry!

src/extension/byok/vscode-node/azureProvider.ts
```diff
@@ -65,4 +70,86 @@ export class AzureBYOKModelProvider extends CustomOAIBYOKModelProvider {
...
+
+		if (authType === AzureAuthMode.EntraId) {
+			// Pre-authenticate during model enumeration (not when sending message)
+			// This mirrors API key behavior where user is prompted during enumeration
+			if (!silent) {
+				try {
+					await vscode.authentication.getSession(
+						AzureAuthMode.MICROSOFT_AUTH_PROVIDER,
+						[AzureAuthMode.COGNITIVE_SERVICES_SCOPE],
+						{ createIfNone: true, silent: false }
```
TylerLeonhardt: sorry if my previous statement with silent was confusing.


```suggestion
						{ createIfNone: true }
```

package.nls.json
eleanorjboyd: this is as far as they go in the docs so I updated to reflect that: https://learn.microsoft.com/en-us/azure/ai-foundry/openai/latest?view=foundry-classic#request-header

PR #2105, opened 2025-11-20
Navigating the codebase like a developer: f12/symbolic navigation tools

OmerMachluf:
This is an exploration PR: It is not fully ready yet.
The core idea behind it is to reduce the number of text searches which are expensive, yield many results and limit the capability to work properly in big/huge repositories.
Instead, this aims to provide tools for copilot that allows him to navigate the codebase like a developer would.
It leverages LSP to provide code referneces, f12 capability, find references/implementations etc or get the document full symbols to understand structure and allow to focus on which lines to read beforehand.
I haven't done official benchmarking yet(And I was curious if there are any suites already designed for that purpose), but from my own internal tests I am able to complete tasks that would have usually yielded or give up due to context loss or too many searches yielding nothing.
Execution happens faster and more realibly as the copilot simply navigates between documents he's exploring or modifying using semantic references.
I wanted to get a ack that there are no blockers with this approach before doubling down on the time allocated to complete this fully.
Among the gaps that I currently recognize are:
- Official benchmarking test suites(if any) validation based more on mere my own experience
- Experimental flag: having this as an opt-in rather than enabled by default
- having fallback for languages or workspaces without lsp support or when language symbols are not loaded yet
- Refining the prompt as sometimes i find myself reminding it to use symbolic navigation over search while symbolic navigation works faster and more reliably


OmerMachluf:
> @OmerMachluf please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree [company="{Microsoft}"]

PR #2116, opened 2025-11-20
Fix up for the GitDiffService

IanMatthewHuff:
The repoInfoTelemetry service was being used by the science team for some model development. This telemetry uses the gitDiffService to provide diffs and there were some git diff patches that were not able to be applied coming specifically from new untracked files.

These new untracked files are actually handled a bit differently as the main git services doesn't pick up these changes. Instead a synthetic patch is created via _getUntrackedChangePatch. This function had some issue with generating an actual synthetic patch that could be applied without fixup.

Fixes:

- [x] For a totally empty file, correctly omit range / lines
- [x] For the range header correctly use lines and not byte length
- [x] Add 'new file mode' header
- [x] Correctly add 'no newline' footer when needed
- [x] Add unit tests for this function

Note: This fix is specifically for the usage of this code path for repoInfoTelemetry, but this existing function was used already by gitCommitMessageService and scmChanges tool. In both those changes the git diffs look to be just handled as a string and passed to an LLM, so improving the accuracy here should not cause any issues. But that is why I avoided added in extra (possibly risky) work for checking file permissions or calculating the SHA1 hash (which are not required to apply the patch) into this work.

Manually tested with pulling the telemetry locally and creating and applying a diff patch for:
1. New untracked file with content and no end new line
2. New untracked file with content and an ending new line
3. New fully empty untracked file
4. New file with just a new line

Patch applied and created the file as it existed before in all four cases.

### DIFF
--- a/src/extension/prompt/vscode-node/gitDiffService.ts
+++ b/src/extension/prompt/vscode-node/gitDiffService.ts
@@ -124,7 +124,7 @@ export class GitDiffService implements IGitDiffService {
 			patch.push('new file mode 100644');
 			patch.push('--- /dev/null', `+++ b/${relativePath}`);
 
-			// For non-empty files, add range header and content (empty file omit this)
+			// For non-empty files, add range header and content (empty files omit this)
 			if (content.length > 0) {
 				const lines = content.split('\n');
 				if (content.endsWith('\n')) {
--- a/src/extension/prompt/vscode-node/test/gitDiffService.spec.ts
+++ b/src/extension/prompt/vscode-node/test/gitDiffService.spec.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { MockInstance, afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import * as vscode from 'vscode';
 import { IGitExtensionService } from '../../../../platform/git/common/gitExtensionService';
 import { API, Change, Repository } from '../../../../platform/git/vscode/git';
@@ -14,7 +14,7 @@ import { createExtensionUnitTestingServices } from '../../../test/node/services'
 import { GitDiffService } from '../gitDiffService';
 
 describe('GitDiffService', () => {
-	let readFileSpy: any;
+	let readFileSpy: MockInstance<typeof vscode.workspace.fs.readFile>;
 	let accessor: ITestingServicesAccessor;
 	let gitDiffService: GitDiffService;
 	let mockRepository: Partial<Repository>;
@@ -22,7 +22,8 @@ describe('GitDiffService', () => {
 	beforeEach(() => {
 		// Create mock workspace.fs.readFile if it doesn't exist
 		if (!vscode.workspace?.fs?.readFile) {
-			(vscode as any).workspace = {
+			const workspaceWithFs = vscode as unknown as { workspace: typeof vscode.workspace };
+			workspaceWithFs.workspace = {
 				...vscode.workspace,
 				fs: {
 					...vscode.workspace?.fs,


src/extension/prompt/vscode-node/test/gitDiffService.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+
+describe('GitDiffService', () => {
+	let readFileSpy: any;
+	let accessor: ITestingServicesAccessor;
+	let gitDiffService: GitDiffService;
+	let mockRepository: Partial<Repository>;
+
+	beforeEach(() => {
+		// Create mock workspace.fs.readFile if it doesn't exist
+		if (!vscode.workspace?.fs?.readFile) {
```
IanMatthewHuff, line 24: The product code here uses workspace.fs directly and not the file service, so I added this mock here so that my changes for this test were just in this test file as opposed to trying to mess with the top level vscode test shims.

src/extension/prompt/vscode-node/gitDiffService.ts
```diff
@@ -114,22 +114,38 @@ export class GitDiffService implements IGitDiffService {
...
 			const buffer = await workspace.fs.readFile(resource);
 			const relativePath = path.relative(repository.rootUri.fsPath, resource.fsPath);
+			const content = buffer.toString();
 
 			// Header
 			patch.push(`diff --git a/${relativePath} b/${relativePath}`);
-
-			// Add original/modified file paths
+			// 100644 is standard file mode for new git files. Saves us from trying to check file permissions and handling
+			// UNIX vs Windows permission differences. Skipping calculating the SHA1 hashes as well since they are not strictly necessary
```
IanMatthewHuff, line 122: I put a comment here on this. Did a bit of research into what the diff patch actually contains and what was required to actually apply a patch.

Using 100644 for the file mode simplifies not having to do extra file permission checks for a synthetic patch and handles the basic git new file scenario. Also the SHA1 hash line is skipped as git patch doesn't require it for applying so it's just more calculations that we don't need.

src/extension/prompt/vscode-node/gitDiffService.ts
```diff
@@ -114,22 +114,38 @@ export class GitDiffService implements IGitDiffService {
...
+			// For non-empty files, add range header and content (empty file omit this)
+			if (content.length > 0) {
+				const lines = content.split('\n');
+				if (content.endsWith('\n')) {
+					// Prevent an extra empty line at the end
+					lines.pop();
+				}
+
+				// Range header and content
+				patch.push(`@@ -0,0 +1,${lines.length} @@`);
```
IanMatthewHuff, line 136: Bug with the old code was here, where it was using the buffer.length and not the lines length for the range header. This would lead to patches that looked incorrect in size.

src/extension/prompt/vscode-node/gitDiffService.ts
```diff
@@ -114,22 +114,38 @@ export class GitDiffService implements IGitDiffService {
...
+					patch.push('\\ No newline at end of file');
+				}
+			}
 		} catch (err) {
 			console.error(err, `Failed to generate patch file for untracked file: ${resource.toString()}`);
 		}
 
-		return patch.join('\n');
+		// The patch itself should always end with a newline per git patch standards
+		return patch.join('\n') + '\n';
```
IanMatthewHuff, line 149: The patch itself should always end with a newline.

src/extension/prompt/vscode-node/gitDiffService.ts
```diff
@@ -114,22 +114,38 @@ export class GitDiffService implements IGitDiffService {
...
+					lines.pop();
+				}
+
+				// Range header and content
+				patch.push(`@@ -0,0 +1,${lines.length} @@`);
+				patch.push(...lines.map(line => `+${line}`));
+
+				// Git standard to add this comment if the file does not end with a newline
+				if (!content.endsWith('\n')) {
+					patch.push('\\ No newline at end of file');
```
IanMatthewHuff, line 141: This comment was also missing in the old code. Prevents you from getting a warning when applying a patch that doesn't end with a new line.

src/extension/prompt/vscode-node/test/gitDiffService.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+
+		accessor = services.createTestingAccessor();
+		gitDiffService = accessor.get(IInstantiationService).createInstance(GitDiffService);
+	});
+
+	afterEach(() => {
+		readFileSpy.mockRestore();
+	});
+
+	describe('_getUntrackedChangePatch', () => {
```
IanMatthewHuff, line 64: GitDiffService didn't have unit tests previously. But while looking into this some of the cases (newline at end, newline not at end, empty file) were a bit quirky, so added some quick tests in for those specific cases.

src/extension/prompt/vscode-node/test/gitDiffService.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+import { IGitExtensionService } from '../../../../platform/git/common/gitExtensionService';
+import { API, Change, Repository } from '../../../../platform/git/vscode/git';
+import { ITestingServicesAccessor } from '../../../../platform/test/node/services';
+import { IInstantiationService } from '../../../../util/vs/platform/instantiation/common/instantiation';
+import { Uri } from '../../../../vscodeTypes';
+import { createExtensionUnitTestingServices } from '../../../test/node/services';
+import { GitDiffService } from '../gitDiffService';
+
+describe('GitDiffService', () => {
+	let readFileSpy: any;
```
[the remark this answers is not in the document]
IanMatthewHuff: Whoops, forgot that I just made this change for another test file. I'll fix this up quick.

src/extension/prompt/vscode-node/test/gitDiffService.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+
+			// Empty file case: git omits range header and content for totally empty files
+			const patch = diffs[0].diff;
+			expect(patch).toContain('diff --git a/empty.txt b/empty.txt');
+			expect(patch).toContain('new file mode 100644');
+			expect(patch).toContain('--- /dev/null');
+			expect(patch).toContain('+++ b/empty.txt');
+			// No range header for empty files
+			expect(patch).not.toContain('@@');
+			// No content lines
```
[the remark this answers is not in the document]
IanMatthewHuff: Don't really feel this is needed for a test line like this, easy enough to ask what the regex is checking versus over-commenting test code.

src/extension/prompt/vscode-node/test/gitDiffService.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+	beforeEach(() => {
+		// Create mock workspace.fs.readFile if it doesn't exist
+		if (!vscode.workspace?.fs?.readFile) {
+			(vscode as any).workspace = {
+				...vscode.workspace,
+				fs: {
+					...vscode.workspace?.fs,
+					readFile: vi.fn()
+				}
+			};
```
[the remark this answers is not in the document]
IanMatthewHuff: Hmm, so I did notice this was a bit odd. My goal here was that I didn't want to mess with the existing gitDiffService since other code paths are using this existing code. And gitDiffService was using workspace.fs directly and not the file system service for file operation. So this test bit here I did so that I could unit test this function without having to change the product code (workspace.fs -> WorkspaceFileService) or make bigger changes to the vscode test shim. If we don't like this for a test file, down to change it to one of the other options.

IanMatthewHuff:
@Yoyokrazy @sbatten has context on this change. But you can ping me as well if you need any info.

src/extension/prompt/vscode-node/test/gitDiffService.spec.ts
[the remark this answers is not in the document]
IanMatthewHuff: I did remove the 'as any' for a more strict type here, but I think the core issue might still apply?

IanMatthewHuff:
@vritant24 Tagging for context while I'm oof

sbatten:
@lszomoru I'd really appreciate your review

lszomoru (approved):
LGTM! Great work 👏🏻

PR #2131, opened 2025-11-21
Inline completions in @vscode/chat-lib

jeffreyhunter77:
Include inline completions in @vscode/chat-lib

This exports a function for creating an `IInlineCompletionsProvider`:

```ts
function createInlineCompletionsProvider(options: IInlineCompletionsProviderOptions): IInlineCompletionsProvider
```

The returned instance provides methods for updating treatment variables and getting inline completions:

```ts
export interface IInlineCompletionsProvider {
	updateTreatmentVariables(variables: Record<string, boolean | number | string>): void;
	getInlineCompletions(textDocument: ITextDocument, position: Position, token?: CancellationToken, options?: IGetInlineCompletionsOptions): Promise<CopilotCompletion[] | undefined>;
	dispose(): void;
}
```

The options object for creating a provider is fairly complex. It allows for customizing much of what was previously passed in to the completions function via a `Context` object. This is probably a good area for simplifying in future iterations.

I also discovered an undeclared dependency that I've added to the main `package.json`. Some of the [completions code](https://github.com/microsoft/vscode-copilot-chat/blob/b8345420d6e831fc3cc59f61f93eec3de4ed6896/src/extension/completions-core/vscode-node/lib/src/prompt/contextProviders/contextItemSchemas.ts#L6-L7) depends on `@sinclair/typebox`. That's being indirectly supplied by `@vitest/snapshot` right now, but it we should have an explicit dependency. I'm using the same version as the completions extension here.

### DIFF
--- a/chat-lib/package-lock.json
+++ b/chat-lib/package-lock.json
@@ -752,7 +752,6 @@
 			"integrity": "sha512-3giAOQvZiH5F9bMlMiv8+GSPMeqg0dbaeo58/0SlA9sxSqZhnUtxzX9/2FzyhS9sWQf5S0GJE0AKBrFqjpeYcg==",
 			"dev": true,
 			"license": "Apache-2.0",
-			"peer": true,
 			"engines": {
 				"node": ">=8.0.0"
 			}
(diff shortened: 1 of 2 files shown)


jeffreyhunter77:

PR #2156, opened 2025-11-23
Move all instructions into model-specific prompt

roblourens:
@bhavyaus are you ok with this? I wanted to customize it for 5.1 codex. Or do you think these rules shouldn't be customizable so we don't drop them by accident?


roblourens:
I see #2202, I guess that's the direction

PR #2165, opened 2025-11-24
Handle partial tool streaming

lramos15:
Allows for streaming of progress as a tool partially streams in. 

This is the extension side of https://github.com/microsoft/vscode/pull/278640

### DIFF
--- a/src/extension/prompt/node/pseudoStartStopConversationCallback.ts
+++ b/src/extension/prompt/node/pseudoStartStopConversationCallback.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as l10n from '@vscode/l10n';
+import { disableErrorLogging, parse as parsePartialJson } from 'best-effort-json-parser';
 import type { ChatResponseStream, ChatVulnerability } from 'vscode';
 import { IResponsePart } from '../../../platform/chat/common/chatMLFetcher';
 import { IResponseDelta } from '../../../platform/networking/common/fetch';
@@ -14,6 +15,8 @@ import { URI } from '../../../util/vs/base/common/uri';
 import { ChatResponseClearToPreviousToolInvocationReason } from '../../../vscodeTypes';
 import { getContributedToolName } from '../../tools/common/toolNames';
 import { IResponseProcessor, IResponseProcessorContext } from './intents';
+
+disableErrorLogging();
 
 export interface StartStopMapping {
 	readonly stop: string;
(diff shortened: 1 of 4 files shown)


src/extension/tools/node/createFileTool.tsx
```diff
@@ -180,6 +180,37 @@ export class CreateFileTool implements ICopilotTool<ICreateFileParams> {
 		};
 	}
 
+	async handleToolStream(options: vscode.LanguageModelToolInvocationStreamOptions<ICreateFileParams>, token: vscode.CancellationToken): Promise<vscode.LanguageModelToolStreamResult> {
+		let invocationMessage: MarkdownString;
+
+		if (options.rawInput && typeof options.rawInput === 'string') {
+			// Try to extract filePath from the raw JSON input
+			const filePathMatch = options.rawInput.match(/"filePath"\s*:\s*"([^"]+)"/);
```
roblourens: Not a fan of this, I think the agent would ideally have a streaming json parser and return partial input objects (for now all you really have to do is append a `}` and parse it right?

src/extension/prompt/node/pseudoStartStopConversationCallback.ts
```diff
@@ -69,7 +69,18 @@ export class PseudoStopStartResponseProcessor implements IResponseProcessor {
...
+				progress.beginToolInvocation(beginCall.id ?? '', getContributedToolName(beginCall.name), {});
+			}
+		}
+
+		if (delta.copilotToolCallStreamUpdates?.length) {
+			for (const update of delta.copilotToolCallStreamUpdates) {
+				if (!update.name) {
+					continue;
+				}
+				progress.updateToolInvocation(update.id ?? '', { partialInput: tryParsePartialToolInput(update.arguments) });
```
connor4312, line 85: I don't think this should try to auto-close input like this. Instead we should use a streaming JSON parser, [`stream-json`](https://www.npmjs.com/package/stream-json) seems popular, in the components that need it. That'll be more reliable than trying to auto-close like this

src/extension/tools/node/createFileTool.tsx
```diff
@@ -180,6 +180,40 @@ export class CreateFileTool implements ICopilotTool<ICreateFileParams> {
...
+
+		if (partialInput && typeof partialInput === 'object') {
+			const filePath = partialInput.filePath;
+			const content = partialInput.content;
+
+			if (filePath) {
+				const uri = resolveToolInputPath(filePath, this.promptPathRepresentationService);
+
+				if (content !== undefined) {
+					const lineCount = content.split('\n').length;
```
connor4312: there's a `count` utility in the common strings.ts that's faster here (especially since I think you parse this on every stream chunk)

PR #2181, opened 2025-11-25
Merging hashes

AbdelrahmanAbouelenin:
Moving model B hashes to Model A.


src/platform/endpoint/common/chatModelCapabilities.ts
```diff
@@ -21,14 +21,15 @@ const VSC_MODEL_HASHES_A = [
...
-
-const VSC_MODEL_HASHES_B = [
 	'e30111497b2a7e8f1aa7beed60b69952537d99bcdc18987abc2f6add63a89960',
 	'df610ed210bb9266ff8ab812908d5837538cdb1d7436de907fb7e970dab5d289',
 	'6db59e9bfe6e2ce608c0ee0ade075c64e4d054f05305e3034481234703381bb5',
 ];
 
+const VSC_MODEL_HASHES_B = [
+
+];
```
karthiknadig: Please leave a comment as to why this is empty.

Also I recommend adding it like this:
`const VSC_MODEL_HASHES_B : string[] = [];`

karthiknadig (changes requested):

src/platform/endpoint/common/chatModelCapabilities.ts
AbdelrahmanAbouelenin: done

PR #2187, opened 2025-11-25
Make auto token bank aware of the chat location

jrieken:
https://github.com/microsoft/vscode/issues/277857


src/platform/endpoint/node/automodeService.ts
```diff
@@ -68,7 +70,11 @@ class AutoModeTokenBank extends Disposable {
 			headers['Copilot-Session-Token'] = this._token.session_token;
 		}
 
-		const autoModeHint = this._expService.getTreatmentVariable<string>('copilotchat.autoModelHint') || 'auto';
+		const expName = this._location === ChatLocation.Editor
```
lramos15, line 73: Should we make it so experiments can be configured for any location?

jrieken: Yeah, I think that's what we want in the future but given the candidate-character here I didn't wanna do it yet

PR #2194, opened 2025-11-25
ChatRequestProvider: clean up a little

ulugbekna:
preparing to add support for nested NES entries

cc @roblourens as most involved in this code 

maybe the most important part to pay attention to is that there was duplication as in this screenshot which I didn't understand why and thought that was a bug and removed

CleanShot 2025-11-25 at 17 50 53@2x


roblourens:
> there was duplication as in this screenshot which I didn't understand why and thought that was a bug and removed

Huh, no clue, either my mistake or ai slop

PR #2202, opened 2025-11-26
Refactor agent prompt system to use centralized customization resolution

bhavyaus:
For: https://github.com/microsoft/vscode/issues/274856

Will clean up and consolidate tool enablement in another pr.

### DIFF
--- a/src/extension/prompts/node/agent/test/summarization.spec.tsx
+++ b/src/extension/prompts/node/agent/test/summarization.spec.tsx
@@ -29,6 +29,7 @@ import { createExtensionUnitTestingServices } from '../../../../test/node/servic
 import { ToolName } from '../../../../tools/common/toolNames';
 import { PromptRenderer } from '../../base/promptRenderer';
 import { AgentPrompt, AgentPromptProps } from '../agentPrompt';
+import { PromptRegistry } from '../promptRegistry';
 import { ConversationHistorySummarizationPrompt, SummarizedConversationHistoryMetadata, SummarizedConversationHistoryPropsBuilder } from '../summarizedConversationHistory';
 
 suite('Agent Summarization', () => {
@@ -91,7 +92,8 @@ suite('Agent Summarization', () => {
 
 		let renderer;
 		if (promptType === 'Agent') {
-			const props: AgentPromptProps = baseProps;
+			const customizations = await PromptRegistry.resolveAllCustomizations(instaService, endpoint);
+			const props: AgentPromptProps = { ...baseProps, customizations };
 			renderer = PromptRenderer.create(instaService, endpoint, AgentPrompt, props);
 		} else {
 			const propsInfo = instaService.createInstance(SummarizedConversationHistoryPropsBuilder).getProps(baseProps);


OrenMe:
This looks great
If now this is all centralized in PromptRegistry it opens the way to add user/workspace customization to override built in system prompts 
Is this on the roadmap?

src/extension/intents/node/agentIntent.ts
```diff
@@ -65,6 +66,14 @@ export const getAgentTools = (instaService: IInstantiationService, request: vsco
 
 		const allowTools: Record<string, boolean> = {};
 
+		if (registryToolAllowlist) {
+			for (const [toolName, allowed] of Object.entries(registryToolAllowlist)) {
+				if (typeof allowed === 'boolean') {
+					allowTools[toolName] = allowed;
```
pwang347, line 72: since this is acting as an override (we allow false values), should this move to the bottom of this section?

src/extension/intents/node/agentIntent.ts
pwang347, line 105: are these going to use `toolAllowlist` customization later?

pwang347: nvm, just read the description

src/extension/prompts/node/agent/simpleSummarizedHistoryPrompt.tsx
```diff
@@ -84,7 +84,7 @@ export class SimpleSummarizedHistory extends PromptElement<SummarizedAgentHistor
 			return <ChunkTag name='conversation-summary' priority={priorityOverride}>
 				{entry.round.summary}
 				{this.props.endpoint.family === 'gpt-4.1' && <Tag name='reminderInstructions'>
-					<KeepGoingReminder modelFamily={this.props.endpoint.family} />
+					<DefaultOpenAIKeepGoingReminder />
```
pwang347, line 87: I assume this will also be cleaned up separately? (since this is currently importing openAI from common area)

src/extension/intents/node/agentIntent.ts
bhavyaus: will remove and fix this up in another pr.

src/extension/prompts/node/agent/simpleSummarizedHistoryPrompt.tsx
bhavyaus: I think this is ok since we check the family here and are intentiionally pulling in the openAI reminder.

bhavyaus:
>This looks great
If now this is all centralized in PromptRegistry it opens the way to add user/workspace customization to override built in system prompts
Is this on the roadmap?

@OrenMe we don't currently have any issues tracking this. Please create an issue.

src/extension/prompts/node/agent/agentPrompt.tsx
```diff
@@ -83,23 +84,20 @@ export class AgentPrompt extends PromptElement<AgentPromptProps> {
...
+			throw new Error('AgentPrompt requires customizations to be provided. Use PromptRegistry.resolveAllCustomizations() to resolve them.');
+		}
+		const instructions = await this.getSystemPrompt(customizations);
+		const CopilotIdentityRules = customizations.CopilotIdentityRulesClass;
+		const SafetyRules = customizations.SafetyRulesClass;
 
 		const omitBaseAgentInstructions = this.configurationService.getConfig(ConfigKey.Advanced.OmitBaseAgentInstructions);
 		const baseAgentInstructions = <>
 			<SystemMessage>
 				You are an expert AI programming assistant, working with a user in the VS Code editor.<br />
```
roblourens, line 98: I'd like to fully customize this section, including this sentence, how about moving it into the identity rules element or even just merging all this stuff into the main prompt? I tried this in https://github.com/microsoft/vscode-copilot-chat/pull/2156

src/extension/prompts/node/agent/agentPrompt.tsx
```diff
@@ -126,19 +128,23 @@ export class AgentPrompt extends PromptElement<AgentPromptProps> {
 					endpoint={this.props.endpoint}
 					tools={this.props.promptContext.tools?.availableTools}
 					enableCacheBreakpoints={this.props.enableCacheBreakpoints}
+					userQueryTagName={userQueryTagName}
```
roblourens, line 131: It seems like there's a lot of plumbing of passing around all these individual things via props. It made it hard to find where they are actually defined and used. I'd suggest either just resolve them at the place where they are needed instead of early on, or pass it around as one unit, or part of promptContext, so we don't have to add the noise of many individual props.

PR #2211, opened 2025-11-26
Add short-lived cache for documents and filter by language in linkifier.

DanielFabian:
## Fix: Optimize inline code symbol linkifier to avoid redundant file I/O
The inline code linkifier (InlineCodeSymbolLinkifier) was causing severe performance issues during LLM response streaming. For each backticked symbol (e.g., [TextDocument](http://_vscodecontentref_/1)), it would re-read every reference file from disk, resulting in O(N×M) file reads where N = symbol count and M = reference count.

### Changes:

- **Language gating**: Skip files entirely when tree-sitter doesn't support their language (.log, .md, .json, etc.) in symbolMatchesOnly mode—zero I/O for unsupported files.

- **Document cache**: Add a per-resolver documentCache so each file is read at most once per response, reducing file reads from O(N×M) to O(M).

**Impact**: Responses with many backticked symbols and large context files (especially non-code files like logs) should no longer hang the UI. Should help with https://github.com/microsoft/vscode/issues/275716

### DIFF
--- a/src/extension/linkify/vscode-node/findWord.ts
+++ b/src/extension/linkify/vscode-node/findWord.ts
@@ -49,6 +49,7 @@ export async function findWordInReferences(
 	word: string,
 	options: FindWordOptions,
 	token: CancellationToken,
+	documentCache?: Map<string, Promise<SimpleTextDocument | undefined>>,
 ): Promise<vscode.Location[]> {
 	const parserService = accessor.get(IParserService);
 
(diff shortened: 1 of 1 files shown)


DanielFabian:
@mjbvz any progress on reviewing this one?

vijayupadya:
s

dmitrivMS (approved):

PR #2229, opened 2025-11-26
add thinking budget header  in getExtraHeaders

bhavyaus:


bhavyaus:
>Should these be modelMetadata.requestHeaders instead of on extra headers? I want to deprecate getExtraHeaders at some point

I'll move that out when I get this as part of the modelMetadata from CAPI.

PR #2259, opened 2025-11-28
Fix experiment lookup

chrmarti:


chrmarti:
https://github.com/microsoft/vscode-copilot-chat/pull/2242

PR #2335, opened 2025-12-02
OpenAI prompt updates

dileepyavan:
Updating prompts and including the flags for new model.

### DIFF
--- a/src/platform/endpoint/common/chatModelCapabilities.ts
+++ b/src/platform/endpoint/common/chatModelCapabilities.ts
@@ -130,14 +130,14 @@ export function modelPrefersInstructionsAfterHistory(modelFamily: string) {
  * Model supports apply_patch as an edit tool.
  */
 export async function modelSupportsApplyPatch(model: LanguageModelChat | IChatEndpoint): Promise<boolean> {
-	return (model.family.includes('gpt') && !model.family.includes('gpt-4o')) || model.family === 'o4-mini' || await isHiddenModelA(model) || await isHiddenModelB(model) || await isHiddenModelC(model) || await isHiddenModelD(model) || await isVSCModelA(model) || await isVSCModelB(model);
+	return (model.family.includes('gpt') && !model.family.includes('gpt-4o')) || model.family === 'o4-mini' || await isHiddenModelA(model) || await isHiddenModelB(model) || await isHiddenModelC(model) || await isHiddenModelD(model) || await isVSCModelA(model) || await isVSCModelB(model) || model.family.includes('arctic-fox');
 }
 
 /**
  * Model prefers JSON notebook representation.
  */
 export function modelPrefersJsonNotebookRepresentation(model: LanguageModelChat | IChatEndpoint): boolean {
-	return (model.family.includes('gpt') && !model.family.includes('gpt-4o')) || model.family === 'o4-mini';
+	return (model.family.includes('gpt') && !model.family.includes('gpt-4o')) || model.family === 'o4-mini' || model.family.includes('arctic-fox');
 }
 
 /**


src/extension/prompts/node/panel/editorIntegrationRules.tsx
```diff
@@ -44,3 +44,53 @@ export class MathIntegrationRules extends PromptElement {
...
 }
+
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Tag } from '../base/tag';
+
+export class FileLinkificationInstructions extends PromptElement<{}> {
```
roblourens: I would actually recommend not using these instructions, which are new, because I believe they came with some code changes to the linkifier as well. We could use the simpler instructions that these replaced.

src/platform/endpoint/common/chatModelCapabilities.ts
```diff
@@ -37,7 +37,9 @@ const HIDDEN_MODEL_B_HASHES = [
 
 const HIDDEN_MODEL_C_HASHES = [
 	'57bc0aad677492da7a00731e3e411055b9828c6439f502fa5abd8fddb7a8a260',
-	'c5f9e7e93624823213aa93017d0d970ce3203d99dfcd616a0446f7bae2d8caf4'
+	'c5f9e7e93624823213aa93017d0d970ce3203d99dfcd616a0446f7bae2d8caf4',
+	'3358d5de6b9eac13ed03cfe6be37e45750f2ff16f588b24d8697837933def2bb',
```
roblourens: You don't have to hash these models. They aren't hashed in main. You can if you want.

src/extension/prompts/node/agent/openAIPrompts.tsx
```diff
@@ -641,12 +641,116 @@ class CodexStyleGPT5CodexPrompt extends PromptElement<DefaultAgentPromptProps> {
...
+	}
+}
+
 class OpenAIPromptResolver implements IAgentPrompt {
 
 	static readonly familyPrefixes = ['gpt', 'o4-mini', 'o3-mini', 'OpenAI'];
 
 	resolvePrompt(endpoint: IChatEndpoint): PromptConstructor | undefined {
 
+		if (endpoint.model.startsWith('gpt-5.1-codex')) {
```
roblourens, line 749: Don't you need to check arctic-fox here? How did you test this, does it work?

src/extension/prompts/node/panel/editorIntegrationRules.tsx
dileepyavan: I think this would impact the linkification PR: [Test: File path and line number linkification · Issue #279734 · microsoft/vscode](https://github.com/microsoft/vscode/issues/279734).

roblourens: This is the PR: https://github.com/microsoft/vscode-copilot-chat/pull/1803/files

you can see that it includes a lot of changes. So I don't think that you can extract just the prompt changes out of that and ship it to stable.

src/extension/prompts/node/agent/openAIPrompts.tsx
roblourens: Oh, this is 'model', not 'family'. We generally check 'family' because I don't know what the logic for model IDs is

src/extension/prompts/node/panel/editorIntegrationRules.tsx
dileepyavan: Updated

src/extension/prompts/node/agent/openAIPrompts.tsx
dileepyavan: for the new model arctic-fox, model is gpt-5.1-codex-max

roblourens (approved):

PR #2340, opened 2025-12-03
Make capabilities checks synchronous

kycutler:


src/platform/endpoint/node/modelMetadataFetcher.ts
```diff
@@ -140,6 +140,9 @@ export class ModelMetadataFetcher extends Disposable implements IModelMetadataFe
 			throw this._lastFetchError;
 		}
 
+		// Cache the model hashes for later use in detecting hidden models
+		await this.cacheModelHashes(resolvedModel);
```
roblourens: Good idea but how long will it take?

DonJayamanne (dismissed):

PR #2370, opened 2025-12-03
Update completions fallback model id

jeffreyhunter77:
We noticed in the CLS that the fallback model id was still `gpt-4o-copilot` and we're updating it to `gpt-41-copilot`. This makes the same change for completions from the chat extension and chat-lib.


jeffreyhunter77:
> @jeffreyhunter77 It seems to break one of the chat-lib tests. Could you take a look?

Thanks for catching that. Just added a fix.

PR #2379, opened 2025-12-04
Disable MCP result image URL for Gemini models

pwang347:
Seems like this was missed for Gemini 2.5 Pro as well
image

After fix
image

### DIFF
--- a/src/platform/endpoint/common/chatModelCapabilities.ts
+++ b/src/platform/endpoint/common/chatModelCapabilities.ts
@@ -147,7 +147,7 @@ export function modelShouldUseReplaceStringHealing(model: LanguageModelChat | IC
  * The model can accept image urls as the `image_url` parameter in mcp tool results.
  */
 export function modelCanUseMcpResultImageURL(model: LanguageModelChat | IChatEndpoint): boolean {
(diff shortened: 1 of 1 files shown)


src/platform/endpoint/common/chatModelCapabilities.ts
```diff
@@ -147,7 +147,7 @@ export function modelShouldUseReplaceStringHealing(model: LanguageModelChat | IC
  * The model can accept image urls as the `image_url` parameter in mcp tool results.
  */
 export function modelCanUseMcpResultImageURL(model: LanguageModelChat | IChatEndpoint): boolean {
-	return !model.family.startsWith('claude') && !model.family.startsWith('Anthropic') && !isHiddenModelE(model);
+	return !model.family.startsWith('claude') && !model.family.startsWith('Anthropic') && !model.family.startsWith('gemini') && !isHiddenModelE(model);
```
KevinBerthier, line 150: Is it not supposed to be : 

```suggestion
	return !model.family.startsWith('claude') && !model.family.startsWith('Anthropic') && !model.family.toLowerCase().startsWith('gemini') && !isHiddenModelE(model);
```

pwang347: yes, this should be lowercased to handle BYOK - fixed a few other references here https://github.com/microsoft/vscode-copilot-chat/pull/2529 thanks!

PR #2394, opened 2025-12-04
Display chat references when delegating to background

DonJayamanne:
Fixes https://github.com/microsoft/vscode/issues/281291


dmitrivMS:
Copilot has some interesting points...

DonJayamanne:
> Copilot has some interesting points...

Not required,

lramos15:
> Not required,

Some of these feel like very basic things that we should fix though

PR #2397, opened 2025-12-04
nes: joint: tweakable yielding strategies and yield-to-completions strategy

ulugbekna:
nes: simplify handling of NES endOfLife

nes: joint provider: yield to NES if last suggestion was NES and it agrees with users edits

Update src/extension/inlineEdits/vscode-node/jointInlineCompletionProvider.ts

Co-authored-by: Copilot <175728472+Copilot@users.noreply.github.com>

nes: dispose JointCompletionsProvider

nes: rework logic

this has a bug that completions gets triggered along with NES on cursor move -- it's very bad for completions because it has no debounce AND it would get cancellation on cursor move

nes: remove unused setting

nes: update lastNesSuggestion only if latest invocation

nes: fix imports

nes: remove unused param

nes: remove redundant check

nes: fix: correctly cancel completions when NES is picked


dmitrivMS:
Are Copilot comments worth addressing?

PR #2424, opened 2025-12-05
Fix async octokit calls to check for auth first

pwang347:
See https://github.com/microsoft/vscode-copilot-chat/pull/2423

### DIFF
--- a/package.json
+++ b/package.json
@@ -2503,11 +2503,6 @@
 						"default": true,
 						"description": "%github.copilot.config.customInstructionsInSystemMessage%"
 					},
-					"github.copilot.chat.customAgents.showOrganizationAndEnterpriseAgents": {
-						"type": "boolean",
-						"default": false,
-						"description": "%github.copilot.config.customAgents.showOrganizationAndEnterpriseAgents%"
-					},
 					"github.copilot.chat.agent.currentEditorContext.enabled": {
 						"type": "boolean",
 						"default": true,
@@ -3439,6 +3434,14 @@
 						"enum": [
 							"electron-fetch",
 							"node-fetch"
+						]
+					},
+					"github.copilot.chat.customAgents.showOrganizationAndEnterpriseAgents": {
+						"type": "boolean",
+						"default": false,
+						"description": "%github.copilot.config.customAgents.showOrganizationAndEnterpriseAgents%",
+						"tags": [
+							"experimental"
 						]
 					}
 				}


justschen (approved):

src/extension/agents/vscode-node/organizationAndEnterpriseAgentProvider.ts
```diff
@@ -148,19 +162,23 @@ export class OrganizationAndEnterpriseAgentProvider extends Disposable implement
...
 
 					const repoName = repos[0];
-					const agents = await this.octoKitService.getCustomAgents(org, repoName, internalOptions);
+					const agents = await this.runWithAuthCheck(() => this.octoKitService.getCustomAgents(org, repoName, internalOptions));
 					for (const agent of agents) {
 						agentsForOrg.set(agent.name, agent);
 					}
 					this.logService.trace(`[OrganizationAndEnterpriseAgentProvider] Fetched ${agents.length} agents from ${org} using repo ${repoName}`);
 				} catch (error) {
+					if (error instanceof Error && error.message === 'User is not signed in') {
```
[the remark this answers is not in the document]
aeschli: +1 to create a UserNotSignedInError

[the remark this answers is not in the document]
aeschli: +1 for using a UserNotSignedInError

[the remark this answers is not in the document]
pwang347: done!

pwang347:
> Should there be a listener for auth changes? How does that work?

This is currently handled by the activation mechanism in vscode-copilot-chat/src/extension/conversation/vscode-node/conversationFeature.ts which detects sign-in changes

src/extension/agents/vscode-node/organizationAndEnterpriseAgentProvider.ts
```diff
@@ -133,9 +149,9 @@ export class OrganizationAndEnterpriseAgentProvider extends Disposable implement
 			this.logService.trace(`[OrganizationAndEnterpriseAgentProvider] Found ${organizations.length} organizations: ${organizations.join(', ')}`);
 
 			// Convert VS Code API options to internal options
-			const internalOptions = options ? {
+			const internalOptions = {
```
TylerLeonhardt, line 152: intentional that options is not used anymore?

src/extension/agents/vscode-node/organizationAndEnterpriseAgentProvider.ts
```diff
@@ -29,31 +35,29 @@ export class OrganizationAndEnterpriseAgentProvider extends Disposable implement
 		@IFileSystemService private readonly fileSystem: IFileSystemService,
 	) {
 		super();
+
+		// Trigger async fetch to update cache. Note: this provider is re-created each time
+		// the user signs in, so this will re-fetch on sign-in. See logic in conversationFeature.ts.
+		this.fetchAndUpdateCache().catch(error => {
```
TylerLeonhardt, line 41: shouldn't this be run in some interval? When will the cache be updated?

src/extension/agents/vscode-node/organizationAndEnterpriseAgentProvider.ts
pwang347: Yes, I think it should be fine. Currently options are actually unused so if we need it later we can filter on top of whatever cached data we already have

lramos15 (approved):

src/extension/agents/vscode-node/organizationAndEnterpriseAgentProvider.ts
pwang347: It will currently update whenever the the sign-in status changes since that will trigger provider reactivation in `conversationFeature.ts`, I'm a bit worried about running in interval since that can lead to heavy load on GH API with just a lot of VS Code windows open. There is a pending GH issue to give us a better API for this, since the current API doesn't actually handle multiple orgs that well (ideally it should just be one API call for us so this is only temporary)

pwang347: so I don't think this feature will leave experimental state (default off) until we get that API, and we also should be less aggressive with the calls until then

TylerLeonhardt: I think it's worth asking the GitHub owners: "Should we be polling this API for changes? If so, at what frequency?" while making it clear that without polling users may go months without seeing updated lists of agents unless they reload VS Code.

TylerLeonhardt (approved):
I think that what you have in the candidate now is fine, since it's turning it off by default and marked as experimental.

src/extension/agents/vscode-node/organizationAndEnterpriseAgentProvider.ts
pwang347: sounds great! thank you

PR #2448, opened 2025-12-05
Enable no-unexternalized-strings in repo

mjbvz:
Enables the same `no-unexternalized-strings` with have in `vscode` in this repo. This make sure we have a more consistent style across repos and when generating edits


aeschli (approved):

hediet:
How did you fix the linting issues in all files? Is there a tool? (this caused some merge conflicts and I need to fix these quotes now in my branch)

mjbvz:
@hediet I created a custom eslint fix for the `code-no-unexternalized-strings.ts`. It's not enabled by default since this fix is often not the correct one, but you can turn it on by [changing the value for  `enableDoubleToSingleQuoteFixes` in that file](https://github.com/microsoft/vscode/blob/5eafa95b6d8a456be8db6f1f7ad16ca5ada23648/.eslint-plugin-local/code-no-unexternalized-strings.ts#L24)

PR #2501, opened 2025-12-08
Fix to ensure correct prompt is selected on final model update

dileepyavan:


src/platform/endpoint/common/chatModelCapabilities.ts
```diff
@@ -237,7 +237,7 @@ export function isGpt51Family(model: LanguageModelChat | IChatEndpoint | string
 	}
 
 	const family = typeof model === 'string' ? model : model.family;
-	return !!family.match(/^gpt-5\.\d+/i) || family === 'arctic-fox';
```
roblourens, line 240: I think this would reject `gpt-5.1-mini`

dileepyavan: oh, there is mini for 5.1, got it.

PR #2516, opened 2025-12-09
check for thinking data before disablingThinking

bhavyaus:
Fixes:
https://github.com/microsoft/vscode/issues/282267

### DIFF
--- a/src/extension/intents/node/toolCallingLoop.ts
+++ b/src/extension/intents/node/toolCallingLoop.ts
@@ -552,20 +552,29 @@ export abstract class ToolCallingLoop<TOptions extends IToolCallingLoopOptions =
 	}
 
 	public static messagesContainThinking(messages: Raw.ChatMessage[]): boolean {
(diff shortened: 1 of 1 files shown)


src/extension/intents/node/toolCallingLoop.ts
```diff
@@ -553,6 +551,23 @@ export abstract class ToolCallingLoop<TOptions extends IToolCallingLoopOptions =
 		});
 	}
 
+	public static messagesContainThinking(messages: Raw.ChatMessage[]): boolean {
+		return messages.some(m => {
```
roblourens: Is it right to check all messages, or do you need to check the asst messages after the last user message?

bhavyaus: Ah. Thanks. Yeah don't need all the messages. Just the last assistant message after user

bhavyaus: Fixed

PR #2570, opened 2025-12-12
Fix NESProvider in chat-lib

jeffreyhunter77:
The addition of the `IVSCodeExtensionContext` dependency in `InlineEditsModelService` causes `NESProvider` in chat-lib to break (and publishing to fail) because of the missing dependency that shows up in a failing test:

Screenshot 2025-12-08 at 11 14 44

Can we do this to get `NESProvider` working again (and with the option for library users to provide an implementation if they choose)?


chrmarti (approved):
LGTM, thanks!

PR #2571, opened 2025-12-12
present cloud session summary and collapse tool call thinking

rebornix:


src/extension/chatSessions/vscode-node/copilotCloudSessionContentBuilder.ts
```diff
@@ -287,6 +287,17 @@ export class ChatSessionContentBuilder {
 						toolPart.isError = true;
 						responseParts.push(toolPart);
 					}
+				} else {
```
rzhao271, line 290: Might need to reindent. This looks like it's on the same level as the else on Line 257.

joshspicer (approved):

src/extension/chatSessions/vscode-node/copilotCloudSessionContentBuilder.ts
joshspicer: might be the github diff, looks ok to me in vscode (formatter has no effect)

Screenshot 2025-12-12 at 2 08 16 PM

PR #2592, opened 2025-12-15
change prompt registration order. (#2578)

eleanorjboyd:
fixes https://github.com/microsoft/vscode/issues/283451


roblourens (approved):
It sounds like there's a dependency on the order of registration in the PromptRegistry, you can probably find a safer fix in main that doesn't depend on the order that registrations get resolved.

eleanorjboyd:
sounds good @AbdelrahmanAbouelenin can you work on a long term fix that can go into the main branch? This is merged into main and release but something better going forward. Thanks!

PR #2611, opened 2025-12-16
Reduce polling for cloud sessions provider

osortega:


src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
```diff
@@ -173,25 +174,70 @@ export class CopilotCloudSessionsProvider extends Disposable implements vscode.C
...
+		getRepoId(this._gitService).then(async repoId => {
+			const telemetryObj: {
+				intervalMs?: number;
+				hasHistoricalSessions?: boolean;
+				isEmptyWindow: boolean;
+			} = {
+				isEmptyWindow: !vscode.workspace.workspaceFolders || vscode.workspace.workspaceFolders.length === 0
+			};
+			if (repoId) {
+				const sessions = await this._octoKitService.getAllSessions(`${repoId.org}/${repoId.repo}`, false);
```
connor4312: If this call fails, we'll never do any polling at all. Maybe pop this into its own function and `.catch` errors

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
```diff
@@ -173,25 +174,70 @@ export class CopilotCloudSessionsProvider extends Disposable implements vscode.C
...
+					// TODO: handle no auth token case more gracefully
+					if (!this._authenticationService.permissiveGitHubSession) {
+						return;
+					}
+					const sessions = await this._octoKitService.getAllSessions(`${repoId.org}/${repoId.repo}`);
+					if (this.cachedSessionsSize !== sessions.length) {
+						this.refresh();
+					}
+				};
+				let interval = setInterval(intervalCallback, intervalMs);
```
connor4312: nit: `this._register(disposableInterval(` takes care of the cleanup for you

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
```diff
@@ -173,25 +174,70 @@ export class CopilotCloudSessionsProvider extends Disposable implements vscode.C
...
+				this._register(toDisposable(() => clearInterval(interval)));
+				this._register(vscode.window.onDidChangeWindowState((e) => {
+					if (!e.active) {
+						clearInterval(interval);
+					} else if (!interval) {
+						interval = setInterval(intervalCallback, intervalMs);
+						this._register(toDisposable(() => clearInterval(interval)));
+					}
+				}));
```
connor4312: If you want to do the posh experience I mentioned:


```suggestion
				let lastRefreshedAt = 0;
				const scheduler = this._register(new RunOnceScheduler(() => {
					lastRefreshedAt = Date.now();
					this.refresh();
				}, intervalMs));
				scheduler.schedule();

				this._register(vscode.window.onDidChangeWindowState((e) => {
					if (!e.active) {
						scheduler.cancel();
					} else if (!scheduler.isScheduled()) {
						scheduler.schedule(Math.max(0, intervalMs - (Date.now() - lastRefreshedAt)));
					}
				}));
```

something like that so we eagerly fetch/resume the interval once the window is active again

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
```diff
@@ -173,25 +174,77 @@ export class CopilotCloudSessionsProvider extends Disposable implements vscode.C
...
+					telemetryObj.hasHistoricalSessions = hasHistoricalSessions;
+					const schedulerCallback = async () => {
+						// TODO: handle no auth token case more gracefully
+						if (!this._authenticationService.permissiveGitHubSession) {
+							return;
+						}
+						const sessions = await this._octoKitService.getAllSessions(`${repoId.org}/${repoId.repo}`);
+						if (this.cachedSessionsSize !== sessions.length) {
+							this.refresh();
+						}
```
connor4312: Oh, after doing this we do still need to actually called `scheduer.schedule()` because it's not an interval automatically any more

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
osortega: Added a log just for visibility but I wouldn't want to add this function into a loop since it's an expensive operation

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
osortega: This is very cool, didn't know it existed, thanks!

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
```diff
@@ -173,25 +174,77 @@ export class CopilotCloudSessionsProvider extends Disposable implements vscode.C
...
+					scheduler.schedule();
+					this._register(vscode.window.onDidChangeWindowState((e) => {
+						if (!e.active) {
+							scheduler.cancel();
+						} else if (!scheduler.isScheduled()) {
+							scheduler.schedule(Math.max(0, intervalMs - (Date.now() - lastRefreshedAt)));
+						}
+					}));
+				} catch (e) {
+					this.logService.error(`Error during background refresh setup: ${e instanceof Error ? e.message : String(e)}`);
```
connor4312, line 199: This'll still not do any scheduling if that first call fails, though it does now log the error.  I think it should probably fail-open with the default 10 minute interval.

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
osortega: Good catch, added it

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
osortega: Good idea, let's default this to 10 minutes and log the error to telemetry as well

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
```diff
@@ -173,25 +174,85 @@ export class CopilotCloudSessionsProvider extends Disposable implements vscode.C
...
+					telemetryObj.error = e instanceof Error ? e.message : String(e);
+				}
+				telemetryObj.intervalMs = intervalMs;
+				telemetryObj.hasHistoricalSessions = hasHistoricalSessions;
+				const schedulerCallback = async () => {
+					// TODO: handle no auth token case more gracefully
+					if (!this._authenticationService.permissiveGitHubSession) {
+						return;
+					}
+					const sessions = await this._octoKitService.getAllSessions(`${repoId.org}/${repoId.repo}`);
```
connor4312: Last last thing, I promise 😅 
```suggestion
					try {
						const sessions = await this._octoKitService.getAllSessions(`${repoId.org}/${repoId.repo}`);
						if (this.cachedSessionsSize !== sessions.length) {
							this.refresh();
						}
					} catch (e) {
						logService.error(`Error during background refresh: ${e}`);
					}
					
					scheduler.schedule();
```

src/extension/chatSessions/vscode-node/copilotCloudSessionsProvider.ts
```diff
@@ -173,25 +174,93 @@ export class CopilotCloudSessionsProvider extends Disposable implements vscode.C
...
+					}
+					let sessions = [];
+					try {
+						sessions = await this._octoKitService.getAllSessions(`${repoId.org}/${repoId.repo}`);
+						if (this.cachedSessionsSize !== sessions.length) {
+							this.refresh();
+						}
+						if (this.cachedSessionsSize !== sessions.length) {
+							this.refresh();
+						}
```
connor4312: unintentional duplication?

osortega: @connor4312 yes lol, thanks for catching

PR #2613, opened 2025-12-16
Filter non-conversation utility requests from request log exports

zhichli:


src/extension/log/vscode-node/requestLogTree.ts
```diff
@@ -419,7 +419,28 @@ export class RequestLogTree extends Disposable implements IExtensionContribution
...
-			const exportableItems = allTreeItems.filter(item => item instanceof ChatPromptItem || (item instanceof ChatRequestItem && item.info.entry.debugName !== 'modelList'));
+			// Whitelist of exportable debugName patterns
+			const isExportable = (debugName: string): boolean => {
+				// Include main conversation requests (location/intent format)
+				if (debugName.includes('/')) {
+					return true;
+				}
+
+				// Include specific API wrapper calls and BYOK providers
+				const exportableNames = [
```
roblourens: What's being filtered out? I'm unsure about hardcoding these names. What do we have to do when we add a new BYOK provider? Something called `debugName` is not usually a load-bearing feature

zhichli: We're filtering out utility/infrastructure requests like `modelList` and `title` (which was floated to the top in #2577) – these don't represent actual conversation content.
The whitelist approach means:
- Adding new BYOK providers: Most already work through `copilotLanguageModelWrapper`, so they're automatically included. Only providers with custom implementations (like Gemini/Anthropic native) need explicit entries.
- Trade-off: Whitelist breaks if we add new exportable patterns, while blacklist breaks if we add new non-exportable utilities.

I chose whitelist because BYOK providers tend to reuse existing wrapper patterns, but I agree this might be ideal. Would you prefer:
- A blacklist of known non-exportable items (modelList, title, etc.)
- A more robust signal than debugName for determining exportability?

roblourens: I would do this with a property that declares that the request is a utility request (or a conversation request). Seems more maintainable than a list of names.

zhichli: Thanks for the suggestion - addressed.

roblourens (approved):
I like this, thanks!

PR #2618, opened 2025-12-17
Hook up prototype external ingest support

mjbvz:
Hook up first version of external ingest for initial testing. Not optimized much at all yet but basic flow works


mjbvz:
Please revert this change if it causes any issues with the build. The feature itself should be gated by settings, but the PR  pulls in a new wasm dependency through `node_modules` so there's a risk that something will break in release

I've tested but just calling out that a revert is the correct fix if I'm not around to do it myself

PR #2642, opened 2025-12-19
Copilot CLI - local background agent

lszomoru:
* Update worktree manager to store more properties of the worktree
* Update local background agent to commit at the end of each turn
* Since changes are committed there is no more Keep/Undo flow
* Apply action will use a patch file to migrate the changes

### DIFF
--- a/src/extension/agents/copilotcli/node/copilotcliSession.ts
+++ b/src/extension/agents/copilotcli/node/copilotcliSession.ts
@@ -250,9 +250,11 @@ export class CopilotCLISession extends DisposableStore implements ICopilotCLISes
 						throw new Error(`Unable to find repository for working directory ${workingDirectory}`);
 					}
 
+					this.logService.trace(`[CopilotCLISession] Generating commit message for working directory ${workingDirectory}. Repository state: ${JSON.stringify(repository.state)}`);
 					let message = await this.gitCommitMessageService.generateCommitMessage(repository, CancellationToken.None);
 					if (!message) {
 						// Fallback commit message
+						this.logService.error(`[CopilotCLISession] Unable to generate commit message for working directory ${workingDirectory}. Repository state: ${JSON.stringify(repository.state)}`);
 						message = `Copilot CLI session ${this.sessionId} changes`;
 					}
 


src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
```diff
@@ -144,23 +186,30 @@ export class CopilotCLIWorktreeManager {
...
-			workingDirectory = sessionWorktrees[sessionId];
-			if (workingDirectory) {
-				this._sessionWorktrees.set(sessionId, workingDirectory);
-			}
+	async saveWorktreeProperties(sessionId: string, properties: string | CopilotCLIWorktreeProperties): Promise<void> {
+		this._sessionWorktrees.set(sessionId, properties);
+
+		const sessionWorktreesProperties = this.extensionContext.globalState.get<Record<string, string | CopilotCLIWorktreeData>>(CopilotCLIWorktreeManager.COPILOT_CLI_SESSION_WORKTREE_MEMENTO_KEY, {});
+		sessionWorktreesProperties[sessionId] = { data: JSON.stringify(properties), version: 1 };
+		await this.extensionContext.globalState.update(CopilotCLIWorktreeManager.COPILOT_CLI_SESSION_WORKTREE_MEMENTO_KEY, sessionWorktreesProperties);
```
DonJayamanne, line 199: I think we need to review this global memento as well.
From what I know global mementos are carried across machines, & this data will grow over time if users don't delete/manage their sessions.
Meaning we'd have a lot of old data in this dictionay.

Perhaps we can clean that up after this PR.
Will think more about this over the next few days & will try to fix/re-work how/where we store this data. E.g. today we just archive/hide old sessions. If users create lots of bg sessions and ignore them, its actually a problem - lots of data that keeps getting accumulated

src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
```diff
@@ -133,7 +170,12 @@ export class CopilotCLIWorktreeManager {
 			const branch = `${branchPrefix}copilot-worktree-${new Date().toISOString().replace(/[:.]/g, '-').slice(0, 19)}`;
 			const worktreePath = await this.gitService.createWorktree(repository.rootUri, { branch });
 			if (worktreePath) {
-				return worktreePath;
+				return {
+					branchName: branch,
+					baseCommit: repository.headCommitHash!,
```
DonJayamanne: Use of `!` feels odd/hacky here `repository.headCommitHash!,`

src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
lszomoru: Great point! What't interesting here is that the changes are happening in a worktree, which is backed by a new branch which may be carried across machines, but the worktree path will probably not be present on other machines. We should probably think of the following:
1. Not sure if it makes sense for this data to travel across machines
2. Gracefully degrade in case the worktree path does not exist on the machine (ex: another machine, user manually deleted the worktree from the SCM view)
3. Delete the entry from the memento when a session is deleted
4. One time cleanup to clean the data that has been accumulated?

src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
lszomoru: Fixed.

src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
DonJayamanne: > Delete the entry from the memento when a session is deleted
One time cleanup to clean the data that has been accumulated?

Yeah,
We can leave this for later.
Today I already clean up such information after 7 days.
We already do the same for chat edits in core, we could extend this, however I fail to see any value in leaving artifacts around after its been unused for more than 7 days (its a good start)
Again, we can focus on this later.

> Not sure if it makes sense for this data to travel across machines

I don't think it needs to.
However if this will not grow, then its fine to leave this in global memento, its a simple/cheap approach.
I.e. if we continue to add more information, we might want to consider moving this into some file store, 
then again, when/what is that point in time when we store this in file?
My concern is as users create more background sessions, we'll have a lot more of these in memento

> Gracefully degrade in case the worktree path does not exist on the machine (ex: another machine, user manually deleted the worktree from the SCM view)

I'll leave that to you & perhaps @osortega

DonJayamanne: @lszomoru I'm planning on re-factoring some part of Copilot CLI files, to better handle delegation and the like.
& would also like to rename some of the files to `copilotCLI` instead of `copilotcli`over the next few days.

src/extension/chatSessions/vscode-node/copilotCLIChatSessionsContribution.ts
```diff
@@ -689,11 +761,18 @@ export class CopilotCLIChatSessionParticipant extends Disposable {
...
+		let worktreeProperties: CopilotCLIWorktreeProperties | undefined;
+
+		if (isNewSession || !isolationEnabled) {
+			if (isolationEnabled) {
+				worktreeProperties = await this.worktreeManager.createWorktree(stream);
+				workingDirectory = worktreeProperties ? Uri.file(worktreeProperties.worktreePath) : undefined;
+			} else {
+				workingDirectory = await this.copilotCLISDK.getDefaultWorkingDirectory();
+			}
+		}
```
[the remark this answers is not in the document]
lszomoru: @DonJayamanne, is this valid feedback? My code just uses the same condition that was already in place.

PR #2664, opened 2025-12-22
Increase diff computation timeout and add quitEarly handling

Hansade2005:
## Summary
This PR addresses timeout issues when computing diffs for large files in the VS Code Copilot chat extension.

## Changes
- Increased `maxComputationTimeMs` from 20 seconds to 60 seconds to allow more time for large file diffs
- Added `diff.quitEarly` check to provide user feedback if computation still times out
- Maintains protection against indefinite UI hangs while enabling larger diff operations

## Why This Change
- Large file edits were timing out at 20 seconds
- Setting timeout to 0 (infinite) could cause UI freezes
- 60-second timeout provides a good balance
- Graceful error handling informs users when diffs are too large  
image

### DIFF
--- a/src/extension/tools/node/editFileToolUtils.tsx
+++ b/src/extension/tools/node/editFileToolUtils.tsx
@@ -105,9 +105,13 @@ export async function formatDiffAsUnified(accessor: ServicesAccessor, uri: URI, 
 	const diffService = accessor.get(IDiffService);
 	const diff = await diffService.computeDiff(oldContent, newContent, {
 		ignoreTrimWhitespace: false,
-		maxComputationTimeMs: 20000,
+		maxComputationTimeMs: 60000, // Increased timeout to allow more time for large diffs
 		computeMoves: false,
 	});
+
+	if (diff.quitEarly) {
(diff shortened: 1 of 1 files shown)


src/extension/tools/node/editFileToolUtils.tsx
```diff
@@ -105,10 +105,14 @@ export async function formatDiffAsUnified(accessor: ServicesAccessor, uri: URI,
 	const diffService = accessor.get(IDiffService);
 	const diff = await diffService.computeDiff(oldContent, newContent, {
 		ignoreTrimWhitespace: false,
-		maxComputationTimeMs: 20000,
+		maxComputationTimeMs: 60000, // Increased timeout to allow more time for large diffs
```
connor4312, line 108: Do we actually see this being hit?

Hansade2005: yes it does hit the timeout when ai  performed complex modification

connor4312: Could you share an example (before/after) case where you hit this? 20s should already be very generous

PR #2683, opened 2025-12-26
Update linkification instructions to clarify valid formats

tamuratak:
The linkification instructions should be updated to clarify which formats are valid. `[file.ts#L10]` is not a valid format and should be removed; otherwise, agents tend to use it frequently.

image


tamuratak:
ping cc: @vijayupadya @dileepyavan

dmitrivMS (approved):

dmitrivMS:
@tamuratak There test failures, could you please take a look?

- - Valid formats: [file.ts](file.ts#L10) or [file.ts#L10] only. Invalid: ([file.ts#L10]) or [file.ts](file.ts)#L10
+ - Valid formats: [file.ts](file.ts#L10) only. Invalid: ([file.ts#L10]) or [file.ts](file.ts)#L10

PR #2689, opened 2025-12-30
Fix markdown content being split from thinking parts

pwang347:
Fixes https://github.com/microsoft/vscode/issues/278557


src/extension/conversation/vscode-node/languageModelAccess.ts
```diff
@@ -495,6 +495,12 @@ export class CopilotLanguageModelWrapper extends Disposable {
 	async provideLanguageModelResponse(endpoint: IChatEndpoint, messages: Array<vscode.LanguageModelChatMessage | vscode.LanguageModelChatMessage2>, options: vscode.ProvideLanguageModelChatResponseOptions, extensionId: string, progress: vscode.Progress<LMResponsePart>, token: vscode.CancellationToken): Promise<void> {
 		let thinkingActive = false;
 		const finishCallback: FinishedCallback = async (_text, index, delta): Promise<undefined> => {
+			if (delta.text) {
+				progress.report(new vscode.LanguageModelTextPart(delta.text));
```
roblourens: Thanks, this is getting closer to the underlying issue. But I see the response chunks in https://github.com/microsoft/vscode/pull/285365#discussion_r2653695319 and I'm confused. I think that in general the original code should be correct because usually the model thinks first, then produces text? Also this file is only used for access to models through the extension API, not for normal chat requests to Gemini through CAPI, is that what you're looking at?

pwang347: Okay, I think I figured it out - seems to be that we're not marking the thinking as encrypted so it's emitting a thinking part (with no visible string) which splits the markdown. Not super sure if there is more context to this as I noticed that in the streaming logic we don't actually use reasoning_opaque when parsing, but I think opaque reasoning should probably be considered encrypted thinking.

pwang347: ie., it was passing the `isEncryptedThinkingDelta` check here when it shouldn't
image

roblourens (approved):
Yes this seems like the root cause to me! Nice. The way that thinking content is handled right now is extremely confusing, we have model/provider-specific formats which are passed around internally and we have been through multiple generations of how it works. Thanks for working through it.

PR #2717, opened 2026-01-05
Fix media_type missing for gemini

pwang347:
If we can use heuristics to determine image media/mime type, we should send it to CAPI to avoid 400 errors. This doesn't fix issues with .gif formats though since Gemini doesn't support those. For that, we need a better story about how to handle general compatibility across models and specifying supported types.


src/extension/prompts/node/panel/image.tsx
```diff
@@ -6,6 +6,7 @@
 import { RequestType } from '@vscode/copilot-api';
 import * as l10n from '@vscode/l10n';
 import { Image as BaseImage, BasePromptElementProps, ChatResponseReferencePartStatusKind, PromptElement, PromptReference, PromptSizing, UserMessage } from '@vscode/prompt-tsx';
+import { ImageMediaType } from '@vscode/prompt-tsx/dist/base/output/rawTypes';
```
connor4312: Let's make sure to export this from the library so you don't need to import a subpath

src/extension/prompts/node/panel/image.tsx
```diff
@@ -55,14 +56,21 @@ export class Image extends PromptElement<ImageProps, unknown> {
...
-					const uri = await this.imageService.uploadChatImageAttachment(variable, this.props.variableName, getMimeType(imageSource) ?? 'image/png', githubToken);
+					const mimeType = getMimeType(imageSource) as ImageMediaType ?? imageMimeType;
+					const uri = await this.imageService.uploadChatImageAttachment(variable, this.props.variableName, mimeType, githubToken);
 					if (uri) {
 						imageSource = uri.toString();
+						imageMimeType = mimeType;
+						// TODO: figure out why gifs return 400 invalid request
+						if (imageMimeType === 'image/gif') {
+							imageMimeType = 'image/png';
+						}
```
connor4312: Won't this make us 'lie' about the media type of the image we're giving to the model? That seems like it would cause downstream issues for models that would depend on it...

pwang347: Fair point, let me check with GH folks to figure out why this is failing first (and if we can get a quick fix). Thanks!

pwang347: Removed the workaround for now, I think we will need to address the general compatibility across models in a separate issue since there are more scenarios to consider here

pwang347: Opened debt item here: https://github.com/microsoft/vscode/issues/286628

src/platform/networking/common/openai.ts
```diff
@@ -147,9 +148,20 @@ export function rawMessageToCAPI(message: Raw.ChatMessage[] | Raw.ChatMessage, c
...
 				part.text = part.text.trimEnd();
+			} else if (part.type === 'image_url' && Array.isArray(message.content) && i < message.content.length) {
+				const rawPart = message.content[i] as Raw.ChatCompletionContentPart;
+				if (rawPart?.type === Raw.ChatCompletionContentPartKind.Image && rawPart.imageUrl?.mediaType) {
+					// CAPI expects `media_type` instead of `mediaType`. This is only used for CAPI and not OpenAI.
+					const { mediaType, ...rawImageUrl } = rawPart.imageUrl;
+					(part.image_url as ChatCompletionContentPartImage.ImageURL & { media_type: string }) = {
+						...rawImageUrl,
+						media_type: mediaType
+					};
```
connor4312, line 163: can you confirm that this doesn't break OpenAI BYOK?

pwang347: yes, I tested this earlier (see following):
image

I don't believe this method is being called except for converting to CAPI format. Though I did notice interestingly that we also only use base64 images currently for OpenAI BYOK

PR #2733, opened 2026-01-07
Gemini BYOK: Handle invalid key

vijayupadya:
Fixes https://github.com/microsoft/vscode/issues/278350

Fix Gemini BYOK invalid API key handling so that users can recover after entering a bad/invalid key.

**Root cause:**
Gemini BYOK provider treated “invalid API key” failures from getAllModels as generic errors: it swallowed the error, kept the bad key in storage/memory, and never re‑prompted for a new key, so Manage Models would just silently return no models and leave you stuck with the invalid key.

### DIFF
--- a/src/extension/byok/vscode-node/test/geminiNativeProvider.spec.ts
+++ b/src/extension/byok/vscode-node/test/geminiNativeProvider.spec.ts
@@ -242,4 +242,51 @@ describe('GeminiNativeBYOKLMProvider', () => {
 		expect(models).toEqual([]);
 		expect(mockHandleAPIKeyUpdate).toHaveBeenCalled();
 	});
+
+	it('retries listing models after re-prompting with a valid API key', async () => {
+		const { GeminiNativeBYOKLMProvider } = await import('../geminiNativeProvider');
+		const genai = await import('@google/genai');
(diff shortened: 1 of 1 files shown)


Tyriar (approved):

src/extension/byok/vscode-node/test/geminiNativeProvider.spec.ts
```diff
@@ -217,4 +217,76 @@ describe('GeminiNativeBYOKLMProvider', () => {
 			tokenSource.token
 		)).rejects.toThrow(/No API key configured/i);
 	});
+
+	it('prompts for a new API key when listing models fails with an invalid key', async () => {
+		const { GeminiNativeBYOKLMProvider } = await import('../geminiNativeProvider');
+		const genai = await import('@google/genai');
+		const MockGoogleGenAI = genai.GoogleGenAI as unknown as { listModelsResult: AsyncIterable<any> };
+		// Simulate the models.list() call throwing an invalid API key error when iterated
+		MockGoogleGenAI.listModelsResult = (async function* () {
```
connor4312, line 226: Overwriting actual implementation methods in unit tests can cause weird stuff to happen later on. I suggest using dependency injection or, failing that, `sinon.stub` and `stub.restore`

PR #2745, opened 2026-01-08
Fix: Wrap Gemini function response arrays in object to prevent API error

bharatvansh:
## Description
Fixes an issue where the Gemini API rejects FunctionResponse payloads when the tool response is a JSON Array. The API requires the `response` field to be a JSON Object (Protobuf Struct).

When a tool returns a list (e.g. from GitHub MCP listing repos), the API throws the following error:
> "Invalid JSON payload received... Proto field is not repeating, cannot start list."

## Changes
- Modified `src/extension/byok/common/geminiMessageConverter.ts` to detect if `responsePayload` is an array and wrap it in an object: `{ result: [...] }`.
- Added regression test in `src/extension/byok/common/test/geminiMessageConverter.spec.ts`.

## Context / References
- **Root Cause**: The Gemini `FunctionResponse` expects the `response` field to be a `google.protobuf.Struct` (JSON Object), not a `ListValue`. Sending a top-level array causes a protocol buffer type mismatch.
- **Similar reports**:
    - [Gemini rejects tool JSON schema with array of types](https://discuss.ai.google.dev/t/gemini-rejects-tool-json-schema-with-array-of-types/4580) (Community discussion on similar Proto error)
    - [Firebase Genkit Issue #1393](https://github.com/firebase/genkit/issues/1393) (Documenting "Proto field is not repeating" error)

## Verification
- Added new unit test `should wrap array responses in an object` which passes.
- Ran existing tests with `npx vitest run src/extension/byok/common/test/geminiMessageConverter.spec.ts`.

### DIFF
--- a/src/extension/byok/vscode-node/anthropicProvider.ts
+++ b/src/extension/byok/vscode-node/anthropicProvider.ts
@@ -299,7 +299,7 @@ export class AnthropicLMProvider implements BYOKModelProvider<LanguageModelChatI
 				type: 'enabled',
 				budget_tokens: thinkingBudget
 			} : undefined,
-			context_management: contextManagement as any,
+			context_management: contextManagement as Anthropic.Beta.Messages.BetaContextManagementConfig | undefined,
 		};
 
 		const wrappedProgress = new RecordedProgress(progress);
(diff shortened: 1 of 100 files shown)


bharatvansh:
@vijayupadya Hi, can you please review this?

Tyriar (approved):

PR #2756, opened 2026-01-09
Custom chat render for agent skills

pwang347:
Depends on https://github.com/microsoft/vscode/pull/286839
Handles the invocation text changes

image
image

Fixes https://github.com/microsoft/vscode/issues/286213

### DIFF
--- a/src/extension/tools/node/readFileTool.tsx
+++ b/src/extension/tools/node/readFileTool.tsx
@@ -178,8 +178,8 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
 				const { skillName } = skillInfo;
 				if (this.customInstructionsService.isSkillMdFile(uri)) {
 					return {
-						invocationMessage: new MarkdownString(l10n.t`Reading skill ${formatUriForFileWidget(uri, { fileName: skillName })}`),
-						pastTenseMessage: new MarkdownString(l10n.t`Read skill ${formatUriForFileWidget(uri, { fileName: skillName })}`),
+						invocationMessage: new MarkdownString(l10n.t`Reading skill ${formatUriForFileWidget(uri, { vscodeLinkType: 'skill', linkText: skillName })}`),
+						pastTenseMessage: new MarkdownString(l10n.t`Read skill ${formatUriForFileWidget(uri, { vscodeLinkType: 'skill', linkText: skillName })}`),
 					};
 				} else {
 					return {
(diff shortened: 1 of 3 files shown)


src/extension/tools/node/test/readFile.spec.tsx
```diff
@@ -243,4 +246,148 @@ suite('ReadFile', () => {
 			expect(resultString).toContain('line 2');
 		});
 	});
+
+	suite('prepareInvocation', () => {
+		class MockCustomInstructionsService implements ICustomInstructionsService {
```
aeschli: Extract MockCustomInstructionsService to a separate module and reuse across tests

src/extension/tools/node/test/readFile.spec.tsx
```diff
@@ -243,4 +246,148 @@ suite('ReadFile', () => {
...
+
+			getAgentInstructions(): Promise<URI[]> {
+				return Promise.resolve([]);
+			}
+		}
+
+		test('should return "Loading/Loaded skill" message for skill files', async () => {
+			const testDoc = createTextDocumentData(URI.file('/workspace/test.skill.md'), 'skill content', 'markdown').document;
+
+			const services = createExtensionUnitTestingServices();
```
aeschli, line 281: Move the code that configures the test services up in 'beforeAll'.

src/extension/tools/node/readFileTool.tsx
```diff
@@ -171,7 +171,14 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
 		}
 
 		const { start, end } = getParamRanges(input, documentSnapshot);
+
 		if (start === 1 && end === documentSnapshot.lineCount) {
+			if (this.customInstructionsService.isSkillFile(uri)) {
+				return {
+					invocationMessage: new MarkdownString(l10n.t`Loading skill ${formatUriForFileWidget(uri)}`),
```
aeschli: I suggest we render: `Reading '${skillName}' skill ${formatUriForFileWidget(readLocation)}`. (or loading, not sure. IMO reading clarifies that we use the read tool.

aeschli: Note that skills can consist of multiple files. isSkill(..) will be true for all.
Rendering the file as a regular file bubble will help to know which file this is.

src/extension/tools/node/findTestsFilesTool.tsx
```diff
@@ -146,7 +146,7 @@ class FindTestFilesTool extends Disposable implements ICopilotTool<IFindTestFile
 	}
 
 	private formatURIs(uris: URI[]): string {
-		return uris.map(formatUriForFileWidget).join(', ');
+		return uris.map(uri => formatUriForFileWidget(uri)).join(', ');
```
aeschli, line 149: nitpick: that change is not really necessary

aeschli (approved):

src/extension/tools/node/readFileTool.tsx
aeschli: We would need to get the skill name. Could take it from the folder name, or really read the 'name' from SKILL.md

src/extension/tools/node/findTestsFilesTool.tsx
pwang347: it's needed because we add a new optional parameter, and TS isn't able to infer it properly

src/extension/tools/node/readFileTool.tsx
pwang347: I am thinking of actually limiting this message to SKILL.md, not the related assets. So something like:
image
image

Depending on whether SKILL.md is the one listed

PR #2779, opened 2026-01-10
Update token usage tracking for Anthropic models and implement summarization triggers based on token limits

bhavyaus:
### Summary

Use usage-based summarization strategy for when context editing is enabled for Anthropic models.
https://platform.claude.com/docs/en/build-with-claude/context-editing#client-side-compaction-sdk

> **Note:** This is a stopgap while we wait for the token counting API from CAPI.

### Key Changes

| File | Change |
|------|--------|
| **toolCallingLoop.ts** | Store `TokenUsageMetadata` (promptTokens, outputTokens) after successful fetch when context editing is enabled for Anthropic models. |
| **agentIntent.ts** | When context editing is enabled, check token usage from the current turn (if mid-turn) or previous turn's persisted metadata (if starting a new turn). If the estimated usage exceeds the model's max prompt tokens, trigger summarization. If no usage info is available, fall back to the prompt renderer throwing `BudgetExceededError` to trigger summarization. |

Known Issue:
- When you resume a chat conversation, we end triggering summarization if the previous usage was not cached. Should be fixed with moving to using token counting API
- I should also remove all previous toolResults before sending off to summary when we DO end up needing to kick off summarization because we will be over the budget.

### DIFF
--- a/src/extension/intents/node/agentIntent.ts
+++ b/src/extension/intents/node/agentIntent.ts
@@ -29,7 +29,7 @@ import { IInstantiationService, ServicesAccessor } from '../../../util/vs/platfo
 import { ICommandService } from '../../commands/node/commandService';
 import { Intent } from '../../common/constants';
 import { ChatVariablesCollection } from '../../prompt/common/chatVariablesCollection';
-import { RenderedUserMessageMetadata, TokenUsageMetadata } from '../../prompt/common/conversation';
+import { AnthropicTokenUsageMetadata, RenderedUserMessageMetadata } from '../../prompt/common/conversation';
 import { IBuildPromptContext } from '../../prompt/common/intents';
 import { getRequestedToolCallIterationLimit, IContinueOnErrorConfirmation } from '../../prompt/common/specialRequestTypes';
 import { IDefaultIntentRequestHandlerOptions } from '../../prompt/node/defaultIntentRequestHandler';
@@ -234,7 +234,7 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
 		if (summarizationEnabled && isAnthropicFamily(this.endpoint) && anthropicContextEditingEnabled) {
 			// First check current turn for token usage (from tool calling loop), then fall back to previous turn's result metadata
 			const currentTurn = promptContext.conversation?.getLatestTurn();
-			const currentTurnTokenUsage = currentTurn?.getMetadata(TokenUsageMetadata);
+			const currentTurnTokenUsage = currentTurn?.getMetadata(AnthropicTokenUsageMetadata);
 			const previousTurn = promptContext.history?.at(-1);
 
 			const promptTokens = currentTurnTokenUsage?.promptTokens ?? previousTurn?.resultMetadata?.promptTokens;
@@ -337,18 +337,7 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
 					e.metadata.getAll(ToolResultMetadata).forEach((metadata) => {
 						promptContext.toolCallResults![metadata.toolCallId] = metadata.result;
 					});
-					// Update props with new promptContext containing tool results
-					const updatedProps: AgentPromptProps = {
-						...props,
-						promptContext: {
-							...promptContext,
-							tools: promptContext.tools && {
-								...promptContext.tools,
-								toolReferences: this.stableToolReferences.filter((r) => r.name !== ToolName.Codebase),
-							}
-						}
-					};
-					result = await renderWithSummarization(`budget exceeded(${e.message})`, updatedProps);
+					result = await renderWithSummarization(`budget exceeded(${e.message})`);
 				} else {
 					throw e;
 				}
--- a/src/extension/prompt/common/conversation.ts
+++ b/src/extension/prompt/common/conversation.ts
@@ -396,7 +396,7 @@ export class GlobalContextMessageMetadata {
  * Stores prompt tokens and output tokens for each turn.
  * This metadata is used to trigger summarization when token usage exceeds thresholds.
  */
-export class TokenUsageMetadata {
+export class AnthropicTokenUsageMetadata {
 	constructor(
 		/** Total number of prompt input tokens */
 		readonly promptTokens: number,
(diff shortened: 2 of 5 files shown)


src/extension/prompt/common/conversation.ts
```diff
@@ -388,19 +392,17 @@ export class GlobalContextMessageMetadata {
 }
 
 /**
- * Metadata capturing context editing information from Anthropic Messages API.
- * When context editing clears tokens on a turn, this metadata is stored so that
- * subsequent turns can use the cleared token count to adjust their budget calculation.
+ * Metadata capturing token usage information from Anthropic Messages API.
```
roblourens, line 395: This should either take a model name so we can use it for other models or have a name that says that it's only for anthropic messages API

src/extension/intents/node/agentIntent.ts
```diff
@@ -218,11 +220,47 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
...
+		// For Anthropic models with context editing, check previous turn's token usage to determine budget
+		// 1. No token usage info (no prev turn) -> use normal safeBudget and let prompt rendering handle BudgetExceededError
+		// 2. Token usage + current turn > threshold -> throw BudgetExceededError to trigger summarization
+		// 3. Token usage + current turn < threshold -> use MAX_SAFE_INTEGER (no summarization needed)
+		let safeBudget: number = -1;
+		let shouldTriggerSummarize = false;
+		const budgetThreshold = Math.floor((baseBudget - toolTokens) * 0.85);
+
+		const anthropicContextEditingEnabled = isAnthropicContextEditingEnabled(this.configurationService, this.expService);
+		if (summarizationEnabled && isAnthropicFamily(this.endpoint) && anthropicContextEditingEnabled) {
```
roblourens, line 234: Is the point of this whole block just to make sure we use the token count from the model instead of the estimate prompt-tsx's tokenizer? That's unfortunate but I don't know a better way.

src/extension/prompt/common/conversation.ts
```diff
@@ -388,19 +392,17 @@ export class GlobalContextMessageMetadata {
...
-		/** Number of context edits applied */
-		readonly editCount: number,
-	) {
-		super();
-	}
+		/** Total number of prompt input tokens */
+		readonly promptTokens: number,
+		/** Number of output/completion tokens */
+		readonly outputTokens: number,
+	) { }
```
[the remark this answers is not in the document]
bhavyaus: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/vscode-copilot-chat/pull/2779#discussion_r2678094251)

src/extension/intents/node/agentIntent.ts
```diff
@@ -218,11 +220,47 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
...
+				const totalEstimatedTokens = (promptTokens + outputTokens) * 1.15;
+
+				if (totalEstimatedTokens > this.endpoint.modelMaxPromptTokens) {
+					// Will exceed budget - trigger summarization
+					shouldTriggerSummarize = true;
+					safeBudget = budgetThreshold; // Use normal budget for the summarization render
+					this.logService.debug(`AgentIntent: token usage exceeds threshold, will trigger summarization (promptTokens=${promptTokens}, outputTokens=${outputTokens}, total=${totalEstimatedTokens}, threshold=${budgetThreshold})`);
+				} else {
+					// Under budget - no summarization needed, use unlimited budget
+					safeBudget = Number.MAX_SAFE_INTEGER;
```
roblourens, line 254: Note that I don't really know what the impact of using this for things like attachments will be, it was used for that responses api setting but I never really teste that

src/extension/intents/node/agentIntent.ts
```diff
@@ -239,59 +277,81 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
...
 					});
-					try {
-						result = await renderer.render(progress, token);
-					} catch (e) {
-						if (e instanceof BudgetExceededError) {
-							this.logService.error(e, `[Agent] final render fallback failed due to budget exceeded`);
-							const maxTokens = this.endpoint.modelMaxPromptTokens;
-							throw new Error(`Unable to build prompt, modelMaxPromptTokens=${maxTokens} (${e.message})`);
+					// Update props with new promptContext containing tool results
+					const updatedProps: AgentPromptProps = {
```
roblourens: Is this supposed to be here? I think we already did this above?

src/extension/intents/node/agentIntent.ts
bhavyaus: Yes. I tried to wire this in as a service that handles it per model but it was too big of a change. I wanted to keep this pr specific to changing up the strategy.

src/extension/intents/node/agentIntent.ts
bhavyaus: From my limited testing with the responses API, I think this is a strategy to prevent our prompt-tsx tokenizer from removing nodes when trying to fit inside the budget. So I reused it here because context editing is happening on the server side anyway.

src/extension/prompt/common/conversation.ts
bhavyaus: Renamed to AnthropicTokenUsageMetadata

src/extension/intents/node/agentIntent.ts
bhavyaus: Fixed! Thanks!

src/extension/intents/node/agentIntent.ts
bhavyaus: Created: https://github.com/microsoft/vscode/issues/287030 to try to clean this up as debt

roblourens: The main thing for me is that rendering all the history should not go through prompt-tsx at all, only the new stuff. That will be a large refactoring some day. And then I believe Connor wants to have proper token counting for different types of models too.

src/extension/intents/node/agentIntent.ts
roblourens: My worry was that we size some things as a portion of the available context window, eg 

https://github.com/microsoft/vscode-copilot-chat/blob/fe1ab3233a98f7ec5dbb2e37ed23cc5e768eebf7/src/extension/prompts/node/agent/agentPrompt.tsx#L358-L360

and I don't know what that code does when the budget is very large.

PR #2793, opened 2026-01-12
Show offline view for all remaining errors

chrmarti:
https://github.com/microsoft/vscode/issues/287027


TylerLeonhardt (commented):
This doesn't feel right. Token endpoint can return 403s that are not due to offline behavior that may be masked by this.

PR #2795, opened 2026-01-12
Gemini function calling mode exp

vijayupadya:
Fixes https://github.com/microsoft/vscode-internalbacklog/issues/6538
Enabling Passing function calling mode option for Gemini3 in Exp


package.json
```diff
@@ -2656,6 +2650,20 @@
...
 					},
+					"github.copilot.chat.gemini.functionCallingMode": {
+						"type": "string",
+						"enum": [
+							"auto",
+							"none",
+							"required",
+							"validated"
+						],
+						"markdownDescription": "Function calling mode for Gemini tool calls. Options: auto (let model decide), none (disable tool calls), required (force at least one tool call), validated (use validated mode for better accuracy).",
```
connor4312: I think the code is fine but is this going to be a setting that we keep around after EXP? As a user I don't feel like this is something I should have to configure (and setting it to `none` would break stuff)

vijayupadya: @connor4312 made the setting internal

PR #2802, opened 2026-01-12
Make savePrompt create a file in proper folder and use kebab-case

TechWatching:
This prompt was creating the new prompt in an undefined file instead of creating it in the workspace github prompt folder. It was also using camel case instead of kebab case like in Awesome Copilot repository.


aeschli:
invalid, the savePrompt is now create-prompt

PR #2839, opened 2026-01-14
Support parallel subagents

roblourens:
Fix microsoft/vscode#286606

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
 		"authLearnMore",
 		"testObserver",
 		"aiTextSearchProvider@2",
-		"chatParticipantPrivate@12",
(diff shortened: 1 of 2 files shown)


osortega (approved):

Tyriar (dismissed):

htekdev:
What else is left for this? How does this show up in the UI?

xihuai18:
Great, when to support different models for different subagents? For example, using gpt and opus for cross-verification.

Sjoerd-Bo3:
> Great, when to support different models for different subagents? For example, using gpt and opus for cross-verification.

Sounds like a good plan, maybe make a new issue for this? @roblourens

PR #2879, opened 2026-01-15
logService: feat: allow creating sub-loggers

ulugbekna:


src/platform/log/common/logService.ts
```diff
@@ -91,6 +91,17 @@ export interface ILogger {
...
+	/**
+	 * Creates a sub-logger with a topic prefix. All messages logged through
+	 * the sub-logger will be prefixed with the topic, e.g., `[Topic] message`.
+	 *
+	 * Sub-loggers can be nested, and the prefixes will accumulate,
+	 * e.g., `[Parent][Child] message`.
+	 *
+	 * @param topic The topic name or array of topic names to prefix messages with
+	 */
+	createSubLogger(topic: string | readonly string[]): ILogger;
```
Tyriar, line 104: I do something similar in `ITerminalLogService`. @sandy081 does this look like a good approach to you?

Tyriar: https://github.com/microsoft/vscode/blob/c36e2b89d9171f212b34491dd9b61eb72abbfb04/src/vs/platform/terminal/common/terminalLogService.ts#L15-L59

Tyriar (changes requested):
See mine and Copilot's comments. I think this should get approval from @sandy081 too

ulugbekna:
@Tyriar Chat's log service was authored by @lramos15

Tyriar:
Just realized this is in chat not core. Still we should try align to keep working in the codebases consistent.

Red CI

benibenj (approved):

src/platform/log/common/logService.ts
sandy081: What we support in core are loggers with log file created for each topic. Therefore one can view each topic separately or can be grouped under a group.

These seem to be recursive prefix (sub prefix) loggers which are just appending a sub topic as prefix. I think we do not want multiple logger files here. So this seems to be fine

PR #2914, opened 2026-01-16
Fix timing issue with flags

chrmarti:
These flags are suspicious because they are inside the for await loop over chunks and the chunking is arbitrary and depends on buffers and network timing.

### DIFF
--- a/src/platform/networking/node/stream.ts
+++ b/src/platform/networking/node/stream.ts
@@ -312,6 +312,13 @@ export class SSEProcessor {
 		let extraData = '';
 		// This flag is set when at least for one solution we finished early (via `finishedCb`).
 		let hadEarlyFinishedSolution = false;
+
+		// The platform agent can return a 'function_call' finish_reason, which isn't a real function call
+		// but is echoing internal function call messages back to us. So don't treat them as real function calls
+		// if we received more data after that
+		let allowCompletingSolution = true;
+		let thinkingFound = false;
+
 		// Iterate over arbitrarily sized chunks coming in from the network.
 		for await (const chunk of this.body) {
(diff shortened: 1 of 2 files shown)


src/platform/networking/node/stream.ts
chrmarti, line 549: @justschen Copilot thinks the thinking flag needs to be reset here, but I have no idea:

Image

justschen (approved):
i think this is fine, looks like it was set false multiple times.

the only thing we really do care about is `thinkingFound ||= !!(thinkingDelta?.text || thinkingDelta?.id);`

src/platform/networking/node/stream.ts
```diff
@@ -324,20 +331,13 @@ export class SSEProcessor {
...
-			let thinkingFound = false;
-
 			for (const dataLine of dataLines) {
 				// Lines which start with a `:` are SSE Comments per the spec and can be ignored
 				if (dataLine.startsWith(':')) {
 					continue;
 				}
 				const lineWithoutData = dataLine.slice('data:'.length).trim();
 				if (lineWithoutData === '[DONE]') {
-					thinkingFound = false;
```
justschen, line 340: well actually, i think this is necessary as it goes thru the loop again

src/platform/networking/node/stream.ts
justschen, line 547: copilot wants to arbitrarily set `thinkingFound` to false here when it's not necessarily going to be false

src/platform/networking/node/stream.ts
chrmarti: Copilot argued that since we return from the function below it's not used anymore.

PR #2916, opened 2026-01-16
Add hasCursorJump field in correlationId

alexdima:


ulugbekna:
fixed in https://github.com/microsoft/vscode-copilot-chat/pull/2923

PR #2918, opened 2026-01-16
Require recent NES trigger before triggering on document switch

alexdima:
Fixes microsoft/vscode#284225

Previously, switching documents could trigger NES even when NES hadn't been recently active, causing unexpected suggestions to appear. This change adds a check in `_maybeTriggerOnDocumentSwitch` to require that NES was triggered within the last 10 seconds before allowing a document switch to trigger a new NES request.

This aligns the document switch behavior with the existing same-document behavior, which already requires a recent NES trigger via the `hasRecentTrigger` check.


alexdima:
fyi @ulugbekna

dmitrivMS (approved):

benibenj (approved):

Tyriar (approved):

PR #2934, opened 2026-01-16
Hook up PermissionMode to an option

TylerLeonhardt:
CC extenion has a similar toggle, let's replicate it for now so that at least I can use Plan mode.

The other two options feel kinda silly in 2026... That's why I made Edit automatically the default since that's what we allow in agent mode today.

So, I think this needs some reworking later... but for now at least we're consistent.


dmitrivMS:
@TylerLeonhardt does Copilot have good points?

PR #2958, opened 2026-01-18
fix: flush linkifier before thinking blocks in Messages API

bhavyaus:


roblourens (approved):
Why do you have to do this for thinking parts, doesn't sending the thinking content itself flush the linkifier?

bhavyaus:
> Why do you have to do this for thinking parts, doesn't sending the thinking content itself flush the linkifier?

The thinking content doesn't flush the linkifier because thinkingProgress is enqueued with flush: false                        
 We still need to flush once at the transition point when the model stops outputting text and starts thinking.

roblourens:
@mjbvz @vijayupadya something still seems really wrong with the linkifier when we have to send random spaces from deep in some message parsing code to get content to show up

mjbvz:
Hm yes this seems off. Feels like tools calls should automatically flush the stream instead of requiring the space

@vijayupadya Want to look into this?

PR #2966, opened 2026-01-19
Retrieve list of MCP servers from VS Code API

DonJayamanne:


src/util/common/test/shims/chatTypes.ts
```diff
@@ -532,4 +532,102 @@ export class CustomAgentChatResource implements vscode.CustomAgentChatResource {
...
 	}
-}
\ No newline at end of file
+}
+
+
+/**
+	 * McpStdioServerDefinition represents an MCP server available by running
+	 * a local process and operating on its stdin and stdout streams. The process
+	 * will be spawned as a child process of the extension host and by default
```
dmitrivMS: formatting can be better here...

PR #2978, opened 2026-01-19
Remove chat_enabled organization policy

benibenj:
```Copilot Generated Description:``` Eliminate the chat_enabled property from various classes and interfaces, simplifying the token management and telemetry reporting processes. Adjust related logic to reflect the removal of chat-related checks.

Related: https://github.com/microsoft/vscode-internalbacklog/issues/6607#issue-3829229263

// cc @TylerLeonhardt @lramos15

### CODE REVIEW

File: src/platform/authentication/common/copilotToken.ts [hunk 1]
```diff
@@ -285,8 +281,6 @@ export interface TokenEnvelope {
 	// Feature flags
 	/** Whether client-side indexing for Blackbird is enabled. */
 	blackbird_clientside_indexing: boolean;
-	/** Whether chat features are enabled. */
-	chat_enabled: boolean;
```
TylerLeonhardt, line 289: This is a representation of what CAPI gives us.

Is CAPI no longer sending this?

If so, I'd rather we keep it

PR #2997, opened 2026-01-20
Skip XtabProvider requests when prediction exceeds token limit

alexdima:
Fixes microsoft/vscode-internalbacklog#6564

Adds a client-side check to skip requests when the prediction/speculation
content exceeds languageContext.maxTokens (default 2000). This prevents
errors of the form "User-provided speculation (X tokens) is longer than
max_tokens (2048)" which occur when large documents with static data
are sent for speculative decoding.


ulugbekna:
closing because incorrect as pointed out in the Copilot comment above. Referenced the PR with how I think this should be fixed

PR #3019, opened 2026-01-20
request logger debug view grouping and ordering

amunger:
fix https://github.com/microsoft/vscode-copilot-evaluation/issues/1920

group all activity from a request together, even when other copilot activity happens midway through. Although subagents still have a separate entry

also added some tests and docs to help the agent make changes.

image

### DIFF
--- a/src/extension/log/vscode-node/requestLogTree.ts
+++ b/src/extension/log/vscode-node/requestLogTree.ts
@@ -648,23 +648,17 @@ class ChatRequestProvider extends Disposable implements vscode.TreeDataProvider<
 		}
 
 		// Second pass: collect prompt items and apply flattening/promotion logic
-		const rootPrompts: ChatPromptItem[] = [];
 		for (const [, promptItem] of tokenToPromptItem) {
 			// Apply flattening and promotion logic
 			if (promptItem.token.flattenSingleChild && promptItem.children.length === 1 && !promptItem.hasSeen) {
+				// Flattened: add the single child directly to result
 				result.push(promptItem.children[0]);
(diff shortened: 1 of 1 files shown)


docs/request-logger-architecture.md
```diff
@@ -0,0 +1,173 @@
+# Request Logger Architecture
+
+## Overview
+
+The Request Logger system is responsible for tracking and displaying all AI-related requests, tool calls, and prompt traces made by the GitHub Copilot Chat extension. It consists of two main components in separate files:
```
DonJayamanne, line 5: @amunger  Do we need this document? I.e. did you notice any difference in the copilot chat responses.

I feel that this is a small component and having an architecture doc for this is perhaps unnecessary or overkill.
Eg. relationship between logger and tree is standard VS Code tree view architecture.

Saying this as I was able to get copilot to make changes in this space without having such a doc.  Given its a small component, i feel the model can easily understand this.

ulugbekna (approved):
i didn't look very closely but hope it doesn't break logs for NES :D

docs/request-logger-architecture.md
amunger: It seemed to help, but it's pretty hard to say. But I wouldn't mind if it was removed if its getting in the way.

ulugbekna:
I'm sorry, breaking of temporal ordering is a deal breaker for ghostText/NES -- I expected that the PR would preserve temporal ordering of the first parent tree entry and keep appending to it. Should've reviewed the PR, sorry, I have to revert the PR

PR #3038, opened 2026-01-21
make memory tool model specific

bhavyaus:


OrenMe:
May I ask why make it model specific? Does memory work best only with those models?
Will they reference the tool directly? Cause afaik today there is no mention of it in the system instructions of any model in copilot chat extension

PR #3052, opened 2026-01-22
Support reading files outside of the current workspace

sbatten:
Enhance the functionality to allow reading files and directories that are outside the current workspace. Implement confirmation prompts for users when attempting to access external files or directories. This change improves user experience by providing clarity and control over file access.

### DIFF
--- a/src/extension/tools/node/readFileTool.tsx
+++ b/src/extension/tools/node/readFileTool.tsx
@@ -18,8 +18,7 @@ import { IExperimentationService } from '../../../platform/telemetry/common/null
 import { ITelemetryService } from '../../../platform/telemetry/common/telemetry';
 import { IWorkspaceService } from '../../../platform/workspace/common/workspaceService';
 import { clamp } from '../../../util/vs/base/common/numbers';
-import { dirname } from '../../../util/vs/base/common/path';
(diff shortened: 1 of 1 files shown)


src/extension/tools/node/readFileTool.tsx
```diff
@@ -164,6 +165,34 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
...
+				accessor => isFileExternalAndNeedsConfirmation(accessor, uri!)
+			);
+
+			if (isExternal) {
+				// Still check content exclusion (copilot ignore)
+				await this.instantiationService.invokeFunction(
+					accessor => assertFileNotContentExcluded(accessor, uri!)
+				);
+
+				const folderUri = uri.with({ path: dirname(uri.path) });
```
connor4312: Should use the dirname version from resources.ts that operates on URIs

connor4312 (changes requested):
We should also apply this treatment to tools like listDirectory and fileSearch. For adopting this there I suggest we wrap stuff up a bit more tightly as we do in `createEditConfirmation`

sbatten:
@connor4312 I already did listDir, but I missed search

connor4312 (dismissed):

PR #3128, opened 2026-01-23
Set validated flag by default to gemini tool calls

vijayupadya:
Fixes https://github.com/microsoft/vscode/issues/289874

Exp was successful, set the 'validated' flag by default.


vijayupadya:
cc @pwang347

PR #3137, opened 2026-01-23
Explore manual summarization

lramos15:
Ideally we want users to be able to trigger summarization themselves whenever they want to keep context down. This explores adding that feature. We can then add this to the context window widget.

@roblourens as the summarization expert I would appreciate your thoughts on this approach as I ideally would like this to be as close to the agentic summarization as possible and reuse as much code as possible

### DIFF
--- a/package.json
+++ b/package.json
@@ -1459,6 +1459,10 @@
 				"when": "config.chat.agent.enabled",
 				"commands": [
 					{
+						"name": "compact",
+						"description": "%copilot.agent.compact.description%"
+					},
+					{
 						"name": "error",
 						"description": "Make a model request which will result in an error",
 						"when": "github.copilot.chat.debug"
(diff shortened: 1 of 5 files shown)


roblourens:
This looks like a separate implementation of summarization? Let's trigger the existing one

roblourens:
Maybe you can get it to reuse more of `ConversationHistorySummarizer`. It looks like this mostly duplicates that.

The other idea I had was to go through the normal agent loop but trigger summarization here: https://github.com/microsoft/vscode-copilot-chat/blob/607a59d3594dc5c351e83059956bba57e29844d4/src/extension/intents/node/agentIntent.ts#L232 then we would return some flag telling the tool calling loop to bail out of the agent loop without sending a request. Maybe it's more complicated. I don't really know how this should look though, what is in the ChatResult. And we need to make sure that we don't add an empty turn to the history after triggering this manually (in both cases). This is all just so complicated.

roblourens:
Handled in anothe rPR

PR #3144, opened 2026-01-24
Add switch agent tool

sbatten:
Introduce a new tool to switch to a different agent mode, currently supporting only the 'Plan' agent. This allows for better handling of complex tasks that require careful planning.

### DIFF
--- a/src/extension/tools/vscode-node/switchAgentTool.ts
+++ b/src/extension/tools/vscode-node/switchAgentTool.ts
@@ -1,0 +1,52 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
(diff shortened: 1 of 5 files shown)


src/extension/tools/vscode-node/switchAgentTool.ts
```diff
@@ -0,0 +1,52 @@
...
+	async invoke(options: vscode.LanguageModelToolInvocationOptions<ISwitchAgentParams>, token: CancellationToken): Promise<vscode.LanguageModelToolResult> {
+		const { agentName } = options.input;
+
+		// Only 'Plan' is supported
+		if (agentName !== 'Plan') {
+			throw new Error(vscode.l10n.t('Only "Plan" agent is supported'));
+		}
+
+		// Execute command to switch agent
+		await vscode.commands.executeCommand('workbench.action.chat.toggleAgentMode', {
```
roblourens, line 28: You could implement this in core, then we don't have to deal with commands

PR #3177, opened 2026-01-26
Refactor PromptPathRepresentationService to support WSL paths and improve file path resolution

bryanchen-d:
Fix https://github.com/microsoft/vscode/issues/289166

### DIFF
--- a/src/platform/prompts/common/promptPathRepresentationService.ts
+++ b/src/platform/prompts/common/promptPathRepresentationService.ts
@@ -9,6 +9,7 @@ import { hasDriveLetter } from '../../../util/vs/base/common/extpath';
 import { Schemas } from '../../../util/vs/base/common/network';
 import { isWindows } from '../../../util/vs/base/common/platform';
 import { URI } from '../../../util/vs/base/common/uri';
+import { IWorkspaceService } from '../../workspace/common/workspaceService';
 
 export const IPromptPathRepresentationService = createServiceIdentifier<IPromptPathRepresentationService>('IPromptPathRepresentationService');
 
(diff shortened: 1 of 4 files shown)


src/platform/prompts/common/promptPathRepresentationService.ts
```diff
@@ -34,18 +35,27 @@ export interface IPromptPathRepresentationService {
...
+	) { }
+
 	protected isWindows() {
 		return isWindows;
 	}
 
 	getFilePath(uri: Uri): string {
-		if (uri.scheme === Schemas.file || uri.scheme === Schemas.vscodeRemote) {
+		if (uri.scheme === Schemas.vscodeRemote) {
+			// Remote filesystems (WSL, SSH, etc.) use POSIX paths.
```
connor4312, line 54: I think this would cause issues for Windows remotes, e.g. SSHing to a Windows box?

src/platform/prompts/common/promptPathRepresentationService.ts
```diff
@@ -60,6 +70,18 @@ export class PromptPathRepresentationService implements IPromptPathRepresentatio
 	 * @returns The resolved URI or undefined if filepath does not look like a file path or URI.
 	 */
 	resolveFilePath(filepath: string, predominantScheme = Schemas.file): Uri | undefined {
+		// Check if workspace uses vscode-remote scheme (WSL, SSH, etc.)
+		// If so, construct URI with remote scheme and authority instead of using the default logic.
+		const folders = this.workspaceService.getWorkspaceFolders();
+		if (folders.length > 0 && folders[0].scheme === Schemas.vscodeRemote && filepath.startsWith('/')) {
+			return URI.from({
+				scheme: Schemas.vscodeRemote,
```
connor4312, line 78: I'm unclear why this is needed -- Copilot Chat is defined to run on the remote workspace, and so it should not need to construct vscode-remote schemes directly -- this should be handled by the URI transformation in the exthost protocol layer.

bryanchen-d: You are right, the extension should be running on the WSL side so this is not needed.

PR #3205, opened 2026-01-27
Search subagent - set configurable exp variables

24anisha:
Rather than making the model endpoint and the tool call limit deterministic for the search subagent, make them configurable via exp settings.

Defaults to: max tools calls = 4 and subagent endpoint = main agent model endpoint

### DIFF
--- a/src/extension/prompt/node/searchSubagentToolCallingLoop.ts
+++ b/src/extension/prompt/node/searchSubagentToolCallingLoop.ts
@@ -68,16 +68,19 @@ export class SearchSubagentToolCallingLoop extends ToolCallingLoop<ISearchSubage
 	private async getEndpoint(request: ChatRequest) {
 		const modelName = this._configurationService.getExperimentBasedConfig(ConfigKey.Advanced.SearchSubagentModel, this._experimentationService) as ChatEndpointFamily | undefined;
 
-		let endpoint;
-		if (!modelName) {
-			endpoint = await this.endpointProvider.getChatEndpoint(this.options.request);
-		} else {
-			endpoint = await this.endpointProvider.getChatEndpoint(modelName);
-			if (!endpoint.supportsToolCalls) {
-				endpoint = await this.endpointProvider.getChatEndpoint(this.options.request);
+		if (modelName) {
+			try {
+				// Try to get the specified model
+				return await this.endpointProvider.getChatEndpoint(modelName);
+			} catch (error) {
+				// Model not available or doesn't support tool calls, fallback to main agent
+				this._logService.warn(`Failed to get model ${modelName}, falling back to main agent endpoint: ${error}`);
+				return await this.endpointProvider.getChatEndpoint(this.options.request);
 			}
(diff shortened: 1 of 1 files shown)


src/extension/prompt/node/searchSubagentToolCallingLoop.ts
```diff
@@ -62,22 +62,32 @@ export class SearchSubagentToolCallingLoop extends ToolCallingLoop<ISearchSubage
 		return context;
 	}
 
+	// get the endpoint to use for the search subagent
```
roblourens: Please use `/**` tsdoc style comments

src/extension/prompt/node/searchSubagentToolCallingLoop.ts
```diff
@@ -62,22 +62,32 @@ export class SearchSubagentToolCallingLoop extends ToolCallingLoop<ISearchSubage
...
 	private async getEndpoint(request: ChatRequest) {
-		let endpoint = await this.endpointProvider.getChatEndpoint(this.options.request);
-		if (!endpoint.supportsToolCalls) {
-			endpoint = await this.endpointProvider.getChatEndpoint('gpt-4.1');
+		const modelName = this._configurationService.getExperimentBasedConfig(ConfigKey.Advanced.SearchSubagentModel, this._experimentationService) as ChatEndpointFamily | undefined;
+
+		let endpoint;
+		if (!modelName) {
+			endpoint = await this.endpointProvider.getChatEndpoint(this.options.request);
```
roblourens: `getChatEndpoint` throws if you ask for a model that it doesn't know about. If you set up an experiment to use haiku, and a user is in an org where Claude is disabled, what do you want to happen? Fall back to a different model? Or should the searchSubagent be disabled when the model you want to use isn't supported?

I think this is why `getChatEndpoint` isn't set up to easily ask for arbitrary models, because we generally don't want to encourage people to write code which does that. Your scenario is fine as long as you handle these sorts of cases.

24anisha: ideally, if it can't get the endpoint with "haiku", it would default to the main agent model. I can update this to include try/catch logic if getChatEndpoint throws an error because a certain model isn't available

24anisha: @roblourens is there a way for me to debug in a state where I don't have access to a certain model, so I can confirm my implementation works as expected?

roblourens: I don't know of a way to configure that. You can hack the code somewhere in the model fetcher or endpoint provider maybe. I think all you need is a try/catch here.

24anisha: Ok, I can use that. I was worried it would be slow, since this function is called many times during the trajectory & falling into the catch statement tends to be slow relative to a regular function call.

roblourens (approved):

PR #3206, opened 2026-01-27
Add /init slash command for generating workspace instructions

digitarald:
Fixes microsoft/vscode#261358

Adds `/init` slash command that generates workspace instructions for AI coding agents.

## Changes
- **init.prompt.md**: New prompt file that discovers existing AI conventions and generates/updates workspace instructions
- **package.json**: Registers the new prompt
- **workspace-instructions.md**: Enhanced documentation with cross-editor support, AGENTS.md hierarchy, and core principles

### DIFF
--- a/assets/prompts/init.prompt.md
+++ b/assets/prompts/init.prompt.md
@@ -1,7 +1,7 @@
 ---
 name: init
 description: Generate or update workspace instructions file for AI coding agents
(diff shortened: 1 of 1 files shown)


aeschli:
@digitarald `/init` is too generic. Please use something like `/init-workspace-instructions` or `/instructions init` (we would then be able to extend that to `/instructions list` `/instructions install`.

PR #3213, opened 2026-01-28
Handle when the Messages API isn't available

TylerLeonhardt:
It's required for Claude Agent so just "be unavailable" when so.


bhavyaus:
What scenario caues Messages API to be unavailable?

bhavyaus (approved):

TylerLeonhardt:
@bhavyaus 

```
"github.copilot.chat.anthropic.useMessagesApi": false
```

😅 

but it's mostly just in case.

PR #3257, opened 2026-01-29
Surface external index state in status item too

mjbvz:
For microsoft/vscode#290782


dmitrivMS:
@mjbvz CI is unhappy...

DonJayamanne (approved):

PR #3265, opened 2026-01-29
allow plan to be used as subagent and do not restrict the use of subagents

aeschli:


digitarald:
`Infer:` @sbatten added switch_agent tool for the use case of invoking planning, which is the one we want to optomize for. Using planning as a sub-agent wouldn't a) use subagent for research for better results and b) not provide the handoff flow. 

`No custom subagents`: Custom agents can have access to all kind of tools. If the user somehow has a custom agent for specific file operations, it can still be invoked _("Make a TDD plan … let me invoke the TDD agent, but its actually meant for writing tests")_ and execute or edit code while the plan mode is active.

aeschli:
not taken

PR #3292, opened 2026-01-29
Add mode to panel.request telemetry event

kevin-m-kent:
Adds a 'mode' property to the panel.request telemetry event to track which chat mode (ask, edit, agent, custom) was used for each request. We currently have Command which does not perfectly map to the mode the user is in. We should have this to capture the mode they have selected, which we can then look at together with Command to understand the scenario. 

Also tagging @digitarald  - I noticed that plan will just be "custom" here (and in case you have other feedback on this as well).


kevin-m-kent:
I think I need to re-think this. My intention was that it would give information about the UI mode that was selected by the user when a request was sent, but it looks like this actually would grab the mode from the intent which is also like what command does, which is not what I want here.

kevin-m-kent:
Going to close this until I have a better idea.

PR #3310, opened 2026-01-30
Add 'subType: subagent' for request telemetry

roblourens:


src/extension/prompt/node/searchSubagentToolCallingLoop.ts
```diff
@@ -130,7 +130,8 @@ export class SearchSubagentToolCallingLoop extends ToolCallingLoop<ISearchSubage
 			userInitiatedRequest: false,
 			telemetryProperties: {
 				messageId: randomUUID(),
-				messageSource: SearchSubagentToolCallingLoop.ID
+				messageSource: 'chat.editAgent',
```
roblourens, line 133: FYI @24anisha if you are looking at this telemetry. I was asked to align it with normal agent requests. But you can look at the new `subType` prop to identify these

PR #3320, opened 2026-01-30
fix: update ask question instruction to better handle defaults

karthiknadig:
Updates the `ask_questions` tool description to better handle defaults.

Fixes https://github.com/microsoft/vscode/issues/291718

The model was setting \
ecommended: true\ on correct quiz answers, which pre-selected those options and spoiled the answers to users. This change updates the instruction to explicitly warn the model that recommended options are visible and pre-selected, so it should never use them for quizzes or polls.


bpasero (approved):

package.json
```diff
@@ -1163,7 +1163,7 @@
...
...ke that action instead of asking.\n\nWhen to use:\n- Clarify ambiguous requirements before proceeding\n- Get user preferences on implementation choices\n- Confirm decisions that meaningfully affect outcome\n\nWhen NOT to use:\n- The answer is determinable from code or context\n- Asking for permission to continue or abort\n- Confirming something you can reasonably decide yourself\n- Reporting a problem (instead, attempt to resolve it)\n\nQuestion guidelines:\n- NEVER use `recommended` for quizzes or polls. Recommended options are PRE-SELECTED and visible to users, which would reveal answers\n- Batch related questions into a single call (max 4 questions, 2-6 options each; omit options for free text input)\n- Provide brief context explaining what is being decided and why\n- Only mark an option as `recommended` to suggest YOUR preferred implementation choice\n- Keep options mutually exclusive for single-select; use `multiSelect: true` only when choices are additive and phrase the question accordingly\n\nAfter receiving answers:\n- Incorporate decisions and continue without re-asking unless requirements change\n\nAn \"Other\" option is automatically shown to users—do not add your own.",
```
digitarald, line 1166: The previous version had "Mark one option as `recommended` **with a short justification**" — should we keep that guidance? A short justification helps users understand why an option is recommended.

eleanorjboyd: @karthiknadig  even though it got closed maybe a new PR for this

digitarald:
Fixed in 1884d464 — added back "with a short justification" for recommended options.

PR #3327, opened 2026-01-30
Removing hash as this is incorrect with gpt-5.2-codex

dileepyavan:
fixes: https://github.com/microsoft/vscode/issues/291958

Removed check with hashed value for the model.

### DIFF
--- a/src/platform/endpoint/common/chatModelCapabilities.ts
+++ b/src/platform/endpoint/common/chatModelCapabilities.ts
@@ -12,11 +12,6 @@ const HIDDEN_MODEL_A_HASHES = [
 	'6b0f165d0590bf8d508540a796b4fda77bf6a0a4ed4e8524d5451b1913100a95'
 ];
 
-const HIDDEN_MODEL_C_HASHES = [
-	'2daf0e6c79009f9234ed9baa5bb930898e2847810617e118518d88e4d3140a2e',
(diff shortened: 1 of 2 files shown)


roblourens:
@dileepyavan You don't need a review from me, you own this code 😁

src/platform/endpoint/common/chatModelCapabilities.ts
```diff
@@ -107,6 +97,11 @@ export function isVSCModelC(model: LanguageModelChat | IChatEndpoint) {
 	return VSC_MODEL_HASHES_SUBSET_C.includes(ID_hash) || VSC_MODEL_HASHES_SUBSET_C.includes(family_hash);
 }
 
+export function isGpt5_2_CodexFamily(model: LanguageModelChat | IChatEndpoint | string): boolean {
```
zhichli, line 100: nit: we have couple of names with no `_` in the middle e.g. `isGpt51Family` and `isGptCodexFamily`, maybe it's better to make the naming a bit more consistent so that we can easily search relevant code in workspace. Like I lierally searched `Gpt51` and `Gpt52` with no underscore in the repo when reporting the issue of this PR but could miss such names with `_` in it.

dileepyavan: got it. will update this in the next pr.

PR #3336, opened 2026-01-30
Removing hash as this is incorrect with gpt-5.2-codex (#3327)

dileepyavan:
Fixes: https://github.com/microsoft/vscode/issues/291958

Removing the hash that is used for tented model name. Current hashes are not resolving to intended gpt-5.1-codex prompt for 5.2-codex model.

### DIFF
--- a/src/platform/endpoint/common/chatModelCapabilities.ts
+++ b/src/platform/endpoint/common/chatModelCapabilities.ts
@@ -12,11 +12,6 @@ const HIDDEN_MODEL_A_HASHES = [
 	'6b0f165d0590bf8d508540a796b4fda77bf6a0a4ed4e8524d5451b1913100a95'
 ];
 
-const HIDDEN_MODEL_C_HASHES = [
-	'2daf0e6c79009f9234ed9baa5bb930898e2847810617e118518d88e4d3140a2e',
-	'854910cb8a637565e3933f64549ac7e83dd55f6721e98f8324eb2065a84e09f8'
-];
-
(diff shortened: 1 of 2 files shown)


src/platform/endpoint/common/chatModelCapabilities.ts
```diff
@@ -272,7 +267,7 @@ export function isGpt51Family(model: LanguageModelChat | IChatEndpoint | string
 	}
 
 	const family = typeof model === 'string' ? model : model.family;
-	return !!family.match(/^gpt-5\.\d+/i) || isHiddenModelC(family);
+	return !!family.match(/^gpt-5\.\d+/i) || isGpt5_2_CodexFamily(family);
```
rebornix, line 270: is this redundant?

rebornix: also maybe for later improvements, the name `isGpt51Family` is a bit misleading here.

kycutler: Agree this seems redundant, as it looks like it's only used in one place which immediately limits it back down to non-codex models:

https://github.com/microsoft/vscode-copilot-chat/blob/c110bfddb99738e53e393e6a3f3aef3d1a379d38/src/extension/prompts/node/agent/openai/gpt51Prompt.tsx#L276

dileepyavan: I have another pr for 5.2 thats taking care of this

PR #3370, opened 2026-02-02
Remove conecpt of CAPI driven default

lramos15:
As suggested in https://github.com/microsoft/vscode/issues/291940#issuecomment-3833889669


isidorn:
Thank you very much for making this change. Anyway to include this as part of some recovery release?

Reason - we should try to minimize GPT 4.1 usage, and this means that all users with telemetry disabled get GPT 4.1 as default

cc @pierceboggan

lramos15:
We can, but I normally like to reserve candidates for critical bugs and this doesn't feel like an urgent issue just an oversight that has existed for awhile

isidorn:
Thanks Logan.

Let's see what @pierceboggan and @sbatten think as well. As my hypothesis is that due to this a lot of users are accidently on gpt-4.1 (those that disabled telemetry for sure)

PR #3392, opened 2026-02-03
Use subAgentInvocationId to determine agent name for trajectories

zhichli:
Changed the trajectory agent name logic to use `subAgentInvocationId` presence instead of `subAgentName` fallback.

## Changes

- Now uses `'subagent'` when `subAgentInvocationId` is defined
- Uses `'GitHub Copilot Chat'` for the main agent

This provides a cleaner determination of agent type based on the invocation ID rather than relying on the name property.


justschen (approved):

src/platform/trajectory/node/trajectoryLoggerAdapter.ts
```diff
@@ -81,8 +82,8 @@ export class TrajectoryLoggerAdapter extends Disposable {
 
 			// Get or create session for this token
 			let sessionId = this.sessionMap.get(entry.token);
-			// Use subAgentName as agent name for subagent trajectories
-			const agentName = entry.token.subAgentName ?? AGENT_NAME;
+			// Use 'subagent' for subagent trajectories, main agent name otherwise
```
zhichli, line 85: @copilot rm the comment here

PR #3401, opened 2026-02-03
Simple classifier

digitarald:


kevin-m-kent:
I really like where this is right now  @digitarald ! It aligns with I was playing with dimensions-wise to generate some synthetic chat data. 

Something I'm wondering about a tweak to the instructions. I like the examples and the signals. Should we add an instruction that tells the classifier how to use the signals and keywords - for instance "The signals serve as example to evaluate but are not exhaustive"? I wonder if as is the model will see the signals as exhaustive keywords rather than a hint about the category.

kevin-m-kent:
One other question - should we add an unknown scope option to be consistent with the others?

PR #3410, opened 2026-02-03
encourage multiple subagents in plan mode

joshspicer:


joshspicer:
agreed, thanks! See https://github.com/microsoft/vscode-internalbacklog/issues/6701

PR #3454, opened 2026-02-04
Remove images from history that don't support vision

lramos15:
Fix https://github.com/microsoft/vscode/issues/291636


justschen (approved):
fire, thanks thanks

PR #3480, opened 2026-02-05
Fix ollama model for plan/agent mode

ctrysbita:
This PR fixes https://github.com/microsoft/vscode/issues/293071

PR https://github.com/microsoft/vscode-copilot-chat/pull/2782 introduces the issue by passing the wrong variable (the raw model list) into `byokKnownModelsToAPIInfo` , this PR fixes it by passing the correct `knownModels`

By the way, this PR also fixes crash on undefined `usage` when using ollama models

### DIFF
--- a/src/extension/codeBlocks/node/codeBlockProcessor.ts
+++ b/src/extension/codeBlocks/node/codeBlockProcessor.ts
@@ -132,7 +132,7 @@ export class CodeBlockTrackingChatResponseStream implements ChatResponseStream {
 	externalEdit = this.forward(this._wrapped.externalEdit.bind(this._wrapped));
 	beginToolInvocation = this.forward(this._wrapped.beginToolInvocation.bind(this._wrapped));
 	updateToolInvocation = this.forward(this._wrapped.updateToolInvocation.bind(this._wrapped));
-	usage = this.forward(this._wrapped.usage.bind(this._wrapped));
+	usage = this.forward(this._wrapped.usage?.bind(this._wrapped) || (() => { }));
 
 	questionCarousel(questions: ChatQuestion[], allowSkip?: boolean): Thenable<Record<string, unknown> | undefined> {
 		this._codeBlockProcessor.flush();


Andygol:
@ctrysbita could you [accept CLA](https://github.com/microsoft/vscode-copilot-chat/pull/3480#issuecomment-3853247712) to give a go to this PR 🙏

ctrysbita:

CristianVGdev (commented):
@ctrysbita  Suggestion: avoid `bind` + permanent no-op for `usage`

Current:
```ts
usage = this.forward(this._wrapped.usage?.bind(this._wrapped) || (() => { }));
````

Proposal:

```ts
usage = this.forward((...args: any[]) => this._wrapped.usage?.apply(this._wrapped, args));
```

Reason:

* Fixes the original problem (`bind` on `undefined`) while maintaining the behavior of `flush()` through `forward`.
* Avoids capturing a null operation at build time (if `usage` is added later in some implementations/flags, this will start working without rebuilding the flow).
* Preserves the correct `this` without `bind`.

If this causes any strange behavior in downstream flow, I think your original patch would remain. This is primarily a “worth trying” improvement for optional proxy methods.

rajasrijan:
Facing the same issue. Is there a work around i can use till the fix is propagated to release?

ctrysbita:
@hediet Hi, could you have a review on this?

CristianVGdev:
@ctrysbita 
> @hediet Hi, could you have a review on this?

The more you push into the pipeline, the more reviews will be required, and you need two.
That said, I mentioned the issue of avoid bind + permanent no-op for usage, and you haven't responded to that either.

CristianVGdev:
Fixed on #3566

ctrysbita:
> @ctrysbita
> 
> > @hediet Hi, could you have a review on this?
> 
> The more you push into the pipeline, the more reviews will be required, and you need two. That said, I mentioned the issue of avoid bind + permanent no-op for usage, and you haven't responded to that either.

Actually it always need one reviewer **with write permission**.
For the `usage`, if the wrapped upstream doesn't have a `usage` implementation it should remain stable during the session so I just want to keep it simple and consistent with existing code like `workspaceEdit` above.

ctrysbita:
> Fixed on #3566

Good, let's keep track on that PR.

PR #3487, opened 2026-02-05
Fix completions context handling for similar files

dbaeumer:
Adjust the NeighborFileModel to manage visible and not visible files separately, improving the handling of completions for similar files. This change addresses issue #290387.


Yoyokrazy (approved):

Yoyokrazy:
reasonable seeming CCR recommendations to take in a future PR potentially.

PR #3511, opened 2026-02-06
use hook progress in chat

justschen:
ref https://github.com/microsoft/vscode/pull/293514

see above PR for screenshot of what this looks like.

cc @pwang347 @roblourens 

519-Pidove

### DIFF
--- a/src/extension/linkify/common/responseStreamWithLinkification.ts
+++ b/src/extension/linkify/common/responseStreamWithLinkification.ts
@@ -77,8 +77,8 @@ export class ResponseStreamWithLinkification implements FinalizableChatResponseS
 		return this;
 	}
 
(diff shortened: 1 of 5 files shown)


src/extension/linkify/common/responseStreamWithLinkification.ts
```diff
@@ -77,6 +77,11 @@ export class ResponseStreamWithLinkification implements FinalizableChatResponseS
 		return this;
 	}
 
+	hookProgress(hookType: ChatHookType, shouldContinue: boolean, stopReason?: string, systemMessage?: string): ChatResponseStream {
```
pwang347: what do you think about merging `shouldContinue` and `stopReason` into `stopReason`? I think @roblourens already did this actually for our internal types

PR #3549, opened 2026-02-07
Support PreCompact hook

roblourens:


src/platform/chat/common/chatHookService.ts
```diff
@@ -140,4 +140,19 @@ export interface SubagentStopHookOutput {
...
+	/**
+	 * How the compaction was triggered.
+	 * "auto" when the conversation is too long for the prompt budget.
+	 */
+	readonly trigger: 'auto';
+	/**
+	 * Custom instructions for the compaction, if any.
+	 */
+	readonly custom_instructions?: string;
+}
```
roblourens, line 160: We don't currently have manual compaction, so keeping the type narrow to `'auto'` for now. The hooks command wizard lists `'manual'` speculatively but it's not wired up. We can widen the type when/if manual compaction is implemented.

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -430,9 +431,13 @@ class ConversationHistorySummarizer {
 		@IConfigurationService private readonly configurationService: IConfigurationService,
 		@IExperimentationService private readonly experimentationService: IExperimentationService,
 		@IEndpointProvider private readonly endpointProvider: IEndpointProvider,
+		@IChatHookService private readonly chatHookService: IChatHookService,
 	) { }
 
 	async summarizeHistory(): Promise<{ summary: string; toolCallRoundId: string; thinking?: ThinkingData }> {
+		// Execute pre-compact hook before summarization to allow hooks to archive transcripts or perform cleanup
+		await this.executePreCompactHook();
+
```
roblourens, line 440: The existing summarization tests in `summarization.spec.tsx` exercise this path — `ConversationHistorySummarizer` is created via `IInstantiationService` and picks up `NullChatHookService` from the test service collection, which returns `[]` and verifies the hook execution doesn't break the summarization flow. All 15 existing tests pass with the hook integrated.

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -473,6 +478,36 @@ class ConversationHistorySummarizer {
...
+			return;
+		}
+
+		try {
+			const results = await this.chatHookService.executeHook('PreCompact', {
+				toolInvocationToken,
+				input: {
+					trigger: 'auto',
+				}
+			}, this.token ?? CancellationToken.None);
```
roblourens: Good suggestion — added `satisfies PreCompactHookInput` annotation.

roblourens (commented):
Addressed the `satisfies` suggestion. The other two comments are responded to inline.

PR #3555, opened 2026-02-07
Fix subagent loading for Claude Code sessions

roblourens:
[Written by GitHub Copilot]

This fixes two bugs that prevented subagents from being loaded when viewing Claude Code sessions.

## Bug 1: Subagent directory not discovered by `getSession`

The service maintained a `_sessionDirs` cache that tracked which session IDs had corresponding directories (containing subagents). This cache was only populated during `getAllSessions` → `_loadSessionMetadataFromDisk`. When `getSession` was called directly (the common path), the cache was empty, so the subagent directory was never checked and subagents were always missing.

**Fix:** Removed the `_sessionDirs` cache entirely and instead directly check the subagents directory on disk via `_tryReadDirectory` during `_loadFullSession`. This ensures subagents are discovered regardless of whether `getAllSessions` was called first. Also renamed `_loadSubagentsForSession` to `_loadSubagentsFromEntries` since the directory entries are now read before calling it.

## Bug 2: Progress entries break leaf node detection in `buildSubagentSession`

In subagent JSONL files, progress entries (`bash_progress`, `hook_progress`) have `uuid` and `parentUuid` fields that match the `vChainLinkEntry` validator schema, so the parser treats them as chain links. In `buildSubagentSession`, chain link parent references were included when building the `referencedAsParent` set used to find leaf nodes. Since progress entries form a chain referencing every message as a parent, ALL messages were marked as non-leaves, no chain could be built, and the function returned `null`.

**Fix:** In `buildSubagentSession`, only consider message-to-message parent references when detecting leaf nodes. Chain link references (which include progress entries) are excluded from the leaf detection set.

## Test

Added a test using real fixture files (a session with a Bash subagent running `sleep 10`) that validates end-to-end subagent parsing through `getSession`.

### DIFF
--- a/src/extension/agents/claude/node/sessionParser/claudeSessionParser.ts
+++ b/src/extension/agents/claude/node/sessionParser/claudeSessionParser.ts
@@ -194,6 +194,15 @@ export function parseSessionFileContent(
  * Convert a user message entry's timestamp from string to Date.
  */
 function reviveUserMessage(entry: UserMessageEntry): StoredMessage {
+	// Extract agentId from toolUseResult if present (links Task tool_use to subagent).
+	// The toolUseResult field is typed as string | object by the validator; for Task tools
+	// it's an object containing { agentId, status, prompt, ... }. The `in` check narrows
+	// the type so no unsafe cast is needed.
+	let toolUseResultAgentId: string | undefined;
+	if (entry.toolUseResult && typeof entry.toolUseResult === 'object' && 'agentId' in entry.toolUseResult && typeof entry.toolUseResult.agentId === 'string') {
+		toolUseResultAgentId = entry.toolUseResult.agentId;
+	}
+
 	return {
 		uuid: entry.uuid,
 		sessionId: entry.sessionId,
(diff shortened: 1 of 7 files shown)


src/extension/agents/claude/node/sessionParser/claudeCodeSessionService.ts
```diff
@@ -564,10 +552,10 @@ export class ClaudeCodeSessionService implements IClaudeCodeSessionService {
 			}
 
 			// Load subagents if available
-			const sessionDirs = this._sessionDirs.get(projectDirUri);
```
roblourens, line 567: I'm not totally sure about this? But it seems ok?

src/extension/agents/claude/node/sessionParser/claudeSessionParser.ts
```diff
@@ -533,18 +533,16 @@ export function buildSubagentSession(
...
+	// non-leaves).
 	const referencedAsParent = new Set<string>();
 	for (const message of messages.values()) {
 		if (message.parentUuid !== null) {
 			referencedAsParent.add(message.parentUuid);
 		}
 	}
-	for (const chainLink of chainLinks.values()) {
-		if (chainLink.parentUuid !== null) {
-			referencedAsParent.add(chainLink.parentUuid);
```
roblourens, line 545: And I'm not sure if there's something else we lose here?

TylerLeonhardt (dismissed):

PR #3558, opened 2026-02-08
ToolCallingLoop: avoid duplicate start hooks

bharatvansh:
### What
Fixes a foot gun where `runStartHooks()` can be invoked by callers (e.g. to run SessionStart/SubagentStart before other hooks) and then `run()` would effectively run the same “start” work again.

### Why
This can lead to duplicate `SessionStart` / `SubagentStart` executions (and transcript initialization/logging) for a single `ToolCallingLoop` instance.

### Changes
- Make `ToolCallingLoop.run()` always call `runStartHooks()` up-front.
- Make `runStartHooks()` idempotent per loop instance via a promise latch, so multiple calls safely share the same work.
- Add regression tests ensuring `SessionStart`/`SubagentStart` aren’t executed twice when `runStartHooks()` is called before `run()`.

### Verification
- `npm run test:unit`
- `npm run test:extension`
- `npm run lint`


bharatvansh:
Hi @pwang347
I’ve rebased this PR on latest main. When you get a chance, could you please take a look and let me know whether this is something you’d want to merge, or if I should just close it?

PR #3565, opened 2026-02-09
Segment for plan agent

digitarald:


src/extension/prompt/node/chatParticipantTelemetry.ts
```diff
@@ -428,7 +428,7 @@ export abstract class ChatTelemetry<C extends IDocumentContext | undefined = IDo
 	}
 
 	protected _getModeName(): string {
-		return this._request.modeInstructions2 ? 'custom' :
+		return this._request.modeInstructions2 ? (this._request.modeInstructions2.name.toLowerCase() === 'plan' ? 'plan' : 'custom') :
```
JamesMGreene, line 431: Will this be limited strictly to the first party agent named "Plan", or would a _user's_ custom agent named "Plan" also leak through here?

digitarald: Doesn't leak, its just not 100% correct. I'd argue that having a custom "plan" mode is fine, and still correctly reporting "a user used planning" for the dashboard.

PR #3595, opened 2026-02-09
fix: enable and robustify interactive question carousel in AskQuestionsTool

bharatvansh:
This PR fixes issues in `AskQuestionsTool` where the interactive question UI was being bypassed and improves the robustness of answer mapping.

### Breakdown of Changes
- **Unlocked the Carousel**: Corrected the invocation logic to ensure the interactive `questionCarousel` UI is displayed to the user instead of being automatically skipped or auto-answered. It now explicitly disallows skipping to ensure user intent is captured.
- **Improved Context Tracking**: Implemented `requestId` and input-based tracking for `promptContext` using `LRUCache` and `WeakMap`. This ensures the tool correctly resolves the active stream and context even when multiple requests are in flight, preventing race conditions.
- **Numbered Headers**: Added 1-based indexing to question headers (e.g., "1. Your Question") to provide a more organized experience for the user.
- **Robust Answer Mapping**: The mapping logic now supports multiple identification keys (`id`, `header`, `title`) and normalized formats (e.g., matching `"1. Your Question"` back to `"Your Question"`). It also includes coercion for various data types (string, number, boolean) and rich option objects returned by the UI.
- **Fail-safe Logic**: Added a fallback mechanism that attempts to ask questions one-by-one if the multi-question carousel fails to return a result in certain host environments.
- **Enhanced Testing**: Expanded unit tests to cover various answer formats and removed `as any` type casts by refining internal signatures.

### Observations / Potential Improvements (Core UI)
The following behaviors are governed by the VS Code core renderer (`microsoft/vscode`). While out of scope for this extension PR, they could be addressed in the core repo:
1. **Auto-Submit on Selection**: Currently, selecting the last option in a single-choice question auto-submits the whole carousel. A review step might be preferable.
2. **Alphabetical Options**: While questions use numeric indexes, the options might feel more natural with alphabetical labels (A, B, C).
3. **Empty Options Handling**: The tool now handles text-only questions, but enforcing at least two options via API constraints could improve consistency.

### DIFF
--- a/src/extension/tools/vscode-node/askQuestionsTool.ts
+++ b/src/extension/tools/vscode-node/askQuestionsTool.ts
@@ -7,6 +7,7 @@ import * as vscode from 'vscode';
 import { ILogService } from '../../../platform/log/common/logService';
 import { ITelemetryService } from '../../../platform/telemetry/common/telemetry';
 import { CancellationToken } from '../../../util/vs/base/common/cancellation';
+import { LRUCache } from '../../../util/vs/base/common/map';
 import { StopWatch } from '../../../util/vs/base/common/stopwatch';
 import { ChatQuestion, ChatQuestionType, LanguageModelTextPart, LanguageModelToolResult, MarkdownString } from '../../../vscodeTypes';
 import { IBuildPromptContext } from '../../prompt/common/intents';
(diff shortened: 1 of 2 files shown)


bharatvansh:
The screenshots:
Screenshot 2026-02-09 155735
Screenshot 2026-02-09 155744

bharatvansh:
The first commit wasn't complete. probably some changes got truncated without me realizing and then i pushed the PR. However Now I have pushed the second commit which does include all the indented changes I wanted to do. The PR us ready for review now!

PR #3608, opened 2026-02-10
Ask agent provider for read-only Q&A mode

benibenj:
```Copilot Generated Description:``` Introduce a new Ask agent provider for read-only Q&A mode, along with additional configuration options for tools and model overrides. This includes updates to the configuration settings and the addition of default read-only tools.


roblourens (approved):
🔥

PR #3653, opened 2026-02-11
fixes for processHookResult

justschen:
makes sure we can processHookResult the right way in the callback

### DIFF
--- a/src/extension/intents/node/hookResultProcessor.ts
+++ b/src/extension/intents/node/hookResultProcessor.ts
@@ -124,12 +124,12 @@ export function processHookResults(options: ProcessHookResultsOptions): void {
 	}
 
 	// Show aggregated warnings via hookProgress
-	if (warnings.length > 0 && outputStream) {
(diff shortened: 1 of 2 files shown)


src/extension/chat/vscode-node/chatHookService.ts
```diff
@@ -259,44 +260,49 @@ export class ChatHookService implements IChatHookService {
...
+				if (!hookSpecificOutput) {
+					return;
+				}
 
-			// Skip results from other hook event types
-			if (hookSpecificOutput.hookEventName !== undefined && hookSpecificOutput.hookEventName !== 'PreToolUse') {
-				continue;
-			}
+				// Skip results from other hook event types
+				if (hookSpecificOutput.hookEventName !== undefined && hookSpecificOutput.hookEventName !== 'PreToolUse') {
```
pwang347, line 280: we should probably do this in the result processor, but this is fine for now

pwang347 (dismissed):

pwang347 (dismissed):

PR #3655, opened 2026-02-11
Fall back to node-fetch when Electron network process crashes

alexdima:
Fixes microsoft/vscode-internalbacklog#6748

---

When Electron's network service process crashes mid-session, all requests through it fail permanently with net::ERR_FAILED (network_process_crashed). The process does not recover until the extension host restarts.

This change detects the crash via chromiumDetails.network_process_crashed on the error object and responds at two levels:

1. FetcherService (fetchWithFallbacks): permanently demotes the crashed fetcher and promotes node-fetch as the primary fetcher. All subsequent requests (chat, auth, ping, etc.) bypass Electron entirely with zero overhead. Gated by FallbackNodeFetchOnNetworkProcessCrash experiment.

2. chatMLFetcher (_retryAfterError): for the first crashing chat request, switches the connectivity check and retry to use node-fetch, so the request can recover. Also gated by the experiment flag.

Added isNetworkProcessCrashedError() to IFetcher/IFetcherService interfaces. ElectronFetcher implements structural detection via chromiumDetails. Error classification methods (isFetcherError, isNetworkProcessCrashedError, getUserMessageForFetcherError) now check all fetchers so they still work correctly after a fetcher demotion.

Added isNetworkProcessCrash flag to ChatFetchResponseType.NetworkError. Added ElectronFetchErrorChromiumDetails type for structured error access.

30 unit tests covering retry logic, fetcher demotion, experiment gating, and error classification after demotion.


deepak1556:
Thank you! Can we have one or two retry on the electron network stack before promoting node-fetch as the primary in the crash case. Reasoning, the network process when crashes **will** restart, the issue is the cached connection in the extension host to the old network process that creates the issue. Having a retry would also allow me to validate the upstream fix when it eventually lands through the experiment you have setup here.

alexdima:
> Can we have one or two retry on the electron network stack before promoting node-fetch as the primary in the crash case.

🤔 I think this was already implemented before this change -- see https://github.com/microsoft/vscode-copilot-chat/pull/3655/changes#diff-d375dd66527fd3cae1fa540bf018e6b014f02e7924060fab57dae8d905a95387R513 but it doesn't seem to work in all cases.

I initially approached this with only making changes in ChatMLFetcher, but then each new interaction would keep trying to reuse the ElectronFetcher which would introduce delays for each and every request. That's when I did the second change in the `fetchWithFallbacks` where the ElectronFetcher is demoted if the networking process crashes. But @chrmarti has a much better understanding of the area than me, I think some retry logic is implemented deep down in `fetchWithFallbacks` and some retry logic is implemented in ChatMLFetcher, which makes things a bit tricky.

I am fairly confident all the changes are behind the exp enabled configuration setting so I think we should be safe to run an experiment and then we can undo everything if it doesn't prove to help. What do you think?

PR #3660, opened 2026-02-11
Show warning message for known issues in specific VS Code versions

sandy081:


tracernz:
This created an unavoidable message box that blocks the entire window each time you open a new one. Extremely annoying for people who do not care for copilot. A good reminder to uninstall it. It would have been much better as a normal message in the tray area like other extensions do.

package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "copilot-chat",
 	"displayName": "GitHub Copilot Chat",
 	"description": "AI chat features powered by Copilot",
-	"version": "0.37.4",
+	"version": "0.37.5",
 	"build": "1",
 	"internalAIKey": "1058ec22-3c95-4951-8443-f26c1f325911",
```
subhamkashyap555555-glitch, line 7: $ npm install standard --global

PR #3676, opened 2026-02-11
Rename README.md to Should read

daotuananh1999tgdd-sudo:


typed-sigterm (changes requested):
do you actually know what you're doing?

PR #3566, opened 2026-02-09
Fix tool calling detection and support for Ollama models

lsby:
## Summary

This PR fixes issues where Ollama models could not be used in **Agent mode** because tool calling capabilities were not correctly identified or propagated.

## Environment

**Ollama**

- Version: 0.15.5
- Model: `mistral-small3.2:latest` (supports vision and tool calling)

**VSCode**

- Version: 1.110.0-insider (user setup)
- Commit: c308cc9f87e1427a73c5e32c81a1cfe9b1b203d1

---

## Changes

- Pass `knownModels` instead of `models` to `byokKnownModelsToAPIInfo`
- Initialize `_knownModels` in `OllamaLMProvider`
- Ensure `supportsToolCalls` is correctly derived for Ollama models

---

### Issue 1: Ollama models are not correctly recognized as supporting tool calls

Ollama models are not correctly identified as supporting the "tools" capability.

QQ20260209-070939

This happens because [`getAllModels` passes `models` to `byokKnownModelsToAPIInfo`](https://github.com/microsoft/vscode-copilot-chat/blob/f9a18ed6848b3e6b0d2bf1124f25680fa67e919e/src/extension/byok/vscode-node/ollamaProvider.ts#L110).

However, based on the expected input of [`byokKnownModelsToAPIInfo`](https://github.com/microsoft/vscode-copilot-chat/blob/5a6bc84c913451778f8d6db280fd0e2a91fe7310/src/extension/byok/common/byokProvider.ts#L152), `knownModels` should be passed instead. After applying this change, the model information is correctly populated.

QQ20260209-070136

---

### Issue 2: The `tools` field is not sent to the Ollama model

When invoking an Ollama model, the `tools` section is not included in the request.

This happens because [the following code](https://github.com/microsoft/vscode-copilot-chat/blob/5a6bc84c913451778f8d6db280fd0e2a91fe7310/src/platform/endpoint/node/chatEndpoint.ts#L261-L263) checks `supportsToolCalls`. If it is false, the `tools` field is removed.

The value of `supportsToolCalls` ultimately comes from [this code](https://github.com/microsoft/vscode-copilot-chat/blob/5a6bc84c913451778f8d6db280fd0e2a91fe7310/src/extension/byok/common/byokProvider.ts#L126-L132).  
[...]

PR #3684, opened 2026-02-12
Fix Ollama 404 error and streaming response parsing

lsby:
## Summary

This PR fixes a `404` error when calling Ollama models.

The root causes include:

1. Incorrect API endpoint being used  
2. Incompatibility between the streaming response parser and Ollama’s response format

image

<sub>Screenshot from [here](https://github.com/microsoft/vscode-copilot-chat/pull/3566#issuecomment-3886283066)</sub>

---

## Environment

**Ollama**

* Version: 0.15.5
* Model: `mistral-small3.2:latest` (supports vision and tool calling)

**VSCode**

* Version: 1.110.0-insider (user setup)
* Commit: f43307e9e5511fa40b15ff8e3edbd84b225fc8d0

---

## Issue 1: Incorrect API Endpoint

Ollama uses the API path `/api/chat` (see [official documentation](https://docs.ollama.com/api/chat)), not `/chat/completions`.

In the current implementation, `OllamaLMProvider` extends `AbstractOpenAICompatibleLMProvider`, which [constructs the request path based on the OpenAI-compatible protocol](https://github.com/microsoft/vscode-copilot-chat/blob/e7fbbbc2dd193fba5e773773534faf30227083c9/src/extension/byok/vscode-node/abstractLanguageModelChatProvider.ts#L161-L166). As a result, requests were incorrectly sent to `/chat/completions`, leading to a `404` error.

image

I overrode `createOpenAIEndPoint` in `OllamaLMProvider` to provide the correct path.

---

## Issue 2: Incompatible Streaming Response Parsing

Ollama’s streaming responses use **JSON Lines** format (one JSON object per line).

The current parser [does not handle this format](https://github.com/microsoft/vscode-copilot-chat/blob/e7fbbbc2dd193fba5e773773534faf30227083c9/src/platform/endpoint/node/chatEndpoint.ts#L373-L381), which prevents it from correctly parsing the streamed content.

To resolve this, I implemented a dedicated response parser `processOllamaStreamResponse` for Ollama.

Since the parser needs to be conditionally applied based on the endpoint type, I introduced a subclass in `src/extension/byok/node/ollamaEndpoint` to indicate whether the current endpoint targets Ollama.


lsby:

kingeke:
LGTM

wssccc:
Ollama has a built-in OAI compatible endpoint /v1/chat/completions, which was working well previously. Why not just fix that path issue?

lsby:
@wssccc

Thanks for pointing that out. I wasn’t aware that Ollama provides a built-in OpenAI-compatible endpoint at /v1/chat/completions.

After testing, it works as expected. Given that, updating the endpoint path is indeed a cleaner solution and avoids the need for a custom streaming parser.

I’ve revised the commits accordingly to use the compatible endpoint and removed the additional parsing logic. Thanks again for the suggestion.

oomek:
It's taking a while to approve just 6 lines of code :/

MarslMarcello:
> It's taking a while to approve just 6 lines of code :/

CoPilot needs its time. 😅 🙃

riverar (approved):

andyathoppersoft:
Can someone with write access review this PR, please?

adifyr:
Now, if y'all could just go ahead and approve this PR right 'ere, the rest of us can move on with our lives. Much obliged...

kingeke:
Just one more person, come on humanity !!!

sandy081:
Reopening to try again to merge

sandy081:
Since this is stuck in approvals, I created another PR to merge this - https://github.com/microsoft/vscode-copilot-chat/pull/3858

riverar:
Thanks @sandy081. Can we talk about why this remained broken for a week+ and required back channel communication? Is there something we can do in the future to better alert folks to critical issues like this?

riverar:
@sandy081 Can you confirm this is scheduled for release ~February 28, 2026? Is there a faster hotfix pipe this could potentially be pushed into?

harn3ss:
the only agent i'm able to use is stuck behind a pay wall. can this not be released sooner than more than a week from now?

PR #3695, opened 2026-02-12
Try to optimize claude controller

mjbvz:
Counterpart of https://github.com/microsoft/vscode/pull/294798

I will optimize our implementation in core further too but spotted a few potentially easy wins here

### DIFF
--- a/src/extension/chatSessions/vscode-node/claudeChatSessionContentProvider.ts
+++ b/src/extension/chatSessions/vscode-node/claudeChatSessionContentProvider.ts
@@ -616,18 +616,17 @@ export class ClaudeChatSessionItemController extends Disposable {
 				};
 				item = this._createClaudeChatSessionItem(newlyCreatedSessionInfo);
 			}
+
+			this._controller.items.add(item);
 		}
 
 		item.status = status;
 		if (status === vscode.ChatSessionStatus.InProgress) {
-			if (!item.timing) {
-				item.timing = {
-					created: Date.now()
-				};
-			}
-			item.timing.lastRequestStarted = Date.now();
+			const timing = item.timing ? { ...item.timing } : { created: Date.now() };
+			timing.lastRequestStarted = Date.now();
 			// Clear lastRequestEnded while a request is in progress
-			item.timing.lastRequestEnded = undefined;
+			timing.lastRequestEnded = undefined;
+			item.timing = timing;
 		} else if (status === vscode.ChatSessionStatus.Completed) {
 			if (!item.timing) {
 				item.timing = {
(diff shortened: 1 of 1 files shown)


src/extension/chatSessions/vscode-node/claudeChatSessionContentProvider.ts
```diff
@@ -616,43 +616,35 @@ export class ClaudeChatSessionItemController extends Disposable {
...
 		const sessions = await this._claudeCodeSessionService.getAllSessions(token);
 		const items = sessions.map(session => this._createClaudeChatSessionItem(session));
-		// If there are already items, use add() to update in place and avoid re-rendering the whole list.
-		// TODO: Could we see if the sessions are the same and only update changed ones?
-		if (this._controller.items.size > 0) {
-			items.forEach(i => this._controller.items.add(i));
-		} else {
-			this._controller.items.replace(items);
-		}
+		this._controller.items.replace(items);
```
mjbvz, line 647: About the re-rendering: let's fix this in core if that's really an issue. I want to make it so extension don't need to optimize  like this

The problem you were running into trying to optimize using `add` is that before https://github.com/microsoft/vscode/pull/294798, I had done zero optimizations. That meant if you had 400 session, this loop would trigger 400 individual ipc message. Each of the 400 ipc updates would then send the entire set of 400 session's data over to the main thread

After https://github.com/microsoft/vscode/pull/294798 calling `add` like this will still potentially trigger 400 ipc message, but now each message sends at most one item. If the item hasn't changed, we skip the message entirely

src/extension/chatSessions/vscode-node/claudeChatSessionContentProvider.ts
```diff
@@ -616,43 +616,35 @@ export class ClaudeChatSessionItemController extends Disposable {
...
 				item.timing = {
 					created: Date.now(),
 					lastRequestEnded: Date.now()
 				};
 			} else {
-				item.timing.lastRequestEnded = Date.now();
+				item.timing = { ...item.timing, lastRequestEnded: Date.now() };
 			}
 		}
-		this._controller.items.add(item);
```
mjbvz, line 641: Once an item is in the collection, you should just be able to set the properties on it and do not need to call `add` again. I think you were working around `timing` not updating. I've made it more clear you need to set the entire `timing` property to get an update with https://github.com/microsoft/vscode/pull/294801

TylerLeonhardt: This makes sense. Thanks!

TylerLeonhardt (approved):

src/extension/chatSessions/vscode-node/claudeChatSessionContentProvider.ts
TylerLeonhardt: I wonder how in progress sessions are handled with this change. Both:
* new ones that have not yet been written to disk
* old sessions that have a new message in them

Based on the code, I'm guessing if we follow the steps:

* Send a message
* Immediately hit refresh sessions button (or perhaps something else triggers refresh automatically)
* :bug: sessions list does something unexpected
      * For a new session it would disappear 
      * For an old session it would say “completed”

Would I need to keep track of those that are in-progress and make sure when I create the item that they are reflected accordingly?

PR #3712, opened 2026-02-13
Better tracking of subagent requests in capi request headers

roblourens:


src/extension/prompt/node/chatMLFetcher.ts
```diff
@@ -781,7 +787,10 @@ export class ChatMLFetcherImpl extends AbstractChatMLFetcher {
 		this._telemetryService.sendGHTelemetryEvent('request.sent', telemetryData.properties, telemetryData.measurements);
 
 		const requestStart = Date.now();
-		const intent = locationToIntent(location);
+		const intent = parentRequestId ? 'conversation-subagent' : locationToIntent(location);
```
TommasoPieroncini, line 790: Is this setting Openai-Intent header or X-Interaction-Type in the request to CAPI?

We want the latter I think

roblourens: Both. Today they are always the same. Is that ok?

TommasoPieroncini: I think it requires small changes in CAPI at 

https://github.com/github/copilot-api/blob/ace2427bb6e0ef0ac4b2937f9054f043959241c4/pkg/speculativedecoding/speculativedecoding.go#L23

and 

https://github.com/github/copilot-api/blob/ace2427bb6e0ef0ac4b2937f9054f043959241c4/pkg/promptcache/promptcache.go#L36


to avoid issues.

How long until this change rolls out in the client?

PR #3740, opened 2026-02-13
Enabling compaction for openAI models

dileepyavan:
**Details:**
To support long-running LLM interactions, a new mechanism in Responses for automatically managing context. This can be enabled compaction on a /responses call (via compact_threshold), and the server will compact the context when the rendered token count crosses the threshold*.  The compaction happens automatically—no separate /compact call—and the response 
stream will include the encrypted compaction item. The returned compaction item carries forward the key prior state and reasoning into the next run using far fewer tokens. It is opaque and not intended to be human‑interpretable.  To make it extremely easy to implement in any orchestration framework, the user can simply append their output items as they’d normally do. The compaction item generated here will carry all necessary information for the next context window.  [Recommendation] Clients who round trip all their input items are encouraged to prune all items that came before the most recent compaction item. This will avoid the size of the request growing infinitely and causing high latency.  

This PR includes
- Enables compaction for all the codex models and any future models.
- Configured under experimental flag.


src/extension/prompt/vscode-node/requestLoggerImpl.ts
```diff
@@ -54,21 +54,22 @@ function processDeltasToMessage(deltas: IResponseDelta[]): string {
 		}
 
 		// Handle context management
-		if (d.contextManagement) {
+		const appliedEdits = (d.contextManagement as { applied_edits?: Array<{ cleared_input_tokens?: number; cleared_tool_uses?: number; cleared_thinking_turns?: number }> } | undefined)?.applied_edits;
```
roblourens: What is this change?

src/platform/endpoint/node/responsesApi.ts
```diff
@@ -53,14 +56,21 @@ export function createResponsesRequestBody(accessor: ServicesAccessor, options:
 		text: verbosity ? { verbosity } : undefined,
 	};
 
+	const contextManagementEnabled = configService.getExperimentBasedConfig(ConfigKey.ResponsesApiContextManagementEnabled, expService) && !modelswithoutContextManagement.has(endpoint.family);
+	if (contextManagementEnabled) {
+		body.context_management = [{
+			'type': openAIContextManagementCompactionType,
+			'compact_threshold': 1000
```
roblourens: What are the units on this threshold?

src/extension/prompt/vscode-node/requestLoggerImpl.ts
dileepyavan: `contextManagement?: ContextManagementResponse | OpenAIContextManagementResponse;`

I think it can be better updated to type cast it as ContextManagementRespons instead of the above. will update

src/platform/endpoint/node/responsesApi.ts
dileepyavan: tokens, will add a comment

dileepyavan: One thing to note is not sure about the threshold value , currently set it to 5000 tokens when compcation will kick off, need to validate the performance using exp.

dileepyavan:
> Does this need a change so that we don't trigger local summarization

What is local summarization? I am not aware of this, how do I trigger this?

src/platform/endpoint/node/responsesApi.ts
roblourens: 5000 tokens? That's less than the tools/system prompt...

roblourens:
We have summarization today. It runs when the context window fills up.

bhavyaus:
How does this compare to our local summarization and do we have any control on the system prompt used for compaction? Would be good to comare this with Rob's built in compaction to see how it compares. 
Another thing to keep in mind which I ran into when context editing for Anthropic models: the thresholds/trigger values defined by the model providers is usually on the entire context window. Our math should account for the fact that we have explicity prompt/output limits.

dileepyavan:
> How does this compare to our local summarization and do we have any control on the system prompt used for compaction? Would be good to comare this with Rob's built in compaction to see how it compares. Another thing to keep in mind which I ran into when context editing for Anthropic models: the thresholds/trigger values defined by the model providers is usually on the entire context window. Our math should account for the fact that we have explicity prompt/output limits.

The new summarization is all handled at the open api side which is triggered based on the threshold value sent in the context_management parameter. When the threshold is met a new output type of "compaction" is returned in the response. For next requests, the compaction item will be the first in the messages and all the messages before this will not be appended. This keeps on updating the compaction item to hold the context size to be within threshold.

src/platform/endpoint/node/responsesApi.ts
dileepyavan: Compaction is applying only to items in the input(system/user/assistant prompts) property. Tools property have no impact. Posted a question to anoop on friday regarding this. Increase the token threshold to 50K . I am not very much sure on this number. When do we consider a session to be large, is it a percentage of the context window available for the model or tokens available to the user?

dileepyavan:
> Does this need a change so that we don't trigger local summarization?

Updated to disable when compaction is enabled

src/platform/endpoint/node/responsesApi.ts
roblourens: 50k also seems low. We have a large context window, we should be able to use most of it. You could describe it as a percentage of the space available. Or see what Codex does.

PR #3774, opened 2026-02-16
Add a maximum date of the comparison commit for the 1p repo telemetry info

IanMatthewHuff:
Context for this fix:

Our first party only repo info telemetry looks to create diffs between the public upstream and the current working tree of the local repo. It starts out by first doing a git diff to see just the names and counts of the files in the diff. Then, if it sees too many files, it exits early to avoid perf issues. We get this upstream branch for comparison by first looking for a set upstream tracking branch, and then if that fails using getMergeBase to get a comparison branch.

However in the case for some Windows devs, the tools they use to branch often don't set an upstream and the branch that they appear to have branched from is sometimes all the way back at the start of official/main. This is over four years old, and when git tries to just see the names and counts of files in the diff it churns up a huge git process causing performance issues.

This PR does two things.
1. Adds in an internal setting to provide a shutoff switch for this telemetry. If we hit some other degenerate case like this, we now have a way to unblock local users, and it could also be put into the settings file for a full repository.
2. When we find our upstream comparison commit, check the date on it, and only accept commits that are < 1 month old for comparison. This will rule out cases like this from happening. Any diff getting longer than that is very unlikely to fit inside the telemetry limits even if it gets past the file check. Return a new status code for this situation so we can see how often it's happening.

Testing:

- Unit tests added to test file for repo info telemetry
- Manual testing
  - With the setting not specified, data was collected
  - Setting set to true, data was not collected
  - Setting set to false, data was collected
  - Opened a repo that was branched > 1 month ago, did not collect with status message
  - Opened a more recent repo, data begin and end were collected as normal


IanMatthewHuff:
@sbatten @mjbvz Need anything from my end on this review? Just wanted to make sure that the windows folks hitting this issue get unblocked.

IanMatthewHuff:
@zhichli Here is the PR that I mentioned along with the PR here: https://github.com/microsoft/vscode-copilot-chat/pull/4018. It's merged with the changes now.

PR #3775, opened 2026-02-16
ghost: fix: do not reuse stale suffix

ulugbekna:
fixes https://github.com/microsoft/vscode/issues/295450

@dbaeumer would appreciate your review


dbaeumer:
@ulugbekna 

I looked at the change and it is unclear to me how often the fist token will change if the suffix changes (e.g. is this completely disables suffix caching since the first token changes by 99%)

I need to check this on my other machine but I think the proxy doesn't send the prompt anymore in form of suffix+prefix so this code might not be necessary anymore. Will report here.

dbaeumer:
I check the proxy code and caching the suffix still makes sense due to the following rendering logic:

```
{{ if gt (len .Context) 0 }}          ← Branch 1: Context exists
    Context + Suffix + [Additional Context] + Prefix

{{ else if .PrefixFirst }}            ← Branch 2: No context, prefix first
    Prefix + Suffix

{{ else }}                            ← Branch 3: No context, suffix first (default)
    Suffix + Prefix

{{ end }}
```

PrefixFirst is a deployment flag and it is false by default.

PR #3780, opened 2026-02-16
chore(): add telemetry event for auto mode routing fallback

Sid200026:
Added telemetry when the auto mode router errors out and fallback model is used.


TylerLeonhardt:
@Sid200026 aren't you more interested in error telemetry in this case?

TylerLeonhardt (approved):

PR #3804, opened 2026-02-17
Adding 90% of model's max window for compact_threshold

dileepyavan:
Currently compact_threshold is hardcoded. Updating it to be percentage of model max window size.


src/platform/endpoint/node/responsesApi.ts
```diff
@@ -58,10 +58,13 @@ export function createResponsesRequestBody(accessor: ServicesAccessor, options:
 
 	const contextManagementEnabled = configService.getExperimentBasedConfig(ConfigKey.ResponsesApiContextManagementEnabled, expService) && !modelsWithoutResponsesContextManagement.has(endpoint.family);
 	if (contextManagementEnabled) {
+		const compactThreshold = endpoint.modelMaxPromptTokens > 0
+			? Math.floor(endpoint.modelMaxPromptTokens * 0.9)
+			: 50000;
```
isidorn, line 63: If there is no maxPromptTokens let take a larger default. 100000 would be more fitting imho. 50000 is too small.

PR #3847, opened 2026-02-19
Fix broken file links in chat when model emits bare line anchors or Windows absolute paths

vijayupadya:
Fixes microsoft/vscode#271573

When models emit markdown links with bare line number anchors (e.g., #1287-1290 instead of #L1287-L1290) or absolute Windows paths (e.g., c:\src\file.ts), the linkifier failed to convert them to clickable file reference widgets, resulting in "Unable to open file" errors.

Two fixes:

- ModelFilePathLinkifier.parseAnchor: Made the L prefix optional in the anchor regex so bare line numbers like `#1287` and `#1287-1290` are parsed correctly and converted into proper LinkifyLocationAnchor widgets.

- Linkifier un-linkify pass: Changed the URI scheme regex from \w+: to \w{2,}: so single-letter Windows drive prefixes (.) are no longer mistaken for URI schemes. This ensures raw markdown links with Windows paths are stripped instead of passed through to the renderer.

### DIFF
--- a/src/extension/linkify/common/linkifier.ts
+++ b/src/extension/linkify/common/linkifier.ts
@@ -264,7 +264,7 @@ export class Linkifier implements ILinkifier {
 							return matched;
 						}
 
-						return /^\w+:/.test(path) ? matched : text;
+						return /^\w{2,}:/.test(path) ? matched : text;
 					});
 				}
 				return part;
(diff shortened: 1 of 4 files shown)


vijayupadya:
Hi @mjbvz, will you be able to help review this change?

PR #3858, opened 2026-02-19
fix ollam provider

sandy081:


grische:
@sandy081 As the commit is from #3684, imho it would have been fair to either cherry-pick or at least give the original author @lsby a "Co-Authored-By:" tag or mention in the commit message, as they are the original author of this fix.

riverar:
(And a friendly reminder that they retain copyright of their code submissions.)

harn3ss:
the only options i can choose from are stuck behind pay walls. typical

PR #3863, opened 2026-02-19
Implement MCP gateway for Claude agent

TylerLeonhardt:
This will hook up any MCP servers installed via VS Code and available via VS Code to Claude.

It think there should be some UX work in Core to enable/disable tools, but from an extension perspective this should be good.

ref https://github.com/microsoft/vscode/issues/287660

### DIFF
--- a/.vscode/mcp.json
+++ b/.vscode/mcp.json
@@ -3,7 +3,7 @@
 		"vscode-playwright-mcp": {
 			"type": "stdio",
 			"command": "npm",
-			// Look at the [README](../vscode/test/mcp/README.md) to see what arguments are supported
+			// See https://github.com/microsoft/vscode/blob/main/test/mcp/README.md for supported arguments
 			"args": [
 				"run",
 				"start-stdio",
--- a/src/platform/mcp/vscode/mcpServiceImpl.ts
+++ b/src/platform/mcp/vscode/mcpServiceImpl.ts
@@ -5,8 +5,9 @@
 
 import { type McpGateway, lm } from 'vscode';
 import { AbstractMcpService } from '../common/mcpService';
+import type { IDisposable } from '../../../util/vs/base/common/lifecycle';
 
-export class McpService extends AbstractMcpService {
+export class McpService extends AbstractMcpService implements IDisposable {
 	declare readonly _serviceBrand: undefined;
 
 	private cachedGateway: Promise<McpGateway | undefined> | undefined;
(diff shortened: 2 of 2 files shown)


src/extension/agents/claude/common/mcpServers/mcpGatewayServer.ts
```diff
@@ -0,0 +1,31 @@
...
+import { IClaudeMcpServerContributor, registerClaudeMcpServerContributor } from '../../common/claudeMcpServerRegistry';
+
+class McpGatewayServerContributor implements IClaudeMcpServerContributor {
+
+	constructor(
+		@IMcpService private readonly mcpService: IMcpService,
+	) { }
+
+	async getMcpServers(): Promise<Record<string, McpServerConfig>> {
+		const gateway = await this.mcpService.getMcpGateway();
```
connor4312, line 17: FYI in the future I will require you to pass the chat sessionResource so we know where to associate e.g. elicitation, sampling, and app requests

src/extension/agents/claude/common/mcpServers/mcpGatewayServer.ts
```diff
@@ -0,0 +1,31 @@
...
+		if (!gateway) {
+			return {};
+		}
+
+		return {
+			'vscode-mcp-gateway': {
+				type: 'http',
+				url: gateway.address.toString(),
```
connor4312, line 25: You'll want to dispose the gateway at some point too (especially once you do per-session)

TylerLeonhardt: gateway is now disposed with the service. It does make me wonder when the best time to dispose will be when it's per-session.

src/extension/agents/claude/common/mcpServers/mcpGatewayServer.ts
TylerLeonhardt: Makes sense, when that comes, we can plumb that through

src/extension/agents/claude/common/mcpServers/mcpGatewayServer.ts
connor4312: It's quite cheap to create, maintain, and tear down gateways. We have a singleton HTTP server and each gateway is just a route on that server. So you could have some handle that you dispose if the session isn't active for a minute or few.

PR #3864, opened 2026-02-19
search subagent -> fine-tuned model flighting

24anisha:
Improves the search subagent configuration by consolidating model settings, fixing experiment-based config usage, and enhancing the search subagent prompt with current working directory.

#### Consolidated Search Subagent Model Configuration:
Removed the separate AgenticProxySearchModelName config in favor of using SearchSubagentModel for both proxy and non-proxy modes. Also updated ProxyAgenticSearchEndpoint to accept model name as a constructor parameter.

When useAgenticProxy is enabled:
uses SearchSubagentModel if specified
Defaults to 'agentic-search-v3' if not specified

When useAgenticProxy is disabled:
Uses SearchSubagentModel if specified
Falls back to the main agent model if not specified

#### Fixed Experiment-Based Config Usage
Changed `SearchSubagentUseAgenticProxy` to use `getExperimentBasedConfig` instead of `getConfig` to prepare for upcoming custom model flight

#### Enhanced Search Subagent Context
Added current working directory to the search subagent's first user prompt, since the search subagent needs to provide absolute paths for tool calls & its final answer

#### Disabled ReadFile V2 by Default
Changed EnableReadFileV2 default from isPreRelease to false
This means the readFile tool uses startLine/endLine parameters instead of offset/limit by default
Note: this change was okayed by @connor4312. We needed to change it so that our experiment uses a consistent read file tool between insiders and public


24anisha:
@bhavyaus @roblourens @connor4312

TylerLeonhardt (approved):

PR #3880, opened 2026-02-20
Remove Ask User Question tool in favor of Core tool

TylerLeonhardt:
Fixes https://github.com/microsoft/vscode/issues/296134

ref https://github.com/microsoft/vscode/pull/296135

### DIFF
--- a/src/platform/networking/common/anthropic.ts
+++ b/src/platform/networking/common/anthropic.ts
@@ -93,7 +93,7 @@ export const nonDeferredToolNames = new Set([
 	// Testing
 	'runTests',
 	// Misc
-	'vscode_askQuestions',
+	'ask_questions',
 	'switch_agent',
 	'memory'
 ]);


src/platform/networking/common/anthropic.ts
```diff
@@ -93,7 +93,7 @@ export const nonDeferredToolNames = new Set([
 	// Testing
 	'runTests',
 	// Misc
-	'ask_questions',
+	'vscode_askQuestions',
```
TylerLeonhardt: @bhavyaus does this make sense?

TylerLeonhardt: it use to be `ask_questions` but I wonder if it should have been either:
* `askQuestions` -> which is is how it can be referenced in the prompt
* `copilot_askQuestions` -> which is now `vscode_askQuestions` which is the actual id of the tool

bhavyaus: The proxyman request still shows ask_questions as the tool in the request? Does claude agent tool name differ?

dmitrivMS (approved):

PR #3930, opened 2026-02-23
Add project setup skill

alexr00:


assets/prompts/skills/project-setup-info-local/SKILL.md
```diff
@@ -0,0 +1,127 @@
+---
+name: project-setup-info-local
+description: 'Provides information about how to setup various kinds of projects. Useful for initializing projects based on requirements and scaffolding the project.'
+---
+
+# How to setup a project
+
+Determine what kind of project the user wants to create, then based on that, choose which setup info below to follow. Only setup a project if the folder is empty or if you've just called the tool first calling the tool to create a workspace.
+
+## vscode-extension
```
OrenMe, line 10: Suggestion break down this file to various smaller files so not all of it gets loaded cause there will always be just one type of project relevant 
So use this file as router to load full details per each project type
Will also be easier to maintain IMO

PR #3947, opened 2026-02-23
Add localhost embeddings support for scenario automation

gryan11:
Updates to support semantic search with blackbird in MSBench. MSBench test run here: https://msbenchapp.azurewebsites.net/run-analysis/22452275228

- ScenarioAutomationCAPIClientImpl: intercepts EmbeddingsCodeSearch requests, strips CAPI auth headers, sends clean JSON to localhost:4443
- ScenarioAutomationGithubCodeSearchService: overrides getRemoteIndexState to always return Ready, bypassing production index checks
- Register MockGithubAvailableEmbeddingTypesService in scenario automation to use metis_1024_I16_Binary without hitting production endpoint
- Move all scenario-automation service overrides into the isScenarioAutomation if/else branch in services.ts


src/platform/endpoint/node/scenarioAutomationCAPIClientImpl.ts
```diff
@@ -0,0 +1,35 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { FetchOptions, RequestMetadata, RequestType } from '@vscode/copilot-api';
+import { IEnvService } from '../../env/common/envService';
+import { IFetcherService } from '../../networking/common/fetcherService';
+import { CAPIClientImpl } from './capiClientImpl';
+
+const SCENARIO_AUTOMATION_CODE_SEARCH_URL = 'https://localhost:4443/embeddings/code/search';
```
rwoll: @gryan11 - Avoid hard coding this. I suggest adding a formal vscode setting, marked as experimental, that allows you to set the URL (and potentially specify an environment variable for auth).

By doing so, you'll be able to open vscode (outside test scenarios) and then configure it to use your local or (remote) staging endpoint to dogfood and debug.

There's similar settings: see `github.copilot.advanced.debug.overrideCapiUrl` and `github.copilot.advanced.debug.overrideProxyUrl` for an example. Essentially, you want a similar feature.

rwoll:
@gryan11 - ping me on teams when you need another review.

src/platform/workspaceChunkSearch/node/scenarioAutomationWorkspaceChunkSearchService.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * In scenario automation the copilot token may not be available at construction
+ * time and remote code-search APIs may be unreachable.  This override ensures
+ * {@link getIndexState} never reports both local *and* remote indexes as
+ * disabled, which would cause callers (e.g. `WorkspaceChunks.prepare()`) to
+ * skip the search entirely.  When both would otherwise be disabled the remote
+ * status is reported as `initializing` so that the search pipeline proceeds
+ * to the TF-IDF / Blackbird fallback strategies.
+ */
+export class ScenarioAutomationWorkspaceChunkSearchService extends WorkspaceChunkSearchService {
```
mjbvz: Instead of extending `extends WorkspaceChunkSearch` (which then requires also overriding `GithubCodeSearchService`, can you implement the `IWorkspaceChunkSearchService` entirely here. Most of the operations should be noops

gryan11: Extending WorkspaceChunkService is a design choice to ensure the only change in MSBench runs is using the local blackbird endpoint and otherwise all production behavior is being evaluated correctly. Otherwise, we can't guarantee MSBench evaluations are using the full production behavior for vscode agent.

Even if the current logic in WorkspaceChunkService will be functionally equivalent in MSBench evaluations now if it is removed, we can't guarantee that will always be the case in the future.

mjbvz: I'd like to keep them separate so that I don't have to worry about refactoring `WorkspaceChunkSearchService`

Look at the existing `SimulationCodeSearchChunkSearchService`. It's doing the same basic thing but also just implements the `IWorkspaceChunkSearchService` interface

gryan11: Updated to address this.

src/platform/configuration/common/configurationService.ts
```diff
@@ -591,6 +591,9 @@ export namespace ConfigKey {
 		 * Note: this should not be used while self-hosting because it might lead to
 		 * a fundamental different experience compared to our end-users.
 		*/
+		export const DebugOverrideChatEngine = defineAndMigrateSetting<string | undefined>('chat.advanced.debug.overrideChatEngine', 'chat.debug.overrideChatEngine', undefined);
```
mjbvz: These are only used for automation, right? Do we need these settings then or can the ScenarioAutomation chunk search just hardcode values?

src/platform/authentication/common/staticGitHubAuthenticationService.ts
```diff
@@ -64,6 +64,23 @@ export class StaticGitHubAuthenticationService extends BaseAuthenticationService
...
+	/**
+	 * When an embeddings override URL is configured (e.g. Blackbird local
+	 * server), return a placeholder token instead of `undefined` or a noAuth
+	 * token.  This allows services like WorkspaceChunkSearchService to
+	 * initialise and route embeddings requests through the override URL.
+	 * Without an override URL the production behaviour is preserved.
+	 */
+	override get copilotToken(): CopilotToken | undefined {
+		const token = this._tokenStore.copilotToken;
+		if (this._configurationService.getConfig(ConfigKey.Advanced.DebugOverrideEmbeddingsUrl)) {
```
mjbvz: I feel like this affects much more than embeddings so it's confusing to have `DebugOverrideEmbeddingsUrl`. Maybe give that setting a better name or see if we can avoid needing it altogether?

mjbvz (dismissed):

src/platform/configuration/common/configurationService.ts
gryan11: Updated to remove this.

src/platform/authentication/common/staticGitHubAuthenticationService.ts
gryan11: Updated to hardcode in chunk search.

src/platform/authentication/common/staticGitHubAuthenticationService.ts
```diff
@@ -64,6 +64,19 @@ export class StaticGitHubAuthenticationService extends BaseAuthenticationService
...
 
+	/**
+	 * In scenario automation the copilot token may be a noAuth token.
+	 * Return a placeholder so that services checking `copilotToken?.isNoAuthUser`
+	 * (e.g. {@link ScenarioAutomationEndpointProviderImpl}) behave correctly.
+	 */
+	override get copilotToken(): CopilotToken | undefined {
+		const token = this._tokenStore.copilotToken;
+		if (token?.isNoAuthUser) {
+			return new CopilotToken(createTestExtendedTokenInfo());
```
mjbvz: This will break real no-auth users in production code, right? I'd like to understand why this is needed. I don't think we should do this if it's only for `ScenarioAutomationEndpointProviderImpl` Is there a better approach that keeps this isolated to just the test code?

gryan11: This isn't needed since msbench env does auth, I've removed it.

package-lock.json
rwoll, line 1: @gryan11 - can you revert the changes to this file? There don't appear to be any package.json changes, so I don't expect this lock file to change.

rwoll (approved):

rwoll:
Moved to microsoft/vscode#308394 (the copilot chat extension now lives in the vscode repo).

PR #3963, opened 2026-02-24
 Fix intermittent null copilotTrackingId in telemetry events

vijayupadya:
**Summary:** TelemetryUserConfigImpl only populated trackingId from the Copilot token via the onDidStoreUpdate listener, missing the case where the token was already available at construction time. This caused copilot.trackingId to be null in early telemetry events (e.g., query-expfeature) until the next token refresh.

The fix eagerly reads the current token in the constructor, eliminating the race between token acquisition and telemetry emission.


ulugbekna:
@vijayupadya oh so glad I got to review this because I spent a couple of hours last week investigating why so many events do not have copilotTrackingId

PR #3979, opened 2026-02-24
Aggressivity settings 

benibenj:
```Copilot Generated Description:``` Introduce aggressivity settings for next edit suggestions, allowing users to control the frequency and delay of suggestions. Add telemetry to track user aggressiveness settings and their impact on edit suggestions.

cc @ulugbekna 

https://github.com/microsoft/vscode-internalbacklog/issues/6771

### DIFF
--- a/package.json
+++ b/package.json
@@ -4351,6 +4351,7 @@
 						],
 						"markdownDescription": "%github.copilot.config.nextEditSuggestions.aggressiveness%",
 						"tags": [
+							"advanced",
 							"experimental"
 						]
 					},

### CODE REVIEW

File: src/platform/configuration/common/configurationService.ts [hunk 1]
```diff
@@ -704,6 +704,7 @@ export namespace ConfigKey {
 		export const InlineEditsNextCursorPredictionCurrentFileMaxTokens = defineAndMigrateExpSetting<number>('chat.advanced.inlineEdits.nextCursorPrediction.currentFileMaxTokens', 'chat.inlineEdits.nextCursorPrediction.currentFileMaxTokens', xtabPromptOptions.DEFAULT_OPTIONS.currentFile.maxTokens);
 		export const InlineEditsRenameSymbolSuggestions = defineSetting<boolean>('chat.inlineEdits.renameSymbolSuggestions', ConfigType.ExperimentBased, true);
 		export const InlineEditsPreferredModel = defineSetting<string | 'none'>('nextEditSuggestions.preferredModel', ConfigType.ExperimentBased, 'none');
+		export const InlineEditsAggressiveness = defineSetting<string>('nextEditSuggestions.aggressiveness', ConfigType.ExperimentBased, 'default');
```
ulugbekna, line 707: let's use an enum to avoid casting and add a validator to get intellisense in settings.json? see eg `InlineEditsSpeculativeRequests` config

src/extension/inlineEdits/common/userInteractionMonitor.ts
```diff
@@ -251,12 +251,21 @@ export class UserInteractionMonitor {
 	 * The score is returned to avoid race conditions when logging telemetry.
 	 */
 	public getAggressivenessLevel(): { aggressivenessLevel: AggressivenessLevel; userHappinessScore: number | undefined } {
+		// User-facing setting takes priority when explicitly set to a non-default value
+		const userAggressiveness = this._configurationService.getExperimentBasedConfig(ConfigKey.Advanced.InlineEditsAggressiveness, this._experimentationService);
+		const userLevel = AggressivenessSetting.toLevel(userAggressiveness as AggressivenessSetting);
```
ulugbekna, line 256: we wouldn't need `as` casting if we used an enum for the setting

src/extension/xtab/node/xtabProvider.ts
```diff
@@ -245,6 +245,24 @@ export class XtabProvider implements IStatelessNextEditProvider {
 			tracer.trace('No extra debounce applied');
 		}
 
+		// Adjust debounce based on user aggressiveness setting for non-aggressiveness models
+		if (!isAggressivenessStrategy(promptOptions.promptingStrategy)) {
```
ulugbekna, line 249: 1. should this add more debounce if there's already extra debounce (because of end of line or something)? 
2. let's have a util method? it's lots of mundane code to put in a function that has important logic to trace through

PR #3980, opened 2026-02-24
Add Copilot auth token to router decision fetcher for CAPI proxy auth

aashna:


src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -59,14 +61,18 @@ export class RouterDecisionFetcher extends Disposable {
...
 		}
 
+		// Get Copilot auth token to authenticate with CAPI router endpoint
+		const authToken = (await this._authService.getCopilotToken()).token;
+
 		let lastError: Error | undefined;
 		for (let attempt = 0; attempt < MAX_RETRIES; attempt++) {
 			let response: Response;
 			try {
 				response = await this._fetcherService.fetch(routerApiUrl, {
```
TylerLeonhardt, line 82: The URI here is user configured. If they put something that isn't a GitHub url, their Copilot token will be still sent there.

We should check if the url is a GitHub url and only send the copilot token in that case

PR #4010, opened 2026-02-25
Do not prompt for permissions when editing files in workspace folder

DonJayamanne:
Fixes https://github.com/microsoft/vscode/issues/297809

### DIFF
--- a/src/extension/tools/node/editFileToolUtils.tsx
+++ b/src/extension/tools/node/editFileToolUtils.tsx
@@ -795,11 +795,12 @@ export const enum ConfirmationCheckResult {
 	OutsideWorkspace,
 }
 
+
 /**
  * Returns a function that returns whether a URI is approved for editing without
  * further user confirmation.
  */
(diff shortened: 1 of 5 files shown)


src/extension/tools/node/editFileToolUtils.tsx
```diff
@@ -923,7 +924,9 @@ export async function createEditConfirmation(accessor: ServicesAccessor, uris: r
 		};
 	}
 
-	const checker = makeUriConfirmationChecker(accessor.get(IConfigurationService), accessor.get(IWorkspaceService), accessor.get(ICustomInstructionsService));
+	const workspaceService = accessor.get(IWorkspaceService);
+	getWorkspaceFolder = getWorkspaceFolder ?? workspaceService.getWorkspaceFolder.bind(workspaceService);
```
DonJayamanne, line 928: In the case of welcome view or Sessions Window, we don't have workspace folders.
But a folder selected by user is considered `workspaceFolder`, hence if user attempts to edit/create files in the folder that was selected as the root for the agent session, then we consider that as a workspace folder.

rebornix (approved):

PR #4018, opened 2026-02-26
fix: remove repoId from MSFT telemetry and optimize repo info for external users

zhichli:
Follow-up to #3956:

- Remove `repoId` from external MSFT telemetry (sensitive data)
- Gate diff computation behind `isInternal` check — external users only get lightweight metadata (`repoType` + `headCommitHash`)
- Extract shared `_resolveRepoContext()` to deduplicate repo context resolution
- Update all 34 tests for new `copilotTokenStore` parameter

### DIFF
--- a/src/extension/prompt/node/repoInfoTelemetry.ts
+++ b/src/extension/prompt/node/repoInfoTelemetry.ts
@@ -7,7 +7,7 @@ import { ICopilotTokenStore } from '../../../platform/authentication/common/copi
 import { IFileSystemService } from '../../../platform/filesystem/common/fileSystemService';
 import { IGitDiffService } from '../../../platform/git/common/gitDiffService';
 import { IGitExtensionService } from '../../../platform/git/common/gitExtensionService';
-import { getOrderedRepoInfosFromContext, IGitService, normalizeFetchUrl, ResolvedRepoRemoteInfo } from '../../../platform/git/common/gitService';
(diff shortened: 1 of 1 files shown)


karthiknadig (dismissed):

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -137,47 +139,48 @@ export class RepoInfoTelemetry {
...
-		// Send full data including remoteUrl and diffs for internal users only
-		const internalProperties: RepoInfoInternalTelemetryProperties = {
-			...repoInfo.properties,
-			location,
-			telemetryMessageId: this._telemetryMessageId
-		};
-		this._telemetryService.sendInternalMSFTTelemetryEvent('request.repoInfo', internalProperties, repoInfo.measurements);
+		const metadata = await this._getRepoMetadata();
+		if (metadata) {
+			this._telemetryService.sendMSFTTelemetryEvent('request.repoInfo', {
```
IanMatthewHuff, line 162: Quick question (that I don't know the answer to off hand) would you need a new name for this event? They might go into totally different tables, which would be ok. Just don't want the folks watching the existing request.repoInfo to get confused by a new event with less properties / data.

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -137,47 +139,48 @@ export class RepoInfoTelemetry {
...
+				repoType: metadata.repoType,
+				headCommitHash: metadata.headCommitHash,
```
IanMatthewHuff, line 164: Is this still useful information without the id?

PR #4027, opened 2026-02-26
show stdout for hooks as well

justschen:
fix https://github.com/microsoft/vscode-internalbacklog/issues/6894
fix https://github.com/microsoft/vscode-internalbacklog/issues/6791

Screenshot 2026-02-26 at 12 14 37 AM

562-Yamask

### DIFF
--- a/src/extension/chat/vscode-node/chatHookService.ts
+++ b/src/extension/chat/vscode-node/chatHookService.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as l10n from '@vscode/l10n';
 import type * as vscode from 'vscode';
 import { IChatHookService, IPostToolUseHookResult, IPreToolUseHookResult } from '../../../platform/chat/common/chatHookService';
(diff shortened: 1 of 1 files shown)


src/platform/chat/node/hookExecutor.ts
```diff
@@ -147,7 +147,25 @@ export class NodeHookExecutor implements IHookExecutor {
 					resolve({ kind: HookCommandResultKind.Success, result });
 				} else if (code === 2) {
 					// Exit code 2: blocking error shown to model
-					resolve({ kind: HookCommandResultKind.Error, result: stderrStr });
+					// Try to extract permissionDecisionReason from structured stdout
+					let permissionReason: string | undefined;
+					if (stdoutStr) {
```
justschen: not sure what we want to do in the case we get both stdout and stderr

pwang347: I think we should continue to follow the semantics for exit codes; code 0 = we look at stdout, anything else we only look at stderr

src/extension/chat/vscode-node/chatHookService.ts
```diff
@@ -19,6 +19,7 @@ import { IToolsService, isToolValidationError } from '../../tools/common/toolsSe
 import { ChatHookTelemetry } from './chatHookTelemetry';
 
 const permissionPriority: Record<string, number> = { 'deny': 2, 'ask': 1, 'allow': 0 };
+const genericHookErrorMessage = 'an unexpected error occurred';
```
pwang347: nit: does this need to be localized since it's user facing? (in which case, probably need to duplicate the string so translators can translate the join)

PR #4033, opened 2026-02-26
/create-agent uses double-quoted tool names

aeschli:
Fixes https://github.com/microsoft/vscode/issues/297444


aeschli:
@digitarald FYI

alexr00 (approved):

PR #4042, opened 2026-02-26
fix question tool name in agent providers

meganrogge:
fix https://github.com/microsoft/vscode/issues/297978


meganrogge:
changed it back to be in vscode toolset vs agent, so these changes are not needed

PR #4071, opened 2026-02-27
Kevin m kent/v2 domain intent taxonomy

kevin-m-kent:
PR that builds up domain and intent from prompt clustering and labeling in our 1p data. Also adds some prompting that improves the classifier's error rate (in particular swapping domain and intent). This approach also reduces the number of categories by about 50%. 

FYI @digitarald , pending our discussion later.


src/extension/prompt/common/promptCategorizationTaxonomy.ts
```diff
@@ -4,250 +4,72 @@
...
- * Types, validation, and prompt generation all derive from this.
+ * Domain + Intent + Scope + Time Estimate classification taxonomy.
+ *
+ * Single source of truth for the domain, intent, and scope categories (derived from
+ * clustering analysis) and time estimate dimensions.
  */
 
 // ============================================================================
 // INTENTS - What action the user wants
 // ============================================================================
```
digitarald, line 15: nit (non-blocking, follow-up): The `promptCategorization` telemetry event in `promptCategorizer.ts` emits `intent` and `domain` as raw strings. Since v2 keys are completely different from v1 (e.g., `code_fixing` → `troubleshoot_debug`, `frontend` → `web_ui`), queries on this data will get mixed v1/v2 values with no way to distinguish them once this ships.

Suggestion: Add a `taxonomyVersion: 'v2'` property to the telemetry event and its GDPR annotation so dashboards can filter correctly.

src/extension/prompts/node/panel/promptCategorization.tsx
```diff
@@ -19,7 +19,7 @@ export interface PromptCategorizationProps extends BasePromptElementProps {
 export class PromptCategorizationPrompt extends PromptElement<PromptCategorizationProps> {
 	override async render(_state: void, sizing: PromptSizing) {
 		const systemPrompt = [
-			'You are an expert intent classifier for AI coding assistants. Classify developer requests in context of their workspace and active file across four dimensions: intent, domain, time estimate, and scope.',
+			'You are an expert classifier for AI coding assistant prompts. Classify developer requests in context of their workspace and active file across domain, intent, time estimate, and scope.',
 			'You MUST use the categorize_prompt tool to provide your classification.',
 			generateTaxonomyPrompt(),
 		].join('\n\n');
```
digitarald, line 25: nit (non-blocking): `systemPrompt` is joined with `\n\n` but then rendered immediately before `<SafetyRules />` with no separator, so the last line of the taxonomy can run together with the safety rules text.

Consider adding a trailing `\n\n`:
```ts
].join('\n\n') + '\n\n';
```
Or a `<br />` before `<SafetyRules />` in the JSX.

src/extension/prompt/common/promptCategorizationTaxonomy.ts
```diff
@@ -468,10 +271,10 @@ export interface PromptClassification {
...
 function formatCategoryForPrompt(key: string, def: CategoryDefinition): string {
-	const parts = [`## ${key}`, def.description];
+	const parts = [`### \`${key}\``, def.description];
 
 	if (def.keywords?.length) {
-		parts.push(`Keywords: ${def.keywords.map(k => `"${k}"`).join(', ')}`);
+		parts.push(`- Keywords: ${def.keywords.join(', ')}`);
 	}
 	if (def.signals?.length) {
 		parts.push(`Signals: ${def.signals.join(', ')}`);
```
digitarald, line 280: nit (non-blocking): `def.signals` rendering is now dead code — no v2 definitions use `signals` anymore (all switched to `keywords`). Could be cleaned up in a follow-up.

src/extension/prompt/common/promptCategorizationTaxonomy.ts
kevin-m-kent: I like this idea. Let me update.

Yoyokrazy (approved):

PR #4079, opened 2026-02-27
Remove duplicated memory tool in picker

sbatten:
fixes https://github.com/microsoft/vscode/issues/297010

### DIFF
--- a/package.json
+++ b/package.json
@@ -1074,7 +1074,6 @@
 				"name": "copilot_memory",
 				"displayName": "Memory",
 				"toolReferenceName": "memory",
-				"canBeReferencedInPrompt": true,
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1206,7 +1206,6 @@
 				"tools": [
 					"getProjectSetupInfo",
 					"installExtension",
-					"memory",
```
roblourens, line 1209: It should be in this toolset, not at top-level, we can't add new top-level tools

roblourens (changes requested):

package.json
roblourens: I think you just want to get rid of `canBeReferencedInPrompt` (yeah this is dumb)

PR #4084, opened 2026-02-28
Palette a11y aria busy 2749993949822189633

mexicodxnmexico-create:


aiday-mar:
Hi could you please indicate here what is the issue this PR aims to fix?

PR #4142, opened 2026-03-03
implement mcp gateway for cli (gated behind sessions flag)

joshspicer:


TylerLeonhardt:
FYI the gateway flow can be very slow if you have many MCP servers. It added a whole minute to claude startup.

there's some work going in to make it faster but we're not resolved there yet https://github.com/microsoft/vscode/pull/298040

make sure perf is ok

joshspicer:
thanks @TylerLeonhardt - chatted with connor. This is ok to unblock the sessions app (it's gated to just sessions app)

PR #4203, opened 2026-03-04
Add image hard gate to bypass router for image requests

aashna:
When a chat request contains an image attachment, bypass the model router entirely and use the heuristic-based path (_resolveWithoutRouterModel) which already handles vision-capable model selection via _applyVisionFallback.

The router model is text-only and cannot reason about image content, so sending image requests through the router would produce unreliable routing decisions. Instead, the existing vision fallback heuristic is the correct path.

Changes in automodeService.ts:
- Check hasImage(chatRequest) before deciding to use router
- When image detected and router configured, emit automode.imageHardGate telemetry event and fall through to heuristic path
- usingRouterModel = routerConfigured && !requestHasImage

Changes in routerDecisionFetcher.ts:
- Add optional hasImage parameter to getRoutedModel()
- Pass has_image: true in request body when set (defense-in-depth for server-side gating in auto-intent-service)

Tests:
- Add test: should send has_image flag when provided
- Add test: should not send has_image when not provided

### DIFF
--- a/src/platform/endpoint/node/automodeService.ts
+++ b/src/platform/endpoint/node/automodeService.ts
@@ -195,7 +195,30 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 
 		// Only use router model for panel chat to avoid latency penalty in inline chat
 		const isPanelChat = !chatRequest?.location || chatRequest?.location === ChatLocation.Panel;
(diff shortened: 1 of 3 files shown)


lramos15:
I believe we've decided this one isn't needed as we will fallback to the client's image safety logic which seems to be working well

PR #4219, opened 2026-03-04
Exempt Search Subagent tool result from write to disk

24anisha:
Include full results found by the search subagent in the main agent model trajectory. Do not write the results to disk, just because it is large.

Reason: we see that the model only reads the first ~200 lines of the file and never really goes back to read more, which functionally makes all the work done by the search subagent redundant and useless.


src/extension/prompts/node/panel/toolCalling.tsx
```diff
@@ -743,7 +747,7 @@ export class ToolResult extends PrimitiveToolResult<IToolResultProps> {
 			this._experimentationService
 		);
 
-		if (isDiskCachingEnabled && this.diskSessionResources && this.props.toolCallId && this.props.sessionId) {
+		if (isDiskCachingEnabled && this.diskSessionResources && this.props.toolCallId && this.props.sessionId && this.props.toolName !== ToolName.SearchSubagent) {
```
connor4312, line 750: Please add an explainer comment here for future reference :)

PR #4231, opened 2026-03-05
Add tests for toInlineSuggestion prefix stripping logic and advanced argument

benibenj:
```Copilot Generated Description:``` Introduce tests for the prefix stripping logic in the `toInlineSuggestion` function and add an advanced argument with a default value to control this behavior.

//cc @ulugbekna


mjbvz:
@benibenj Does this need to go into 1.111 or should it be pushed to 1.112?

PR #4234, opened 2026-03-05
try to be more loose on when autopilot stops

justschen:
don't immediately stop if we've already nudged once.


meganrogge:
should this have a tracking issue?

pwang347 (approved):

meganrogge (approved):

PR #4238, opened 2026-03-05
Add interactionId to response telemetry events

kevin-m-kent:
With sub agents there is a divergence between a turn and user interaction (there are also edge cases like max tool call limits where it spawns a new requestid but is the same interaction_id header). I think it is going to make sense to re-calibrate our metrics around interaction-id grouping rather than requestid to capture the semantic intent behind "turn" in our current metrics. 

This PR adds interactionId property to response.success, response.cancelled, and response.error telemetry, panel.request, and inline.request events to address that gap. I know that on paper we could join up via parentrequestid but for experiment metrics and other contexts that can get infeasible (especially if there are multiple levels of nesting).


kevin-m-kent:
@roblourens curious your thoughts

roblourens:
Note that interaction id is still not really computed right- sessions running at the same time will end up sharing an interaction id. I plan to do another round of work on that soon.

kevin-m-kent:
> Note that interaction id is still not really computed right- sessions running at the same time will end up sharing an interaction id. I plan to do another round of work on that soon.

Ah good to know. I will work on this PR still but maybe it shouldn't be used in scorecards etc until your work is completed.

PR #4267, opened 2026-03-06
Send context signals in routing predict request

aashna:
Forward optional context fields from the client to the routing service to enable richer routing decisions in future model versions.

Signals sent:
- chat_location (panel/editor/terminal)
- session_id (conversation identifier)
- has_image (image attachment present)
- reference_count (number of #file / @workspace references)
- prompt_char_count (character length of the current prompt)
- previous_model (model used on the prior turn)
- turn_number (position in conversation)

All fields are optional and the server silently ignores unknown fields. Current models do not use these signals.

### DIFF
--- a/src/platform/endpoint/node/automodeService.ts
+++ b/src/platform/endpoint/node/automodeService.ts
@@ -20,7 +20,7 @@ import { IExperimentationService } from '../../telemetry/common/nullExperimentat
 import { ITelemetryService } from '../../telemetry/common/telemetry';
 import { ICAPIClientService } from '../common/capiClient';
 import { AutoChatEndpoint } from './autoChatEndpoint';
-import { RouterDecisionFetcher } from './routerDecisionFetcher';
+import { RouterDecisionFetcher, RoutingContextSignals } from './routerDecisionFetcher';
 
 interface AutoModeAPIResponse {
 	available_models: string[];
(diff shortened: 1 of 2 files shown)


src/platform/endpoint/node/automodeService.ts
```diff
@@ -230,7 +230,16 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 				// Router fallback reason isn't set here because we don't want telemetry for this case
 			} else {
 				try {
-					const result = await this._routerDecisionFetcher.getRouterDecision(prompt, token.session_token, token.available_models);
+					const contextSignals: RoutingContextSignals = {
+						chat_location: chatRequestLocationToString(chatRequest?.location),
```
lramos15, line 234: The router always runs in panel right now

src/platform/endpoint/node/automodeService.ts
```diff
@@ -230,7 +230,16 @@ export class AutomodeService extends Disposable implements IAutomodeService {
...
+						session_id: conversationId !== 'unknown' ? conversationId : undefined,
+						has_image: hasImage(chatRequest),
```
lramos15, line 236: We will never run the router with images so this will always be false

src/platform/endpoint/node/automodeService.ts
```diff
@@ -230,7 +230,16 @@ export class AutomodeService extends Disposable implements IAutomodeService {
...
+						reference_count: chatRequest?.references?.length,
+						prompt_char_count: prompt.length,
+						previous_model: entry?.endpoint?.model,
+						turn_number: entry ? 2 : 1,
```
lramos15, line 240: this can only be 2 or 1? Can't there be 100 turns

lramos15:
Closing in favor of https://github.com/microsoft/vscode-copilot-chat/pull/4377

PR #4271, opened 2026-03-07
fix: populate valid fields on invalidClassification instead of blanking all

kevin-m-kent:
## Problem

When prompt categorization encounters an `invalidClassification` (one or more core fields fail validation), **all** telemetry fields are left blank — even fields that parsed and validated successfully.

@digitarald 

## Fix

- Added `extractBestEffortFields()` which validates each field independently, populating whichever fields are valid
- Per-field error reasons are tracked and included in `errorDetail` for diagnostics (e.g. `domain: expected string, got number; scope: invalid value "foo_bar"`)
- Confidence uses `-1` sentinel instead of `0` when it cannot be determined (since `0` is a valid confidence score)
- Updated GDPR comments to reflect new field population behavior
- Outcome remains `invalidClassification` — the populated fields themselves provide signal on severity

### DIFF
--- a/src/extension/prompt/node/promptCategorizer.ts
+++ b/src/extension/prompt/node/promptCategorizer.ts
@@ -18,7 +18,7 @@ import { isCancellationError } from '../../../util/vs/base/common/errors';
 import { IInstantiationService } from '../../../util/vs/platform/instantiation/common/instantiation';
 import { renderPromptElement } from '../../prompts/node/base/promptRenderer';
 import { PromptCategorizationPrompt } from '../../prompts/node/panel/promptCategorization';
-import { CATEGORIZE_PROMPT_TOOL_NAME, CATEGORIZE_PROMPT_TOOL_SCHEMA, isValidDomain, isValidIntent, isValidScope, PromptClassification } from '../common/promptCategorizationTaxonomy';
+import { CATEGORIZE_PROMPT_TOOL_NAME, CATEGORIZE_PROMPT_TOOL_SCHEMA, isValidDomain, isValidIntent, isValidScope, PromptClassification, PromptDomain, PromptIntent, PromptScope } from '../common/promptCategorizationTaxonomy';
 
 /** Experiment flag to enable prompt categorization */
 const EXP_FLAG_PROMPT_CATEGORIZATION = 'copilotchat.promptCategorization';
@@ -82,10 +82,11 @@ function extractPartialClassification(obj: unknown): PromptClassification | unde
 	if (!fields.intent || !fields.domain || !fields.scope || fields.confidence === -1 || !fields.reasoning) {
 		return undefined;
 	}
+	// Fields are validated by extractBestEffortFields via isValidIntent/isValidDomain/isValidScope
 	return {
-		intent: fields.intent,
-		domain: fields.domain,
-		scope: fields.scope,
+		intent: fields.intent as PromptIntent,
+		domain: fields.domain as PromptDomain,
+		scope: fields.scope as PromptScope,
 		confidence: fields.confidence,
 		reasoning: fields.reasoning,
 		timeEstimate: { bestCase: fields.timeEstimateBestCase, realistic: fields.timeEstimateRealistic },


src/extension/prompt/node/promptCategorizer.ts
```diff
@@ -117,6 +117,102 @@ function extractPartialClassification(obj: unknown): PromptClassification | unde
...
+	}
+
+	if (typeof c.timeEstimate === 'object' && c.timeEstimate !== null) {
+		const te = c.timeEstimate as Record<string, unknown>;
+		if (typeof te.bestCase === 'string' && isValidIsoDuration(te.bestCase)) {
+			fields.timeEstimateBestCase = te.bestCase;
+		}
+		if (typeof te.realistic === 'string' && isValidIsoDuration(te.realistic)) {
+			fields.timeEstimateRealistic = te.realistic;
+		}
```
digitarald, line 197: The `timeEstimate` fields are silently extracted without being tracked in `recovered` or `errors`, unlike every other field. This defeats the diagnostic purpose of the function — if you're investing in per-field error tracking, the time estimate fields should participate too.

Agree with the Copilot reviewer's suggestion to add `recovered.push(...)` / `errors.push(...)` for both `timeEstimateBestCase` and `timeEstimateRealistic`, plus an outer error entry when `timeEstimate` isn't an object at all.

src/extension/prompt/node/promptCategorizer.ts
```diff
@@ -117,6 +117,102 @@ function extractPartialClassification(obj: unknown): PromptClassification | unde
...
+	recovered: string[];
+	/** Per-field error reasons for fields that failed validation */
+	errors: string[];
+}
+
+function extractBestEffortFields(obj: unknown): BestEffortResult {
+	const fields: BestEffortClassificationFields = {
+		intent: '',
+		domain: '',
+		scope: '',
```
digitarald, line 124: This function duplicates most of the validation logic from `extractPartialClassification` above (lines 88-117): same type checks, same `isValidIntent`/`isValidDomain`/`isValidScope` guards, same time estimate extraction. The two will drift apart as new fields are added.

Consider refactoring so `extractPartialClassification` is implemented in terms of `extractBestEffortFields`:

```ts
function extractPartialClassification(obj: unknown): PromptClassification | undefined {
	const { fields, errors } = extractBestEffortFields(obj);
	// Core fields must all be valid (no errors on required fields)
	if (!fields.intent || !fields.domain || !fields.scope || fields.confidence === -1 || !fields.reasoning) {
		return undefined;
	}
	return {
		intent: fields.intent,
		domain: fields.domain,
		scope: fields.scope,
		confidence: fields.confidence,
		reasoning: fields.reasoning,
		timeEstimate: { bestCase: fields.timeEstimateBestCase, realistic: fields.timeEstimateRealistic },
	};
}
```

This keeps one source of truth for validation rules.

src/extension/prompt/node/promptCategorizer.ts
```diff
@@ -117,6 +117,102 @@ function extractPartialClassification(obj: unknown): PromptClassification | unde
...
+		confidence: -1,
+		reasoning: '',
+		timeEstimateBestCase: '',
+		timeEstimateRealistic: '',
+	};
```
digitarald, line 129: Using `-1` as a sentinel for an `isMeasurement: true` field is a bit unusual — some telemetry backends/dashboards may filter negative measurements or skew aggregations (averages, percentiles). Worth confirming that the downstream Kusto queries / dashboards handle `confidence == -1` gracefully and exclude it from mean/percentile calculations.

Alternatively, you could keep the sentinel in the typed fields but emit `NaN` or omit the field entirely in the telemetry payload when confidence isn't determined.

src/extension/prompt/node/promptCategorizer.ts
```diff
@@ -117,6 +117,102 @@ function extractPartialClassification(obj: unknown): PromptClassification | unde
...
+		recovered.push('intent');
+	} else if (typeof c.intent !== 'string') {
+		errors.push(`intent: expected string, got ${typeof c.intent}`);
+	} else {
+		errors.push(`intent: invalid value "${c.intent}"`);
+	}
+
+	if (typeof c.domain === 'string' && isValidDomain(c.domain)) {
+		fields.domain = c.domain;
+		recovered.push('domain');
```
digitarald, line 148: The error messages here embed the raw LLM output value (e.g. `invalid value "${c.intent}"`). Since these end up in `errorDetail` which is sent to internal telemetry, this is likely fine — but be aware that the LLM could return unexpected content here. The `errorDetail` is already truncated to 500 chars, so the risk is bounded. Just flagging for awareness.

src/extension/prompt/node/promptCategorizer.ts
```diff
@@ -347,19 +446,19 @@ export class PromptCategorizerService implements IPromptCategorizerService {
...
+				reasoning: classification?.reasoning ?? bestEffortFields?.reasoning ?? '',
 				prompt: truncatedPrompt,
 			},
 			{
 				promptLength: request.prompt.length,
 				numReferences: request.references?.length ?? 0,
 				numToolReferences: request.toolReferences?.length ?? 0,
-				confidence: classification?.confidence ?? 0,
+				confidence: classification?.confidence ?? bestEffortFields?.confidence ?? -1,
 				latencyMs,
```
digitarald, line 453: nit: This debug log still only prints `classification?.intent` etc., so it won't reflect the best-effort fields on `invalidClassification` outcomes. Consider also logging the best-effort fields when they're populated:

```ts
this.logService.debug(`[PromptCategorizer] Classification complete: outcome=${outcome || 'success'}, latencyMs=${latencyMs}, intent=${classification?.intent ?? bestEffortFields?.intent}, domain=${classification?.domain ?? bestEffortFields?.domain}, scope=${classification?.scope ?? bestEffortFields?.scope}`);
```

src/extension/prompt/node/promptCategorizer.ts
kevin-m-kent: Thank you, this should be addressed now.

src/extension/prompt/node/promptCategorizer.ts
kevin-m-kent: good point. I updated it to exclude the field from the measures bag when it is missing instead of -1. I think this improves things and makes the KQL side less prone to problems

src/extension/prompt/node/promptCategorizer.ts
kevin-m-kent: Thanks, I think I am ok with this since it's 1p telemetry.

PR #4308, opened 2026-03-09
Skip repo info diff for VFS and sparse checkout enlistments

tyrielv:
VFS (core.virtualfilesystem) and sparse checkout (core.sparsecheckout) repos are unlikely to have all blobs available locally, making diff operations expensive or impossible. Return early with a 'virtualFileSystem' result when either git config key is set.


tyrielv:
This is one proposal for fixing the performance issues that VFS users in large repositories are seeing with this extension. I have another at #4309.

PR #4309, opened 2026-03-09
Skip repo info diff when commit count exceeds limit

tyrielv:
Add MAX_DIFF_COMMITS (30) check using repository.log() with a range of mergeBase..HEAD. If the number of commits returned hits the limit, return early with a 'tooManyCommits' result to avoid expensive diff operations on branches with large commit ranges.

Add optional commitCount measurement to telemetry, populated only when the commit count is within the limit.


tyrielv:
This is one proposal for fixing the performance issues that VFS users in large repositories are seeing with this extension. I have another at https://github.com/microsoft/vscode-copilot-chat/pull/4308.

PR #4339, opened 2026-03-10
More repo info telemetry check to support windows repo perf issues

IanMatthewHuff:
More follow up to: https://github.com/microsoft/vscode-copilot-chat/pull/3774

The configuration of certain windows repos is still having perf issues after the change above. With help from a windows dev the most probably issues were seen as:

1. The configuration of virtualfilesystem and sparsecheckout causing issues with calculating diffs:
https://github.com/microsoft/vscode-copilot-chat/pull/4308

2. Enough extensive local commits created (especially with renames) that just calculating the number of diffs in the change slows down vscode with expensive git processes
https://github.com/microsoft/vscode-copilot-chat/pull/4309

For this date our usual thing is to prioritize avoiding nasty perf issues at the expense of possibly dropping some data, so I'm implementing both changes here. I also do believe that in most cases of the commit count check we'd be pushing on too big a diff to log in telemetry in any case possibly. As with the other change we are logging why we don't check telemetry so if we start to not even enough events through we can re-examine.

Unit tests also updated.

For my local testing I did the following tests with a local build:

- [x] Events fired for a standard scenario
- [x] Too old event fires for >30day old diff
- [x] Does not collect and logs 'virtualfilesystem' with core.virtualfilesystem or core.sparsecheckout specified
image

- [x] Does collect after those settings are turned off in the repo
- [x] Collects after 25 local commits from upstream are added
- [x] Does not collect after 10 more local commits are added and logs 'tooManyCommit'
image

### DIFF
--- a/src/extension/prompt/node/repoInfoTelemetry.ts
+++ b/src/extension/prompt/node/repoInfoTelemetry.ts
@@ -231,11 +231,14 @@ export class RepoInfoTelemetry {
 
 		// VFS and sparse checkout enlistments are unlikely to have all blobs available locally,
 		// making diff operations expensive or impossible. Skip early if either is configured.
+		// core.virtualfilesystem is a path to a hook script, any non-empty value means VFS is active.
+		// core.sparsecheckout is a git boolean: true/yes/on/1 are truthy per git-config spec.
 		// If we can't determine the config, skip to be safe.
 		try {
 			const virtualFileSystem = await repository.getConfig('core.virtualfilesystem');
 			const sparseCheckout = await repository.getConfig('core.sparsecheckout');
-			if (virtualFileSystem === 'true' || sparseCheckout === 'true') {
+			const GIT_TRUE_VALUES = new Set(['true', 'yes', 'on', '1']);
+			if (virtualFileSystem || GIT_TRUE_VALUES.has(sparseCheckout.toLowerCase())) {
 				return skipDiffResult('virtualFileSystem');
 			}
 		} catch {
--- a/src/extension/prompt/node/test/repoInfoTelemetry.spec.ts
+++ b/src/extension/prompt/node/test/repoInfoTelemetry.spec.ts
@@ -840,12 +840,12 @@ suite('RepoInfoTelemetry', () => {
 		mockGitServiceWithRepository();
 		mockGitExtensionWithUpstream('abc123');
 
-		// Override getConfig to return a value for core.virtualfilesystem
+		// Override getConfig to return a hook path for core.virtualfilesystem (any non-empty string means VFS is active)
 		const mockApi = gitExtensionService.getExtensionApi();
 		const mockRepo = mockApi!.getRepository(URI.file('/test/repo'))!;
 		vi.spyOn(mockRepo, 'getConfig').mockImplementation(async key => {
 			if (key === 'core.virtualfilesystem') {
-				return 'true';
+				return '/path/to/vfs-hook';
 			}
 			return '';
 		});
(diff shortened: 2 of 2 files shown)


src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -210,36 +213,61 @@ export class RepoInfoTelemetry {
...
+		// making diff operations expensive or impossible. Skip early if either is configured.
+		// If we can't determine the config, skip to be safe.
+		try {
+			const virtualFileSystem = await repository.getConfig('core.virtualfilesystem');
+			const sparseCheckout = await repository.getConfig('core.sparsecheckout');
+			if (virtualFileSystem || sparseCheckout) {
+				return skipDiffResult('virtualFileSystem');
+			}
+		} catch {
+			return skipDiffResult('virtualFileSystem');
```
IanMatthewHuff, line 245: For all these err on the side of not collecting if we can't calculate the check we are doing.

IanMatthewHuff:
@tyrielv for context from the windows side. Thanks for providing the diagnosis info.

@zhichli for context from vscode / repo info side and for review / merge help. Much appreciated!

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -210,36 +213,61 @@ export class RepoInfoTelemetry {
...
-		};
+		});
+
+		// VFS and sparse checkout enlistments are unlikely to have all blobs available locally,
+		// making diff operations expensive or impossible. Skip early if either is configured.
+		// If we can't determine the config, skip to be safe.
+		try {
+			const virtualFileSystem = await repository.getConfig('core.virtualfilesystem');
+			const sparseCheckout = await repository.getConfig('core.sparsecheckout');
+			if (virtualFileSystem || sparseCheckout) {
```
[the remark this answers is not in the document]
IanMatthewHuff: Updating...

src/extension/prompt/node/repoInfoTelemetry.ts
```diff
@@ -210,36 +213,61 @@ export class RepoInfoTelemetry {
...
-		};
+		});
+
+		// VFS and sparse checkout enlistments are unlikely to have all blobs available locally,
+		// making diff operations expensive or impossible. Skip early if either is configured.
+		// If we can't determine the config, skip to be safe.
+		try {
+			const virtualFileSystem = await repository.getConfig('core.virtualfilesystem');
+			const sparseCheckout = await repository.getConfig('core.sparsecheckout');
+			if (virtualFileSystem === 'true' || sparseCheckout === 'true') {
```
tyrielv: core.virtualFileSystem isn't necessarily a boolean. If it has any truthy (for C-definition of truthy) value then it is enabled.

tyrielv: "true" is used sometimes, but typically only in unit tests. Real repositories will have it set to the path of a hook so git can talk to the VFS.

tyrielv: core.sparsecheckout should also be truthy - I see examples of both "true" and "1" for it in the git source code.

IanMatthewHuff: So it looks like virtualFileSystem is just if it's specified, but sparse checkout is actually this https://git-scm.com/docs/git-config#Documentation/git-config.txt-boolean ? Looks like vscode's git extension doesn't do any type of type coercion there. So ['yes','true','on','1'] looks like the set to me, case insensitive. Was honestly a bit hard to tell from the docs exact what sparsecheckout was for sure, but I always saw enable and disable used to describe it, and since true and 1 mentioned went with the git boolean defintion.

sbatten (approved):

PR #4359, opened 2026-03-11
Add sticky_threshold and sticky_override to router decision API

aashna:
Support threshold-based sticky routing to prevent mid-conversation model switching on low-confidence predictions.

Changes:
- Add sticky_override bool to RouterDecisionResponse interface and validator
- Add optional sticky_threshold parameter to getRouterDecision()
- Send sticky_threshold in request body when provided
- Log sticky_override in trace output and request logger markdown
- AutomodeService logs sticky override events for observability

### DIFF
--- a/src/platform/endpoint/node/routerDecisionFetcher.ts
+++ b/src/platform/endpoint/node/routerDecisionFetcher.ts
@@ -6,7 +6,7 @@
 import { RequestType } from '@vscode/copilot-api';
 import { Codicon } from '../../../util/vs/base/common/codicons';
 import { IAuthenticationService } from '../../authentication/common/authentication';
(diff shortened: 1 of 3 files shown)


src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -34,7 +35,8 @@ const routerDecisionResponseValidator: IValidator<RouterDecisionResponse> = vObj
 	scores: vRequired(vObj({
 		needs_reasoning: vRequired(vNumber()),
 		no_reasoning: vRequired(vNumber())
-	}))
+	})),
+	sticky_override: vRequired(vBoolean())
```
lramos15: Is stick override always required? 

We don't want this validator to fail if it can be optional

PR #4366, opened 2026-03-11
Change default Anthropic thinking effort to medium

bhavyaus:


copilot-chat-0.40.0.vsix
bryanchen-d, line 1: why do we need to check in the vsix?

PR #4370, opened 2026-03-12
Fix duplicate semantic search results from git worktrees

dzamoshchin:
Fixes microsoft/vscode#295628

When a repo has git worktrees, the `semantic_search` tool returns the same file multiple times - once from the main repo and once from each worktree. This pollutes the context of the LLM, reducing the agent's ability to find relevant code and degrading accuracy.

This happens because the Git extension exposes each worktree as a separate repository, and `CodeSearchChunkSearch` creates a `CodeSearchRepo` for each one.

### Fix

Two filters in `CodeSearchChunkSearch.openGitRepo()`:

1. Skip repos that aren't related to the workspace (filters worktrees at external paths like `*.worktrees/` or `~/.cursor/worktrees/`)

2. Skip repos whose remote is already covered by an existing repo (handles worktrees physically inside the workspace like `.claude/worktrees/`)


dzamoshchin:

PR #4381, opened 2026-03-12
Add new rate limiter

itsibitzi:
This implements the multibucket client side rate limiter for blackbird. This is required because we made it so that different RPCs use different quotas (you need to be able to upload very quickly when uploading documents, but we dont want you to be able to request hundreds of new checkpoints per second).

It also brings the sliding timing and N window up to where the reference implementation is. This mainly just clears out a kinda pointless background task and adds a slow start.

This is not my code base so I may have messed up some stuff stylistically, feel free to push commits on top or let me know if there's anything you'd like different.

I did a test of uploading some things and was able to get around 100 document per second.


itsibitzi:

mjbvz:
Awesome! Thanks for handling this. I fixed the merge conflicts and opened a new PR here: #4395

PR #4384, opened 2026-03-12
Fix: NES debounce and language context fetch do not honor cancellation token

etvorun:
Fixes microsoft/vscode#308373

## Problem

When Visual Studio cancels a Next Edit Suggestions (NES) request (e.g. because the user moved the cursor or a new request superseded the old one), the cancellation was effectively a no-op for two critical phases of the request pipeline in \XtabProvider\:

1. **Debounce phase** — \debounce()\ used \wait timeout(debounceTime)\ unconditionally. Cancelling the request had no effect; the debounce timer ran to completion before the cancellation check fired.
2. **Language context fetch phase** — \getLanguageContext()\ used \wait raceTimeout(getContextPromise(), debounceTime)\ without passing the cancellation token. The fetch would block the full debounce window even when the outer request was already cancelled.

### Impact on request queuing

Because cancelled requests would not yield until the debounce (and optionally the language context timeout) fully elapsed, subsequent NES requests accumulated behind them. Each queued request inherited an artificially inflated wait time, causing a visible multi-hundred-millisecond delay before later requests could start their own debounce or LLM fetch. In scenarios where VS fires rapid NES requests (e.g. caret moves after Tab acceptance) this manifested as a stall: the cancelled request's debounce blocked the pipeline, making healthy follow-on requests appear delayed or entirely unresponsive.

## Fix

The debounce now resolves immediately when the cancellation token fires, instead of always waiting the full configured window.

## Behavior Change

| Scenario | Before | After |
|---|---|---|
| VS cancels NES request during debounce | Debounce runs to completion (~100–500ms depending on config) | Debounce exits immediately |
| VS cancels NES request during language context fetch | Fetch waits for full \debounceTime\ budget | Fetch exits immediately |
| Next request behind a cancelled one | Delayed by predecessor's residual wait | Starts immediately after cancellation |
| Simulation tests | Unaffected (debounce is skipped in simulation context) | Unchanged |
| Normal (non-cancelled) requests | Unchanged | Unchanged |

### New unit tests (\src/extension/xtab/test/node/xtabProvider.spec.ts\)

Two tests were added to the \XtabProvider integration > debounce behavior\ suite:

**\cancellation during debounce exits early with GotCancelled before LLM fetch\**
- Configures a 500 ms debounce.
- Starts \provideNextEdit\ and cancels the token after 20 ms.
- Asserts the result is \NoNextEditReason.GotCancelled\.
- Asserts total elapsed time is well under 500 ms (debounce was interrupted).
- Asserts \streamingFetcher.callCount === 0\ — the LLM fetch was never reached.

**\pre-cancelled token resolves without waiting for debounce\**
- Configures a 500 ms debounce.
- Cancels the token *before* calling \provideNextEdit\.
- Asserts the result is \NoNextEditReason.GotCancelled\.
- Asserts total elapsed time is well under 500 ms — no blocking wait occurred.

### Existing test coverage

- Prior tests already verify that \debounce()\ is skipped entirely in simulation mode (\isInSimulationTests = true\), so simulation tests remain unaffected.
- The cancellation token is already propagated throughout \doGetNextEditWithSelection\; the fix wires it into the two previously-exempt awaits in the same call chain.

### Manual validation

In VS, cancel rapid successive NES requests (cursor moves, new edits) and verify via ETW / LSP logs that follow-on \NES.GetNextEdit.Start\ events are no longer delayed by the predecessor's debounce window.


andreamah:
LGTM!

etvorun:
@chrmarti, @benibenj or someone else, can you please merge this PR. I do not have the rights to do that.

etvorun:
@ulugbekna, @benibenj or anyone else, this PR was removed from merge queue due to test error "Failed to download and unzip VS Code insiders-unreleased". Any recommendations on how to resolve it?

PR #4396, opened 2026-03-13
Support links from Terminal that wasn't opened via VS Code

DonJayamanne:


anthonykim1:
See direct message, and TPI https://github.com/microsoft/vscode/issues/301574

PR #4398, opened 2026-03-13
Disable debug log aggregation by default

pwang347:


vijayupadya (dismissed):

mjbvz:
@pwang347 Can you also make sure this change goes into main if it hasn't already

mjbvz (dismissed):

PR #4429, opened 2026-03-16
Search subagent -- resolve relative and absolute paths

24anisha:
Currently, the search subagent expects the model to return absolute paths to the relevant code snippets. With this change, it will try prepending the cwd to the file paths if the paths as-is don't point to any files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3597,7 +3597,7 @@
 					},
 					"github.copilot.chat.languageContext.typescript.enabled": {
 						"type": "boolean",
-						"default": false,
+						"default": true,
 						"scope": "resource",
 						"tags": [
 							"experimental",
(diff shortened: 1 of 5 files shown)


src/extension/tools/node/searchSubagentTool.ts
```diff
@@ -158,8 +159,12 @@ class SearchSubagentTool implements ICopilotTool<ISearchSubagentParams> {
 			const startLine = parseInt(startLineStr, 10);
 			const endLine = parseInt(endLineStr, 10);
 
+			const resolvedFilePath = cwd && !filePath.startsWith(cwd)
```
connor4312: this

- Doesn't account for case insensitivity
- Never allows the subagent to return a path outside the cwd

Logic like this should probably only be applied to non-absolute paths

24anisha: edited the implementation, which resolves this concern

PR #4486, opened 2026-03-18
Allow user to retry model rate limits automatically with auto

lramos15:


src/extension/conversation/vscode-node/chatParticipants.ts
```diff
@@ -261,6 +282,20 @@ Learn more about [GitHub Copilot](https://docs.github.com/copilot/using-github-c
 		stream.warning(messageString);
 		return request;
 	}
+
+	private async switchToAutoModel(request: vscode.ChatRequest, stream: vscode.ChatResponseStream, alwaysSwitchToAuto: boolean): Promise<ChatRequest> {
+		const autoModel = (await vscode.lm.selectChatModels({ id: 'auto', vendor: 'copilot' }))[0];
```
roblourens, line 287: I don't think we should use extension API in here, shouldn't we use the endpoint provider?

src/extension/conversation/vscode-node/chatParticipants.ts
```diff
@@ -261,6 +282,20 @@ Learn more about [GitHub Copilot](https://docs.github.com/copilot/using-github-c
...
+		if (!autoModel) {
+			return request;
+		}
+		await vscode.commands.executeCommand('workbench.action.chat.changeModel', { vendor: autoModel.vendor, id: autoModel.id, family: autoModel.family });
```
roblourens, line 291: Ideally this would go through the chat response stream, this command could change some other chat session instead

lramos15: Does the chat response stream have a way to switch models? This is how we do it for the quota exhaustion too

src/extension/conversation/vscode-node/chatParticipants.ts
lramos15: Yes, but the endpoint provider doesn't have a concept of `vendor` which the core command for switching model needs. We could hardcode `copilot` but that felt wrong too

src/extension/conversation/vscode-node/chatParticipants.ts
roblourens: Not currently. Apparently that command will flip it for _all_ chat sessions, that seems wrong

roblourens: Because some chat sessions may be using other models that didn't fail

PR #4492, opened 2026-03-18
When rendering an image in a prompt, also include its uri so the LLM can reference it later (e.g. as input to the artifacts tool).

hediet:


src/extension/prompts/node/panel/toolCalling.tsx
```diff
@@ -933,3 +944,9 @@ function sendNotebookEditToolValidationTelemetry(invokeOutcome: ToolInvocationOu
 		}
 	);
 }
+
+export function buildToolImageResourceUri(sessionId: string, coreToolCallId: string, imageIndex: number, ext: string): string {
```
roblourens, line 948: What is this?

hediet: This creates the uri that is handled [here](https://github.com/microsoft/vscode/blob/adacf1f44b46fcb2dee6a95bc4e57e179c20f51a/src/vs/workbench/contrib/chat/common/model/chatModel.ts#L2794-L2796) in core.

PR #4499, opened 2026-03-18
Auto mode: re-route at summarization boundaries + per-conversation routing

aashna:
## Summary

Adds two routing strategies to the binary model router, controlled by experiment flags, plus comprehensive cost/caching telemetry for measuring the impact of routing decisions on token economics.

### 1. Re-route at Summarization Boundaries
When conversation history is compacted (foreground or background), the router re-evaluates model selection using the summarized text. The current model performs summarization (leveraging its warm KV cache), then the router picks the optimal model for the next phase.

**Flow:** Model A summarizes -> router classifies summary -> picks Model B -> Model B starts fresh with small context (~1K tokens)

**Why:** Summarization is a natural cache boundary — the old KV cache is discarded anyway. Re-routing here is effectively free and enables downgrading to cheaper models when task complexity decreases.

**Hooks:** Foreground (BudgetExceeded), background (budget-exceeded path), post-render background. Manual /compact excluded.

### 2. Per-Conversation Routing Mode
Opt-in via \UsePerConversationRouting\ experiment flag. Routes once on the first turn and reuses that model for all subsequent turns, eliminating mid-conversation model switching.

Both per-turn (default) and per-conversation paths coexist via if/else.

### \automode.routerBeforeSummarization\
Fires when the router re-evaluates at a compaction boundary.
- **Strings:** previousModel, nextModel, predictedLabel, modelChanged, conversationId, routingMode
- **Measurements:** confidence, routerLatencyMs, turnNumber, previousModelDiscountedCost, nextModelDiscountedCost, preCompactionPromptTokens

### \automode.routerModelSelected\ (NEW)
Fires on every per-turn model selection for cost tracking.
- **Strings:** model, previousModel, routingMode, modelChanged
- **Measurements:** turnNumber, discountedCost

### \automode.routerBeforeSummarizationError\
Fires when the pre-summarization router call fails.

## Files Changed

| File | Change |
|------|--------|
| \automodeService.ts\ | +resolveEndpointForSummarization(), per-conversation routing, _isPerConversationRouting(), enhanced telemetry | | \gentIntent.ts\ | Import IAutomodeService, inject via DI, 3 summarization hooks | | \agentPrompt.tsx\ | Thread automodeService through prompt props | | \summarizedConversationHistory.tsx\ | Accept/forward automodeService, call router before summarization | | \configurationService.ts\ | Add UsePerConversationRouting config key | | \automodeService.spec.ts\ | 2 per-conversation routing tests |

## Cost Analysis Design
No USD computation on client. Raw inputs logged (discountedCost multipliers, token counts, model names) for server-side cost analysis via KQL joins with copilot_v0_quota_event billing data.


src/extension/intents/node/agentIntent.ts
```diff
@@ -474,13 +476,18 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
 			}
 		}
 
-		// Helper function for synchronous summarization flow with fallbacks
+		// --- Post-summarization routing: after Model A produces the summary,
+		// the router picks Model B for subsequent turns. ---
+		const resolveNextEndpoint = (summaryText: string) => this._resolveNextEndpointAfterSummarization(promptContext, summaryText);
```
bhavyaus: Is there a way to check if we're in automode first and only then try to resolve the endpoint?

src/extension/intents/node/agentIntent.ts
```diff
@@ -710,6 +717,32 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
...
+			const result = await this.automodeService.resolveEndpointForSummarization(
+				conversationId, summaryText, allEndpoints);
+			if (result) {
+				this.logService.debug(
+					`[Agent] Post-summarization routing: ${result.previousModel} -> ${result.endpoint.model} `
+					+ `(label=${result.routerDecision.predicted_label}, `
+					+ `confidence=${(result.routerDecision.confidence * 100).toFixed(1)}%)`);
+			}
+		} catch (e) {
+			this.logService.warn('[Agent] Post-summarization routing failed', (e as Error).message);
```
bhavyaus: do we need to capture this telemetry?

aashna: id recommend capturing this one unless theres a strong reason to not

bhavyaus: yeah sorry. I meant you should send a telemetry event for this with the telmetryService.

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
```diff
@@ -409,6 +415,14 @@ export interface SummarizedAgentHistoryProps extends BasePromptElementProps, Age
 	readonly summarizationInstructions?: string;
 	/** Whether this summarization was triggered as a background or foreground operation. Defaults to 'foreground'. */
 	readonly summarizationSource?: 'background' | 'foreground';
+	/**
+	 * Optional callback invoked after Model A produces the summary.
```
bhavyaus, line 418: Do we need the full summaryText for this routing? Or could we use the original user request itself?

src/extension/intents/node/agentIntent.ts
```diff
@@ -474,13 +476,18 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
...
+		const resolveNextEndpoint = (summaryText: string) => this._resolveNextEndpointAfterSummarization(promptContext, summaryText);
+
 		const renderWithSummarization = async (reason: string, renderProps: AgentPromptProps = props): Promise<RenderPromptResult> => {
 			this.logService.debug(`[Agent] ${reason}, triggering summarization`);
 			try {
 				const renderer = PromptRenderer.create(this.instantiationService, endpoint, this.prompt, {
 					...renderProps,
 					triggerSummarize: true,
+					summarizationSource: 'foreground',
+					resolveNextEndpoint,
```
bhavyaus, line 495: To confirm, we are only enabling this for foreground compaction with this PR correct?

src/extension/prompts/node/agent/summarizedConversationHistory.tsx
aashna: the summary is intentional because it represents the next phase's complexity after compaction, and the router needs to classify what comes next, not what the user just said

bhavyaus:
Closing. Reference https://github.com/microsoft/vscode-copilot-chat/pull/4687 instead

PR #4515, opened 2026-03-19
Refactor thinking and effort control: per-request opt-in

bhavyaus:
## Summary

Refactors how thinking and reasoning effort are controlled in chat requests. Instead of checking ChatLocation to decide whether to enable thinking, callers now explicitly opt in via enableThinking and reasoningEffort on IMakeChatRequestOptions.

### New request options
- enableThinking - opt-in flag, defaults to off
- reasoningEffort - effort level from model picker

### Removed
- disableThinking (replaced by enableThinking)
- AnthropicThinkingEffort config setting
- ResponsesApiReasoningEffort config setting
- Location-based thinking gates

### Model picker integration
- Parse reasoning_effort from CAPI model list
- Build configurationSchema with effort dropdown
- No default effort, user must choose

### Who enables thinking
- Agent mode: enableThinking true
- ResponsesProxy: enableThinking true
- MessagesProxy: enableThinking true
- Everything else: thinking off by default

### DIFF
--- a/src/extension/conversation/vscode-node/languageModelAccess.ts
+++ b/src/extension/conversation/vscode-node/languageModelAccess.ts
@@ -54,10 +54,13 @@ function buildConfigurationSchema(endpoint: IChatEndpoint): { configurationSchem
 		return {};
 	}
 
-	// Compute picker default: Anthropic and Gemini default to 'high', OpenAI to 'medium'
+	// Only enable effort picker for Claude and GPT models
 	const family = endpoint.family.toLowerCase();
-	const isHighDefault = family.startsWith('claude') || family.startsWith('gemini') || family.startsWith('strudel');
-	const preferred = isHighDefault ? 'high' : 'medium';
+	if (!family.startsWith('claude') && !family.startsWith('gpt-')) {
+		return {};
+	}
+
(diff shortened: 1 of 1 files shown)


src/extension/conversation/vscode-node/languageModelAccess.ts
```diff
@@ -44,6 +44,41 @@ import { PromptRenderer } from '../../prompts/node/base/promptRenderer';
...
+					enumDescriptions: effortLevels.map(level => {
+						switch (level) {
+							case 'none': return vscode.l10n.t('No reasoning applied');
+							case 'low': return vscode.l10n.t('Faster responses with less reasoning');
+							case 'medium': return vscode.l10n.t('Balanced reasoning and speed');
+							case 'high': return vscode.l10n.t('Maximum reasoning depth');
+							default: return level;
+						}
+					}),
+					default: undefined,
```
sandy081: There should be a default value right otherwise users will not know what is the default thinking effort.

hediet (approved):

illusive5943:
This caused a feature regression, we had "xhigh" thinking for OpenAI GPT models recently added in Settings, but now it's gone and we only have "high" as selectable in chat ui.

PR #4518, opened 2026-03-19
Enable linter tags by default for cursor prediction

benibenj:
cc @ulugbekna


bpasero (approved):

src/extension/xtab/node/xtabNextCursorPredictor.ts
```diff
@@ -235,49 +241,11 @@ export class XtabNextCursorPredictor {
 		}
 
 		const expLintOptions = this.configService.getExperimentBasedConfig(ConfigKey.TeamInternal.InlineEditsNextCursorPredictionLintOptionsString, this.expService);
-		if (!expLintOptions) {
-			return undefined;
+		if (expLintOptions) {
+			return parseLintOptionString(expLintOptions);
```
ulugbekna, line 245: @benibenj a heads up that the parsing function throws if the lint options is not valid json and would break the feature

PR #4538, opened 2026-03-19
[WIP] Remove hard-coded completions model

andreamah:
This PR removes the explicit reference to a fallback model for completions. It seems that a considerable number of users face a race condition where they end up using the fallback during initialization, and having something hard-coded like this could be dangerous. 

Not having the fallback does cause a lot of async downstream churn, so I'm also open to other solutions. This was also drafted up mostly by copilot, so it might not be the final solution, but it gives us an idea of what a non-fallback solution could look like.

cc @chrmarti

### DIFF
--- a/src/extension/completions-core/vscode-node/completionsServiceBridges.ts
+++ b/src/extension/completions-core/vscode-node/completionsServiceBridges.ts
@@ -232,7 +232,7 @@ function registerStatusBar(accessor: ServicesAccessor): IDisposable {
 					new Promise(resolve => setTimeout(resolve, 100)),
 				]);
 			}
-			instantiationService.createInstance(CopilotStatusBarPickMenu).showStatusMenu();
+			await instantiationService.createInstance(CopilotStatusBarPickMenu).showStatusMenu();
 		});
 	}
 	disposables.add(registerStatusMenu(CMDToggleStatusMenuClient));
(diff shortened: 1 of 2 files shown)


src/extension/completions-core/vscode-node/lib/src/openai/model.ts
```diff
@@ -48,32 +48,31 @@ export class AvailableModelsManager extends Disposable implements ICompletionsMo
 		super();
 
 		if (shouldFetch) {
-			this._register(onCopilotToken(authenticationService, () => this.refreshAvailableModels()));
```
andreamah, line 51: Previously, the code relied on `onCopilotToken` to be called in order to init the available models. OnCopilotToken should be called when auth details are init https://github.com/andreamah/vscode-copilot-chat/blob/c920622fd973fd20e8e1a5e940846a2e13a8d8f5/src/extension/completions-core/vscode-node/lib/src/auth/copilotTokenNotifier.ts#L9-L17, however, there seems to be some ghost text generated before all of this happened. I was querying telemetry and it seemed like most people who were getting a fallback model eventually resolved to the non-fallback model after a few reqs, which leads me to believe that there is a pocket of time during init that we are somehow not initializing this yet.

cc @benibenj

PR #4554, opened 2026-03-20
feat: show 'Copied!' status bar feedback on copy action

ajithraghavan:
## Summary                            
  - Adds a brief 'Copied!' status bar message (2 seconds) when clicking copy in chat response action bar
  - Uses vscode.l10n.t for localization support                                                                   
                                                                                                                  
  ## Test plan                                                                                                    
  - [ ] Launch Extension Development Host, trigger a chat response                                                
  - [ ] Click the copy icon, confirm 'Copied!' appears in the status bar for ~2 seconds


connor4312:
We don't do this for data anywhere else in the editor, so I don't think we'd do it for chat, sorry.

ajithraghavan:
Thank you for the feedback, Connor! 

It would be nice if some visual feedback was shown when the copy action is performed

PR #4561, opened 2026-03-20
Work on new fetch module

lramos15:
Features
Callsite kill-switch — Every request requires a [callSite](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) identifier. Individual callsites can be disabled remotely via experiment treatment (copilot-fetch-disabled-callsites), allowing instant traffic shedding without a release.

Circuit breaker — Per-callsite circuit breaker (closed → open → half-open) that trips after N consecutive failures (default 5) and rejects requests until the endpoint recovers. Half-open probes are serialized, and cancelled probes release the slot to avoid deadlock.

Concurrency limiting — Per-callsite cap on in-flight requests with an optional queue timeout. Waiters are granted slots FIFO and respect AbortSignal cancellation.

Retry with exponential backoff — Configurable retries for 5xx errors (with jittered exponential backoff capped at 30 s) and 429 rate-limit responses (respecting Retry-After headers, capped at 60 s). Transient network errors (ECONNRESET, ETIMEDOUT, etc.) are also retried; non-transient errors (cert failures, invalid URLs) fail immediately.

Response caching — TTL-based in-memory cache (LRU eviction at configurable max entries) with optional persistent storage via a Memento-compatible backend. Cache keys incorporate method, URL, callsite, and a fingerprint of auth/accept headers to prevent cache poisoning. Supports:

Conditional requests — Stored ETags/Last-Modified values are sent as If-None-Match/If-Modified-Since; 304 responses refresh the TTL without re-transferring the body.
Stale-while-revalidate — Expired entries can be served for a configurable grace window while a background re-fetch updates the cache.
Persistence byte budget — Configurable [maxPersistBytes](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) limits disk usage; newest entries are kept when the budget is exceeded.
Request deduplication — Concurrent identical GET requests share a single in-flight promise. All callers receive an independent, re-readable [CachedFetchResponse](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) so consumed streams aren't an issue.

GitHub API quota-aware throttling — Automatic PID-controller–inspired throttling for github.com/ghe.com URLs. Learns endpoint → quota-bucket mappings from response headers ([x-github-quota-bucket-name](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), [x-github-total-quota-used](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)) and dynamically adjusts inter-request delay to stay near a configurable target (default 80%).

Polling utility ([PollingFetcher](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)) — Background poller that periodically invokes a fetch function and exposes the latest result as an observable value. Supports:

Window-state awareness (skip polls when inactive, resume on focus)
Skip-when-unused (saves cycles when nobody is consuming the value)
Dynamic interval computation (e.g. refresh 5 min before token expiry)
Initial cached value for synchronous first access
[createPollingFetch()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) convenience that routes polls through the full protection pipeline
AbortSignal support — All waits (retries, concurrency queue, throttle delays, sleep) are cancellable via a minimal [AbortSignalLike](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) interface compatible with DOM, Node.js, and VS Code signals.

Request defaults — Module-level defaults for retries, cache TTL, and timeout that apply as fallbacks; per-request values always win.

[fetchJson()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) convenience — Fetches a URL through the full pipeline and parses the response as JSON, throwing on non-OK status.


src/platform/fetch/common/newFetchService.ts
```diff
@@ -0,0 +1,178 @@
...
+	/**
+	 * Optional callback to compute a dynamic interval from the latest value.
+	 * Return `undefined` to fall back to {@link intervalMs}.
+	 */
+	readonly getNextIntervalMs?: (value: T) => number | undefined;
+	/**
+	 * Window state provider for skipping polls while the window is inactive.
+	 * Compatible with {@link IEnvService}.
+	 */
+	readonly windowStateProvider?: {
```
mjbvz, line 40: Can this be passed in when creating the service? I feel like each caller would need to duplicate this and it may be easy to forget

src/platform/fetch/common/newFetchService.ts
```diff
@@ -0,0 +1,178 @@
...
+	 * function and exposes the latest result as an observable value.
+	 *
+	 * The returned {@link PollingFetcher} is disposable and must be cleaned up
+	 * by the caller (e.g. via `_register`).
+	 *
+	 * Callsite kill-switching is handled by the fetch service layer that the
+	 * {@link fetchFn} should be calling into, so there is no separate callsite
+	 * parameter here.
+	 */
+	createPollingFetcher<T>(fetchFn: () => Promise<T>, options: PollingFetcherOptions<T>): IPollingFetcher<T>;
```
mjbvz, line 99: Do we need this generic version right now? I'd try to keep this focused on ones that make requests

src/platform/fetch/common/newFetchService.ts
```diff
@@ -0,0 +1,178 @@
...
+	/**
+	 * Performs a fetch request, subject to experiment-based callsite kill-switching,
+	 * circuit breaking, concurrency limiting, retry, and caching.
+	 *
+	 * When caching is enabled via `cacheTtlMs` on a GET request, the returned
+	 * response may be a `CachedFetchResponse` instead of a full platform
+	 * `Response`. To check, use the `isCachedFetchResponse()` helper exported
+	 * from `vscode-fetch/common/responseCache`, or structurally check for the
+	 * absence of platform-specific properties.
+	 */
```
mjbvz, line 68: What happens of I call this but the call site has been disabled?

src/platform/fetch/common/newFetchService.ts
```diff
@@ -0,0 +1,178 @@
...
+	fetch(url: string, options: FetchOptions): Promise<Response | CachedFetchResponse>;
```
mjbvz, line 69: The note about caching makes me feel that maybe `fetch` and `fetchWithCaching` should be two different methods

PR #4594, opened 2026-03-22
refactor: improve HTTP/2 GOAWAY and network error retry resilience

deepak1556:
- Retry graceful HTTP/2 GOAWAY (code 0) transparently in node-fetcher before surfacing as an error.

- Retry transient electron errors (ERR_NETWORK_CHANGED, etc.) on the same fetcher with a fresh connection in fetchWithFallbacks before falling back to a different fetcher.

- Fix canRetryOnceNetworkError to also check error .message for electron net:: errors, which were previously silently missed on the networkRequest path.

### DIFF
--- a/src/platform/networking/node/fetcherFallback.ts
+++ b/src/platform/networking/node/fetcherFallback.ts
@@ -95,36 +95,16 @@ export async function fetchWithFallbacks(availableFetchers: readonly IFetcher[],
 	try {
 		return { response: await fetcher.fetch(url, options) };
 	} catch (err) {
+		// Transient electron errors: retry once on the same fetcher with a fresh connection.
+		if (isRetryableElectronError((err as Error)?.message)) {
+			return { response: await retryAfterDisconnect(fetcher, url, options, err, (err as Error).message, logService, telemetryService) };
+		}
+
 		// For net::ERR_FAILED from network process crash, disconnect and retry once.
 		if (fetcher.isNetworkProcessCrashedError(err)) {
-			const fetcherId = fetcher.getUserAgentLibrary();
-			logService.info(`FetcherService: ${fetcherId} hit network process crash error (${(err as Error)?.message}), retrying after disconnect...`);
 			try {
-				await fetcher.disconnectAll();
-				const response = await fetcher.fetch(url, options);
-				logService.info(`FetcherService: ${fetcherId} retry after crash succeeded.`);
-				/* __GDPR__
-					"fetcherCrashRetry" : {
-						"owner": "deepak1556",
-						"comment": "Sent when a fetcher retries after a network process crash error",
-						"fetcher": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The fetcher that crashed" },
-						"outcome": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "Whether the retry recovered or failed" },
-						"error": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The error message" }
-					}
-				*/
-				telemetryService?.sendTelemetryEvent('fetcherCrashRetry', { github: true, microsoft: true }, {
-					fetcher: fetcherId,
-					outcome: 'recovered',
-					error: collectSingleLineErrorMessage(err, true),
-				});
-				return { response };
+				return { response: await retryAfterDisconnect(fetcher, url, options, err, 'network-process-crash', logService, telemetryService) };
 			} catch (retryErr) {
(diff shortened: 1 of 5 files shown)


src/platform/networking/node/fetcherFallback.ts
```diff
@@ -95,36 +95,18 @@ export async function fetchWithFallbacks(availableFetchers: readonly IFetcher[],
 	try {
 		return { response: await fetcher.fetch(url, options) };
 	} catch (err) {
+		// Transient electron errors retry once on the
+		// same fetcher with a fresh connection before letting the error propagate to the caller
+		// (which may fall back to a different fetcher).
+		if (retryableElectronErrorMessages.has((err as Error)?.message)) {
+			return { response: await retryAfterDisconnect(fetcher, url, options, err, (err as Error).message, logService, telemetryService) };
+		}
```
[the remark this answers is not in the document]
deepak1556: the tryfetch path exercised by token acquisition should not have the transient retry logic since we use that logic to demote to fallback fetchers. I am only adding the retry to direct fetch path to recover individual chat requests on transient failures.

src/platform/networking/node/fetcherFallback.ts
```diff
@@ -153,6 +133,59 @@ export async function fetchWithFallbacks(availableFetchers: readonly IFetcher[],
...
+			"fetcherTransientRetry" : {
+				"owner": "AurSol",
+				"comment": "Sent when a fetcher retries after a transient error",
+				"fetcher": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The fetcher that retried" },
+				"reason": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The type of error that triggered the retry" },
+				"outcome": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "Whether the retry recovered or failed" },
+				"error": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The error message" }
+			}
+		*/
+		telemetryService?.sendTelemetryEvent('fetcherTransientRetry', { github: true, microsoft: true }, {
```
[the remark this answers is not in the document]
deepak1556: the previous event was only available for a couple of days and no other queries rely it, its fine to replace with this generic event.

src/platform/networking/node/nodeFetchFetcher.ts
```diff
@@ -35,15 +54,37 @@ export class NodeFetchFetcher extends BaseFetchFetcher {
 	}
 }
 
+/**
+ * Detects a graceful HTTP/2 GOAWAY error (error code 0 = NO_ERROR).
+ * These occur during server-side connection draining
+ * and are safe to retry on a new connection.
+ */
+export function isGracefulGoawayError(e: any): boolean {
```
chrmarti, line 62: It seems GOAWAY is on the session level and comes with a Last-Stream-ID that indicates which requests might still have been processed. Only those with later ids can be retried safely if I understand correctly. Undici doesn't seem to handle or surface Last-Stream-ID currently.

src/platform/networking/node/nodeFetchFetcher.ts
```diff
@@ -26,6 +25,26 @@ export class NodeFetchFetcher extends BaseFetchFetcher {
 		return NodeFetchFetcher.ID;
 	}
 
+	override async fetch(url: string, options: FetchOptions): Promise<Response> {
+		try {
+			return await super.fetch(url, options);
+		} catch (e) {
+			if (isGracefulGoawayError(e)) {
+				await this.disconnectAll();
```
chrmarti, line 33: Do we need to disconnect all? https://github.com/nodejs/undici/issues/3140 seems to suggest that this is done automatically. Does still doing this here risk cancelling newly issued requests? (E.g. when several requests of a session get the same GOAWAY, could they step on each other here?)

src/platform/networking/node/fetcherFallback.ts
```diff
@@ -153,6 +133,66 @@ export async function fetchWithFallbacks(availableFetchers: readonly IFetcher[],
...
+	options: FetchOptions,
+	originalErr: unknown,
+	reason: string,
+	logService: ILogService,
+	telemetryService: ITelemetryService | undefined,
+): Promise<Response> {
+	const fetcherId = fetcher.getUserAgentLibrary();
+	logService.info(`FetcherService: ${fetcherId} hit ${reason}, retrying after disconnect...`);
+	try {
+		await fetcher.disconnectAll();
```
chrmarti, line 164: We disconnectAll for all 3 error types we retry here? Could this make requests step on each other as some are being retried while later ones come here to first disconnect all?

PR #4608, opened 2026-03-23
Set up experiment for `prompt_cache_key` in Responses API

Giuspepe:
For https://github.com/microsoft/vscode-internalbacklog/issues/7004

Since in the Codex CLI the `prompt_cache_key`  is set to the conversation ID ([see here](https://github.com/openai/codex/blob/d807d44ae7fb69e8e05fc6e6fddea65f7e9421f5/codex-rs/core/src/client.rs#L727)), let's try doing the same.
I also appended the model name to the `prompt_cache_key` for the case that the user changes the model inside a conversation.

From [OpenAI prompt caching docs](https://developers.openai.com/api/docs/guides/prompt-caching):
```
Cache Routing:
- Requests are routed to a machine based on a hash of the initial prefix of the prompt. The hash typically uses the first 256 tokens, though the exact length varies depending on the model.
- If you provide the prompt_cache_key parameter, it is combined with the prefix hash, allowing you to influence routing and improve cache hit rates. This is especially beneficial when many requests share long, common prefixes.
- If requests for the same prefix and prompt_cache_key combination exceed a certain rate (approximately 15 requests per minute), some may overflow and get routed to additional machines, reducing cache effectiveness.

...

Best practices

- Use the prompt_cache_key parameter consistently across requests that share common prefixes. Select a granularity that keeps each unique prefix-prompt_cache_key combination below 15 requests per minute to avoid cache overflow.
```

From [OpenAI API](https://developers.openai.com/api/reference/resources/responses/methods/create#responses-create-prompt_cache_key):
```
prompt_cache_key: optional string 
Used by OpenAI to cache responses for similar requests to optimize your cache hit rates. Replaces the user field.
```

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -345,7 +345,7 @@
 	"github.copilot.config.useResponsesApi": "Use the Responses API instead of the Chat Completions API when supported. Enables reasoning and reasoning summaries.\n\n**Note**: This is an experimental feature that is not yet activated for all users.\n\n**Important**: URL API path resolution for custom OpenAI-compatible and Azure models is independent of this setting and fully determined by `url` property of `#github.copilot.chat.customOAIModels#` or `#github.copilot.chat.azureModels#` respectively.",
 	"github.copilot.config.responsesApiReasoningSummary": "Sets the reasoning summary style used for the Responses API. Requires `#github.copilot.chat.useResponsesApi#`.",
 	"github.copilot.config.responsesApiContextManagement.enabled": "Enables context management for the Responses API. Requires `#github.copilot.chat.useResponsesApi#`.",
-	"github.copilot.chat.responsesApi.promptCacheKey.enabled": "Enables prompt cache key being set for the Responses API.",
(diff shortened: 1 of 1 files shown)


bhavyaus:
One question about expected behavior:
https://developers.openai.com/api/reference/resources/responses/methods/create#responses-create-prompt_cache_key

image


What does replacing the `user` field mean?

Would be good to get a vscbench eval run once before checking this in.

bhavyaus (approved):

Giuspepe:
> What does replacing the `user` field mean?

The `user` field was deprecated in favor of the `prompt_cache_key` field:
```
Deprecated user: optional string
This field is being replaced by safety_identifier and prompt_cache_key.   
Use prompt_cache_key instead to maintain caching optimizations. 
A stable identifier for your end-users. 
Used to boost cache hit rates by better bucketing similar requests and to help OpenAI detect and prevent abuse.
```

PR #4623, opened 2026-03-23
Internal variable resolution service for skills

pwang347:
Remove hardcoded references to /troubleshoot. Also, we should not use `getCurrentCapturingToken` as it can have timing issues. Fixes https://github.com/microsoft/vscode/issues/302219

CC @aeschli

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -16,7 +16,7 @@ permissions:
 jobs:
   check-test-cache:
     name: Check test cache
-    runs-on: [ self-hosted, 1ES.Pool=1es-vscode-ubuntu-22.04-x64 ]
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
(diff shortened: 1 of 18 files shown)


alexdima (dismissed):

src/extension/prompts/node/panel/promptFile.tsx
```diff
@@ -32,7 +33,8 @@ export class PromptFile extends PromptElement<PromptFileProps, void> {
 		@IPromptPathRepresentationService private readonly promptPathRepresentationService: IPromptPathRepresentationService,
 		@IIgnoreService private readonly ignoreService: IIgnoreService,
 		@IWorkspaceService private readonly workspaceService: IWorkspaceService,
-		@IChatDebugFileLoggerService private readonly chatDebugFileLoggerService: IChatDebugFileLoggerService,
+		@ICustomInstructionsService private readonly customInstructionsService: ICustomInstructionsService,
+		@IBuildPromptContext private readonly promptContext: IBuildPromptContext | undefined,
```
aeschli: The IBuildPromptContext is not a service and should not be injected like one. Services don't have state, while the IBuildPromptContext has state.
The IBuildPromptContext should come in as part of the param.
I know that @connor4312 pioneered this with the IPromptEndpoint but given that we already pass throigh the IBuildPromptContext to most element, I think we should follow this way for now.

aeschli:
Other things to look at:
- make sure we propertly get the sessionId from a sessionResource. Looks like this is harneess dependend
- find variables in the full content, not just the first one
- give the variable name a unique name, to avoid unintended matches. e.g. `{{copilot:sessionId}}`

bpasero (approved):

PR #4672, opened 2026-03-24
fix: resolve broken ToolName interpolation in NoChangeError message

ShehabSherif0:
Fixes microsoft/vscode#308365

## Problem

In `editFileToolUtils.tsx` (line 671), the `NoChangeError` message uses single quotes for a string containing `${ToolName.ReadFile}`. Template interpolation only works inside backtick strings, so the model literally receives the text `${ToolName.ReadFile}` instead of `read_file`. Additionally, `ToolName` is not imported in the file, and the message contains a doubled word ("and and").

**What the model sees today:**
```
Failed to apply edit. Use the ${ToolName.ReadFile} tool to re-read the file and and determine the correct edit.
```

**What it should see:**
```
Failed to apply edit. Use the read_file tool to re-read the file and determine the correct edit.
```

## Root Cause

The string was written with single quotes (`'...'`) instead of backtick template literals (`` `...` ``). Every other usage of `ToolName.*` in the codebase correctly uses backticks. This is the only instance using single quotes.

## Fix

1. Add `import { ToolName } from '../common/toolNames';`
2. Change single quotes to backticks for proper `${ToolName.ReadFile}` interpolation
3. Fix `and and` -> `and`

## Impact

This error triggers when `replace_string_in_file` produces no actual change (the model's edit matched via fuzzy/whitespace/similarity matching but yielded identical content). The garbled tool name prevents the model from following the recovery instruction, degrading agent mode's self-correction ability.

## Testing

All 108 existing tests in `editFileToolUtils.spec.ts` pass.


ShehabSherif0:
@DonJayamanne Please review this. Thank you!

DonJayamanne (approved):

bpasero (approved):

PR #4673, opened 2026-03-24
Add chat perf markers

pwang347:
Reverts microsoft/vscode-copilot-chat#4508


src/util/common/performance.ts
```diff
@@ -0,0 +1,111 @@
...
+		mark: (name, markOptions) => performance.mark(name, markOptions),
+		getMarks: () => performance.getEntries().filter(e => e.entryType === 'mark').map(e => ({ name: e.name, startTime: e.startTime })),
+		clearMarks: prefix => {
+			for (const entry of performance.getEntries()) {
+				if (entry.entryType === 'mark' && entry.name.startsWith(prefix)) {
+					performance.clearMarks(entry.name);
+				}
+			}
+		},
+	};
```
deepak1556, line 27: Does the same suggestion apply here https://github.com/microsoft/vscode/pull/302950/changes#r2955613369 ?

pwang347: updated, thanks! I hope this polyfill doesn't get used (it's mainly for tests and if things go wrong and we don't get the implementation) but doesn't hurt to update

PR #4680, opened 2026-03-25
Fix bug with Execution Subagent not being disabled (vscode#304648)

vikramnitin9:
Addresses issue with execution subagent showing up even when all tools are disabled ([vscode#304648](https://github.com/microsoft/vscode/issues/304648)).

This adds `executionSubagent` to the family of `execute` tools, meaning that it will show up in the tool picker. Additionally, it disables `canBeReferencedInPrompt` in order to prevent it being triggered by the user.


vikramnitin9:
Sorry @roblourens - could you take another review? I realized that there was an accidentally deleted line in `package.json`, so I just pushed a commit to restore it.

vikramnitin9:
@roblourens - sorry, I caught another bug in `package.json` that had to be fixed 😅
Just pushed a commit and re-requested approval.

PR #4683, opened 2026-03-25
pass sessionId as arguments to prompt elements

aeschli:
Passing around the sessionResource to every prompt element that needs it.


aeschli:
See #4623

PR #4718, opened 2026-03-26
fix: Fix condition to check parts length for model mapping for claude

sukanta1991:
fix for claude models like `claude opus 4.6`, where is was rendered as `claude-opus-4.6` instead of `claude-opus-4-6`


sukanta1991:
@bhavyaus Can you please review?

sukanta1991:
@joaomoreno

sukanta1991:
@bhavyaus @joaomoreno Can I get a review on this please.

PR #4732, opened 2026-03-26
PromptsServiceImpl: avoid using openTextDocument to not trigger validations

aeschli:


ulugbekna:
you can ignore these -- they're not merge blockers AND fail due to another PR from this branch against main 

CleanShot 2026-03-26 at 18 17 50@2x

PR #4742, opened 2026-03-26
Support dynamic prompt variables

pwang347:


src/extension/prompt/node/chatParticipantTelemetry.ts
```diff
@@ -222,9 +222,9 @@ function getSlashCommandForTelemetry(request: vscode.ChatRequest): string {
 		return command;
 	}
 
-	// Built-in skills (copilot-skill:// URIs) are safe to send as plain text
+	// Built-in skills (extension-provided) are safe to send as plain text
 	for (const ref of request.references) {
-		if (URI.isUri(ref.value) && ref.value.scheme === 'copilot-skill') {
+		if (URI.isUri(ref.value) && ref.value.path.includes('/assets/prompts/skills/')) {
```
pwang347: fix this

src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
```diff
@@ -161,19 +162,28 @@ export class CopilotCLISession extends DisposableStore implements ICopilotCLISes
...
 		});
+		this._injectSkillVariables();
 		this.add(toDisposable(() => {
 			this._debugFileLogger.endSession(this.sessionId).catch(err => {
 				this.logService.error('[CopilotCLISession] Failed to end debug log session', err);
 			});
 		}));
 	}
 
+	private _injectSkillVariables(): void {
```
pwang347: rename

src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
```diff
@@ -161,19 +162,28 @@ export class CopilotCLISession extends DisposableStore implements ICopilotCLISes
...
+		const context = this._promptVariablesService.buildTemplateVariablesContext(this.sessionId);
+		if (context) {
+			this._sdkSession.updateOptions({ systemMessage: { mode: 'append', content: context } });
```
DonJayamanne: We don't use `updateOptions` or at least try to avoid it.
As this will not be in the stable API.
Instead we need to add these into the options when we create/get a session.
Please see `createSessionsOptions` in copilotcliSessionService.ts

The `CopilotCLISessionOptions` class is what shoudl be used to store all of these options.

DonJayamanne:
Looks good, apart from the minor change I've requested.

src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
pwang347: fixed

vijayupadya (dismissed):

PR #4751, opened 2026-03-27
Add tool search support for OpenAI Responses API

bhavyaus:


package.nls.json
```diff
@@ -341,6 +341,7 @@
...
 	"github.copilot.config.anthropic.toolSearchTool.enabled": "Enable tool search tool for Anthropic models. When enabled, tools are dynamically discovered and loaded on-demand using natural language search, reducing context window usage when many tools are available.",
 	"github.copilot.config.anthropic.toolSearchTool.mode": "Controls how tool search works for Anthropic models. 'server' uses Anthropic's built-in regex-based tool search. 'client' uses local embeddings-based semantic search for more accurate tool discovery.",
+	"github.copilot.config.responsesApi.toolSearchTool.enabled": "Enable tool search for OpenAI Responses API models. When enabled, tools are dynamically discovered and loaded on-demand using embeddings-based search, reducing context window usage when many tools are available.",
```
Giuspepe, line 344: should we also add `github.copilot.config.responsesApi.toolSearchTool.mode` to control whether we use the server side or client side tool search tool?

Giuspepe:
It finds the tool and calls it, but then displays an error message because of the `namespace` field
image

image


Here in the docs they just set `namespace` to `name` when the tool doesn't have a `namespace`

image

However we don't set a `namespace` at the moment:
image

src/platform/endpoint/node/responsesApi.ts
```diff
@@ -35,16 +38,67 @@ export function createResponsesRequestBody(accessor: ServicesAccessor, options:
...
+				...tool.function,
+				type: 'function',
+				strict: false,
+				parameters: (tool.function.parameters || {}) as Record<string, unknown>,
+				...(isDeferred ? { defer_loading: true } : {}),
+			});
+		}
+	}
+
+	// Build final tools array
```
dileepyavan, line 69: Is this going to happen for every request? can we make this and toolsMap as static on this and use it for all toolserach enabled requests?

src/platform/endpoint/node/responsesApi.ts
```diff
@@ -35,16 +38,67 @@ export function createResponsesRequestBody(accessor: ServicesAccessor, options:
...
+			if (!tool.function.name || tool.function.name.length === 0) {
+				continue;
+			}
+			const isDeferred = shouldDeferTools && !toolDeferralService!.isNonDeferredTool(tool.function.name);
+			functionTools.push({
+				...tool.function,
+				type: 'function',
+				strict: false,
+				parameters: (tool.function.parameters || {}) as Record<string, unknown>,
+				...(isDeferred ? { defer_loading: true } : {}),
```
dileepyavan, line 64: https://developers.openai.com/api/docs/guides/tools-tool-search, Are we grouping this by type like below?:

For maximum token savings, we recommend grouping deferred functions into namespaces or MCP servers with clear, high-level descriptions that give the model a strong overview of what is contained within them, so it can effectively search and load only the relevant functions. As a best practice, aim to keep each namespace to fewer than 10 functions for better token efficiency and model performance.

{
  "tools": [
    {
      "type": "namespace",
      "name": "crm",
      "description": "CRM tools for customer lookup and order management.",
      "tools": [
        {
          "type": "function",
          "name": "list_open_orders",

PR #4761, opened 2026-03-27
In-memory search hint usage

itsibitzi:
Proposal for using the search hint for external search.

Only keeps the search hint in-memory and for a limited amount of time. I think things can get weird with a mismanaged persistent search hint since it effectively allows you to time travel backwards to different versions of the index. If you wanted to you could persist it for longer and/or store it to disk.

### DIFF
--- a/src/platform/workspaceChunkSearch/test/node/externalIngest.spec.ts
+++ b/src/platform/workspaceChunkSearch/test/node/externalIngest.spec.ts
@@ -21,7 +21,6 @@ import { createPlatformServices, TestingServiceCollection } from '../../../test/
 import { IWorkspaceService, NullWorkspaceService } from '../../../workspace/common/workspaceService';
 import { ExternalIngestClient, ExternalIngestFile, ExternalIngestUpdateIndexResult, IExternalIngestClient } from '../../node/codeSearch/externalIngestClient';
 import { ExternalIngestIndex } from '../../node/codeSearch/externalIngestIndex';
-import { StrategySearchSizing, WorkspaceChunkQueryWithEmbeddings } from '../../common/workspaceChunkSearch';
 
 const emptyProgressCb: (message: string) => void = () => { };
 const testCallTracker = new CallTracker('externalIngest.spec.ts');


src/platform/workspaceChunkSearch/node/codeSearch/externalIngestIndex.ts
```diff
@@ -404,6 +391,24 @@ export class ExternalIngestIndex extends Disposable {
...
+	/**
+	 * Returns the search hint if it is still within its lifetime.
+	 */
+	private getSearchHint(): string | undefined {
+		const hint = this._searchHint;
+		if (!hint) {
+			return undefined;
+		}
+
+		if (Date.now() - hint.setAt > ExternalIngestIndex.SEARCH_HINT_LIFETIME_MS) {
```
itsibitzi, line 403: We might want to consider some kind of reuse limit (e.g. 5 uses) in case a user changes their clock backwards.

src/platform/workspaceChunkSearch/node/codeSearch/externalIngestIndex.ts
```diff
@@ -92,18 +92,27 @@ export class ExternalIngestIndex extends Disposable {
...
+	 * Must be consumed within {@link SEARCH_HINT_LIFETIME_MS} of being set.
+	 */
+	private _searchHint: { readonly value: string; readonly setAt: number } | undefined;
+
+	/**
+	 * Search hint only lasts 5 minutes. After that the index should have fully ingested and synchronized the
+	 * fileset to all the clusters. The hint is only valid for a limited amount of time anyway, and keeping it around
+	 * indefinitely could result in some confusion, for example being able to search a deleted fileset.
+	 */
+	private static readonly SEARCH_HINT_LIFETIME_MS = 300_000;
```
itsibitzi, line 114: Metrics show we're syncing in < 10 seconds p99 right now so this is a lot of extra margin fwiw, but I don't think there's much harm as long as the hint is properly bound to the correct fileset.

PR #4768, opened 2026-03-27
fix: normalize Claude model IDs to hyphen format

### LINKED ISSUE #305827: Claude agent passes invalid model ID format (`claude-opus-4.6` instead of `claude-opus-4-6`)
## Bug Description

When using Claude as an agent in Copilot Chat, it passes a dot-format model ID (e.g. `claude-opus-4.6`) to the Claude Code Agent SDK. Claude Code CLI only accepts hyphen-format IDs (`claude-opus-4-6`), causing immediate failure:

```
There's an issue with the selected model (claude-opus-4.6). It may not exist or you may not have access to it.
```

This also reproduces with `claude-sonnet-4.6`.

## Steps to Reproduce

1. Install Claude Code v2.1.84
2. Open Copilot Chat, select Claude agent
3. Send any message
4. Error appears in the Claude Code session immediately but is swallowed on the Copilot Chat side

## Expected Behavior

Copilot Chat should pass valid Anthropic model IDs using hyphens (e.g. `claude-opus-4-6`, `claude-opus-4-6[1m]`) instead of dots.

## Environment

- VS Code: v1.113.0
- VS Code Copilot Chat: v0.41.1
- Claude Code: v2.1.84
- macOS 15

abdullahkhfb:
Replaced dots with hyphens in ClaudeLanguageModelServer.ts  Fixes microsoft/vscode#305827


TylerLeonhardt:
Fixed in https://github.com/microsoft/vscode/pull/308589

PR #4775, opened 2026-03-27
drop images past 20 to unblock error

justschen:
fix issue where claude messages 20+ images are causing errors


justschen:
deferring this to 1.115 and waiting for upstream advice

PR #4785, opened 2026-03-27
Attach response output message IDs only to phased assistant messages

dileepyavan:
Fixes microsoft/vscode#305838.

This change removes the redundant standalone response-output-message-id prompt path for OpenAI Responses API round-tripping and keeps that identifier attached only to phased assistant messages.

What changed:
- Moved `responseOutputMessageId` into phase metadata so phased assistant messages carry the phase and id together.
- Stopped attaching/sending a response output message id for non-phased assistant messages.
- Kept non-phased assistant history in the Responses API request body as plain assistant input messages.
- Simplified assistant prompt reconstruction to read the id from phase metadata instead of a separate opaque payload.
- Added regression coverage for the Responses API request body conversion.

Why:
- The output message id is only meaningful for assistant messages that are emitted with a phase.
- Keeping it separate added extra serialization and prompt plumbing for OpenAI model requests.
- Collapsing it into phase metadata reduces metadata overhead and simplifies the round-trip path.

Validation:
- `npm run test:unit -- src/platform/endpoint/node/test/responsesApi.spec.ts`
- TypeScript watch compile reported 0 errors

Notes:
- The generic streaming delta still exposes `phase` and `responseOutputMessageId` separately so provider-agnostic response handling remains unchanged.
- The simplification is applied in the persisted prompt metadata and resend path, which is where the redundant payload previously affected OpenAI Responses requests.


DonJayamanne (dismissed):

__probe3__
hyperare80-pixel, line 1: remove-

PR #4791, opened 2026-03-28
add prommptC for vsc models and refine hash set with model family nam…

thirteenflt:
…e only to support future models

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -102,7 +102,7 @@
 				"copyfiles": "^2.4.1",
 				"csv-parse": "^6.0.0",
 				"dotenv": "^17.2.0",
-				"electron": "^39.8.4",
+				"electron": "^39.8.5",
 				"esbuild": "^0.25.6",
 				"eslint": "^9.30.0",
 				"eslint-import-resolver-typescript": "^4.4.4",
@@ -11594,9 +11594,9 @@
 			"license": "MIT"
 		},
 		"node_modules/electron": {
-			"version": "39.8.4",
-			"resolved": "https://registry.npmjs.org/electron/-/electron-39.8.4.tgz",
-			"integrity": "sha512-eXYKxr4y+s31xs78keVJYg+XY20tGQMQzyIhZvc5L0XRDH2Gp08mbeFlbR1OjAeM5h5l/T2JYT2MFK2kYe2fMg==",
+			"version": "39.8.5",
+			"resolved": "https://registry.npmjs.org/electron/-/electron-39.8.5.tgz",
+			"integrity": "sha512-q6+LiQIcTadSyvtPgLDQkCtVA9jQJXQVMrQcctfOJILh6OFMN+UJJLRkuUTy8CZDYeCIBn1ZycqsL1dAXugxZA==",
 			"dev": true,
 			"hasInstallScript": true,
 			"license": "MIT",
--- a/package.json
+++ b/package.json
@@ -6297,7 +6297,7 @@
 		"copyfiles": "^2.4.1",
 		"csv-parse": "^6.0.0",
 		"dotenv": "^17.2.0",
-		"electron": "^39.8.4",
+		"electron": "^39.8.5",
 		"esbuild": "^0.25.6",
 		"eslint": "^9.30.0",
 		"eslint-import-resolver-typescript": "^4.4.4",
--- a/src/extension/chatSessions/vscode-node/chatPromptFileService.ts
+++ b/src/extension/chatSessions/vscode-node/chatPromptFileService.ts
@@ -48,13 +48,17 @@ export class ChatPromptFileService extends Disposable implements IChatPromptFile
 			this._onDidChangeSkills.fire();
 		}));
 
-		this._register(vscode.chat.onDidChangeHooks(() => {
-			this._onDidChangeHooks.fire();
-		}));
+		if (vscode.chat.onDidChangeHooks) {
+			this._register(vscode.chat.onDidChangeHooks(() => {
+				this._onDidChangeHooks.fire();
+			}));
+		}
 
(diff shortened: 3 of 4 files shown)


karthiknadig (approved):

zelinms (approved):

src/extension/prompts/node/agent/agentPrompt.tsx
```diff
@@ -169,6 +170,19 @@ export class AgentPrompt extends PromptElement<AgentPromptProps> {
 	private async getSystemPrompt(customizations: AgentPromptCustomizations) {
 		const modelFamily = this.props.endpoint.family ?? 'unknown';
 
+		if (this.props.endpoint.family.startsWith('oswe')) {
```
karthiknadig: You should use hashes here.

src/extension/prompts/node/agent/agentPrompt.tsx
```diff
@@ -185,6 +199,18 @@ export class AgentPrompt extends PromptElement<AgentPromptProps> {
 		/>;
 	}
 
+	private getReminderInstructionsClass(customizations: AgentPromptCustomizations): ReminderInstructionsConstructor {
+		if (this.props.endpoint.family.startsWith('oswe')) {
```
karthiknadig: Same thing as above you should use hash

thirteenflt:
> The agentPrompt.tsx changes and getVSCModelPromptOverride should be removed. Instead, add VSCModelPromptResolverC (and D if needed) following the same registry pattern as A and B. The variant config can be handled inside the resolver's matchesModel or resolveSystemPrompt.

how to handle variant config in resolver? Is configurationService availabel in resolver?

Resolver is determined by variant config. Is it not true? @bhavyaus

bhavyaus:
Right now your PR adds special-case logic into `agentPrompt.tsx` (`getVSCModelPromptOverride`, `getReminderInstructionsClass`). We don't want that — `agentPrompt.tsx` shouldn't need to know about VSC model variants. Instead, all the routing should happen inside `vscModelPrompts.tsx` using the registry.

## What You Should Do

### Step 1: Remove these changes from `agentPrompt.tsx`

- Remove the `import { getVSCModelPromptOverride }` line
- Remove the `getReminderInstructionsClass()` method
- Remove the `if (this.props.endpoint.family.includes('vscode'))` block in `getSystemPrompt()`
- Put `const ReminderInstructionsClass = customizations.ReminderInstructionsClass;` back where it was before

After this, `agentPrompt.tsx` should have **zero changes** from your PR.

### Step 2: Remove `getVSCModelPromptOverride` from `vscModelPrompts.tsx`

You don't need this function at all.

### Step 3: Add new resolvers in `vscModelPrompts.tsx`

Copy the pattern from `VSCModelPromptResolverA` and `VSCModelPromptResolverB` that already exist in the file. Here's exactly what to add:

```typescript
class VSCModelPromptResolverC implements IAgentPrompt {
    static readonly familyPrefixes = ['vscModelC'];

    static async matchesModel(endpoint: IChatEndpoint): Promise<boolean> {
        return isVSCModelC(endpoint);
    }

    resolveSystemPrompt(endpoint: IChatEndpoint): SystemPrompt | undefined {
        return VSCModelPromptC;
    }

    resolveReminderInstructions(endpoint: IChatEndpoint): ReminderInstructionsConstructor | undefined {
        return VSCModelReminderInstructionsC;
    }
}

class VSCModelPromptResolverD implements IAgentPrompt {
    static readonly familyPrefixes = ['vscModelD'];

    static async matchesModel(endpoint: IChatEndpoint): Promise<boolean> {
        return isVSCModelD(endpoint);
    }

    resolveSystemPrompt(endpoint: IChatEndpoint): SystemPrompt | undefined {
        return VSCModelPromptC; // or a different prompt if D needs its own
    }

    resolveReminderInstructions(endpoint: IChatEndpoint): ReminderInstructionsConstructor | undefined {
        return VSCModelReminderInstructionsC;
    }
}
```

Then register them at the bottom of the file alongside A and B:

```typescript
PromptRegistry.registerPrompt(VSCModelPromptResolverA);
PromptRegistry.registerPrompt(VSCModelPromptResolverB);
PromptRegistry.registerPrompt(VSCModelPromptResolverC);  // add
PromptRegistry.registerPrompt(VSCModelPromptResolverD);  // add
```

### Step 4 (if you need variant config): Inject services into the resolver

> "Is configurationService available in resolver?"

**Yes!** The registry creates resolvers using dependency injection, so you can add `configurationService` to the constructor. Here's how:

```typescript
class VSCModelPromptResolverC implements IAgentPrompt {
    static readonly familyPrefixes = ['vscModelC'];

    constructor(
        @IConfigurationService private readonly configurationService: IConfigurationService,
        @IExperimentationService private readonly experimentationService: IExperimentationService,
    ) { }

    static async matchesModel(endpoint: IChatEndpoint): Promise<boolean> {
        return isVSCModelC(endpoint);
    }

    resolveSystemPrompt(endpoint: IChatEndpoint): SystemPrompt | undefined {
        // You can read config/experiment values here!
        const variant = this.configurationService.getExperimentBasedConfig(
            ConfigKey.VSCModelPromptVariant, this.experimentationService);
        if (variant === 'D') {
            return VSCModelPromptD;
        }
        return VSCModelPromptC;
    }

    resolveReminderInstructions(endpoint: IChatEndpoint): ReminderInstructionsConstructor | undefined {
        return VSCModelReminderInstructionsC;
    }
}
```

thirteenflt:
@bhavyaus  @karthiknadig , can you please take another look?

karthiknadig (approved):

bhavyaus (approved):
LGTM! Thanks for the changes!

PR #4869, opened 2026-03-31
nes: fix: don't yield edits when fetch errors or is cancelled

ulugbekna:
fixes https://github.com/microsoft/vscode/issues/306810


ulugbekna:
don't mind -- these're failing because I first incorrectly created the PR against main, but also they're not required and needed

CleanShot 2026-03-31 at 20 16 12@2x

PR #4873, opened 2026-03-31
Make sure we still enable code search when only external ingest is available 

mjbvz:
For https://github.com/microsoft/vscode/issues/306845


mjbvz:
Actually wrong branch? Not seeing a 0.43 yet though. Let me check and reopen a new pr

mjbvz:
Nevermind, looks like 0.42 is the right branch. Was confused since stable vscode 1.113 was using it too

PR #4884, opened 2026-03-31
Add skillContentRead telemetry event for skill provenance tracking

AbhitejJohn:
Adds a new `skillContentRead` telemetry event fired when the `read_file` tool successfully reads a skill file. This is separate from the existing `readFileToolInvoked` event and runs off the critical path via `queueMicrotask`.

Addresses https://github.com/microsoft/vscode/pull/306520#issuecomment-4160588650 — moves skill content telemetry from vscode core into the extension.

## What's captured

**GH telemetry event** (`sendGHTelemetryEvent`) — hashed values:
| Property | Description |
|---|---|
| `skillNameHash` | SHA-256 hash of the skill name |
| `skillStorage` | Storage provenance: `extension`, `internal`, `user`, or `local` |
| `extensionIdHash` | SHA-256 hash of the contributing extension ID (empty if not extension-sourced) |
| `extensionVersion` | Semver version of the contributing extension |
| `contentHash` | SHA-256 hash of the full skill file content |

**Enhanced GH telemetry event** (`sendEnhancedGHTelemetryEvent`) — plaintext, opt-in only:
| Property | Description |
|---|---|
| `skillName` | Skill name in plaintext |
| `skillPath` | Full URI of the skill file |
| `extensionId` | Contributing extension ID |
| `extensionVersion` | Extension version |
| `skillStorage` | Same provenance classification |

## Storage classification (`getSkillStorage`)
- **`extension`** — contributed by a VS Code extension via `extensionPromptFileProvider`
- **`internal`** — built-in skill using `vscode-chat-internal` URI scheme
- **`user`** — personal skill under `userHome` + `PERSONAL_SKILL_FOLDERS` (e.g. `~/.copilot/skills/`)
- **`local`** — workspace-level skill (e.g. `.github/skills/`)

## Design decisions
- Separate event from `readFileToolInvoked` to keep concerns clean
- Deferred via `queueMicrotask` so hashing/lookups don't block the tool response
- Snapshot text captured synchronously before deferring (snapshot may not be valid later)
- `pluginNameHash`/`pluginVersion` deferred until vscode core exposes `IAgentPluginService` metadata (TODO in code)

### DIFF
--- a/src/extension/tools/node/readFileTool.tsx
+++ b/src/extension/tools/node/readFileTool.tsx
@@ -8,11 +8,9 @@ import type * as vscode from 'vscode';
 import { ConfigKey, IConfigurationService } from '../../../platform/configuration/common/configurationService';
 import { ObjectJsonSchema } from '../../../platform/configuration/common/jsonSchema';
 import { ICustomInstructionsService } from '../../../platform/customInstructions/common/customInstructionsService';
-import { PERSONAL_SKILL_FOLDERS } from '../../../platform/customInstructions/common/promptTypes';
 import { NotebookDocumentSnapshot } from '../../../platform/editing/common/notebookDocumentSnapshot';
 import { TextDocumentSnapshot } from '../../../platform/editing/common/textDocumentSnapshot';
 import { IEndpointProvider } from '../../../platform/endpoint/common/endpointProvider';
-import { INativeEnvService } from '../../../platform/env/common/envService';
 import { IExtensionsService } from '../../../platform/extensions/common/extensionsService';
 import { IFileSystemService } from '../../../platform/filesystem/common/fileSystemService';
 import { IAlternativeNotebookContentService } from '../../../platform/notebook/common/alternativeContent';
@@ -132,7 +130,6 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
 		@ICustomInstructionsService private readonly customInstructionsService: ICustomInstructionsService,
 		@IFileSystemService private readonly fileSystemService: IFileSystemService,
 		@IExtensionsService private readonly extensionsService: IExtensionsService,
-		@INativeEnvService private readonly envService: INativeEnvService,
 	) { }
 
 	async invoke(options: vscode.LanguageModelToolInvocationOptions<ReadFileParams>, token: vscode.CancellationToken) {
@@ -379,7 +376,7 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
 		// Reuses extensionSkillInfo/skillInfo already computed above.
 		// TODO: Add pluginNameHash and pluginVersion properties once vscode core's
 		// extensionPromptFileProvider command exposes IAgentPluginService metadata.
-		if (skillInfo && documentSnapshot) {
+		if (skillInfo && documentSnapshot && uri && this.customInstructionsService.isSkillMdFile(uri)) {
 			const content = documentSnapshot instanceof TextDocumentSnapshot ? documentSnapshot.getText() : '';
 			const extensionId = extensionSkillInfo?.extensionId ?? '';
 			const extensionVersion = extensionId ? this.extensionsService.getExtension(extensionId)?.packageJSON?.version ?? '' : '';
@@ -388,10 +385,10 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
 			// Plaintext properties shared by enhanced GH and internal MSFT events
 			const plaintextProps = {
 				skillName: skillInfo.skillName,
-				skillPath: uri!.toString(),
+				skillPath: uri.toString(),
 				extensionId,
 				extensionVersion,
-				skillStorage: this.getSkillStorage(uri!, extensionSkillInfo),
+				skillStorage: skillInfo.storage,
 				contentHash,
 			};
 
@@ -409,29 +406,6 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
 
 			this.telemetryService.sendInternalMSFTTelemetryEvent('skillContentRead', plaintextProps);
 		}
-	}
-
-	/**
-	 * Classifies skill storage provenance. Mirrors the discovery logic in
-	 * {@link CustomInstructionsService._matchInstructionLocationsFromSkills}
-	 * which uses the same {@link PERSONAL_SKILL_FOLDERS} + userHome check.
-	 *
-	 * Skills from config-based locations (`chat.skills.location` setting) fall
-	 * through to `'local'` since they are user-configured workspace-adjacent paths.
-	 */
-	private getSkillStorage(uri: URI, extensionSkillInfo: { extensionId?: string } | undefined): string {
-		if (extensionSkillInfo?.extensionId) {
-			return 'extension';
-		}
-		if (uri.scheme === 'vscode-chat-internal') {
-			return 'internal';
-		}
-		for (const folder of PERSONAL_SKILL_FOLDERS) {
-			if (extUriBiasedIgnorePathCase.isEqualOrParent(uri, extUriBiasedIgnorePathCase.joinPath(this.envService.userHome, folder))) {
-				return 'user';
-			}
-		}
-		return 'local';
 	}
 
 	async resolveInput(input: IReadFileParamsV1, promptContext: IBuildPromptContext): Promise<IReadFileParamsV1> {
--- a/src/extension/tools/node/test/readFile.spec.tsx
+++ b/src/extension/tools/node/test/readFile.spec.tsx
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { afterAll, beforeAll, expect, suite, test } from 'vitest';
-import { ICustomInstructionsService } from '../../../../platform/customInstructions/common/customInstructionsService';
+import { ICustomInstructionsService, SkillStorage } from '../../../../platform/customInstructions/common/customInstructionsService';
 import { IExtensionsService } from '../../../../platform/extensions/common/extensionsService';
 import { IFileSystemService } from '../../../../platform/filesystem/common/fileSystemService';
 import { MockFileSystemService } from '../../../../platform/filesystem/node/test/mockFileSystemService';
@@ -763,7 +763,7 @@ suite('ReadFile', () => {
 
 			const event = telemetry.events.find(e => e.eventName === 'skillContentRead');
 			expect(event).toBeDefined();
-			expect(event!.properties!.skillStorage).toBe('local');
+			expect(event!.properties!.skillStorage).toBe(SkillStorage.Workspace);
 			expect(event!.properties!.skillNameHash).not.toBe('');
 			expect(event!.properties!.extensionIdHash).toBe('');
 			expect(event!.properties!.extensionVersion).toBe('');
@@ -775,7 +775,7 @@ suite('ReadFile', () => {
 			expect(enhanced!.properties!.skillPath).toBe(skillUri.toString());
 			expect(enhanced!.properties!.extensionId).toBe('');
 			expect(enhanced!.properties!.extensionVersion).toBe('');
-			expect(enhanced!.properties!.skillStorage).toBe('local');
+			expect(enhanced!.properties!.skillStorage).toBe(SkillStorage.Workspace);
 			expect(enhanced!.properties!.contentHash).not.toBe('');
 
 			const internal = telemetry.internalEvents.find(e => e.eventName === 'skillContentRead');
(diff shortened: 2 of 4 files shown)


src/extension/tools/node/readFileTool.tsx
```diff
@@ -370,6 +376,82 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
...
+			if (!skillInfo) {
+				return;
+			}
+
+			const skillStorage = this.getSkillStorage(uri, extensionSkillInfo);
+			const extensionId = extensionSkillInfo?.extensionId;
+			const extensionIdHash = extensionId ? getCachedSha256Hash(extensionId) : '';
+			const extensionVersion = extensionId ? this.extensionsService.getExtension(extensionId)?.packageJSON?.version ?? '' : '';
+			const contentHash = content ? getCachedSha256Hash(content) : '';
+			const skillNameHash = getCachedSha256Hash(skillInfo.skillName);
```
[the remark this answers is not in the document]
AbhitejJohn: Addressed these comments.

AbhitejJohn:
@pwang347 : Thank you. looped you in on a mail. It would be nice to have this in sooner. I'm happy to restructure this to fit with the new tool when that's ready as well. 

Also tagging @aeschli on this PR.

src/extension/tools/node/readFileTool.tsx
```diff
@@ -370,6 +376,86 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
...
+		// Only capture text now that we know this is a skill file.
+		const content = documentSnapshot instanceof TextDocumentSnapshot ? documentSnapshot.getText() : '';
+
+		// Defer hashing and telemetry sends off the read_file critical path.
+		queueMicrotask(() => {
+			const skillStorage = this.getSkillStorage(uri, extensionSkillInfo);
+			const extensionId = extensionSkillInfo?.extensionId;
+			const extensionIdHash = extensionId ? getCachedSha256Hash(extensionId) : '';
+			const extensionVersion = extensionId ? this.extensionsService.getExtension(extensionId)?.packageJSON?.version ?? '' : '';
+			// Use non-caching hash for content to avoid retaining full file text in the
```
pwang347: I would prefer to cache here, I think the same skills will still be read potentially multiple times in a session and storing a 256 bit entry here doesn't seem bad.

pwang347:
@AbhitejJohn thank you for looping me in, the motivation makes sense. Will take a closer look! I think we can start here and just be careful about porting it over later when we refactor things.

src/extension/tools/node/readFileTool.tsx
```diff
@@ -370,6 +376,86 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
...
+			// global hash cache indefinitely. Skill names and extension IDs are small and
+			// repeat often, so getCachedSha256Hash is appropriate for those.
+			const contentHash = content ? createSha256HashSyncInsecure(content) : '';
+			const skillNameHash = getCachedSha256Hash(skillInfo.skillName);
```
pwang347: @aeschli should these be SHA-1 for consistency/slight perf? I understand that we don't have the helpers available in ext. currently though. Maybe we can just copy from https://github.com/microsoft/vscode/tree/main/src/vs/base/common/hash.ts?

src/extension/tools/node/readFileTool.tsx
```diff
@@ -176,6 +181,7 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
 			ranges = getParamRanges(options.input, documentSnapshot);
 
 			void this.sendReadFileTelemetry('success', options, ranges, uri);
+			void this.sendSkillContentReadTelemetry(uri, documentSnapshot);
```
pwang347: nitpick: can we just move this helper inside of `this.sendReadFileTelemetry`, there are already a few skill-related calls made there and we can leverage those instead of calling again

src/extension/tools/node/readFileTool.tsx
```diff
@@ -370,6 +376,86 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
...
+		const skillInfo = extensionSkillInfo || this.customInstructionsService.getSkillInfo(uri);
+		if (!skillInfo) {
+			return;
+		}
+
+		// Only capture text now that we know this is a skill file.
+		const content = documentSnapshot instanceof TextDocumentSnapshot ? documentSnapshot.getText() : '';
+
+		// Defer hashing and telemetry sends off the read_file critical path.
+		queueMicrotask(() => {
```
pwang347: if we move inside the other telemetry call we should be able to remove this (since it's a `void` async func)

src/extension/tools/node/readFileTool.tsx
```diff
@@ -370,6 +376,86 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
...
+
+	/**
+	 * Classifies skill storage provenance. Mirrors the discovery logic in
+	 * {@link CustomInstructionsService._matchInstructionLocationsFromSkills}
+	 * which uses the same {@link PERSONAL_SKILL_FOLDERS} + userHome check.
+	 *
+	 * Skills from config-based locations (`chat.skills.location` setting) fall
+	 * through to `'local'` since they are user-configured workspace-adjacent paths.
+	 */
+	private getSkillStorage(uri: URI, extensionSkillInfo: { extensionId?: string } | undefined): string {
```
pwang347: I think we can add this to the `_matchInstructionLocationsFromSkills` observable directly to return this information, so no need to do extra work to loop etc.

src/extension/tools/node/readFileTool.tsx
```diff
@@ -370,6 +376,86 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
...
+		// Defer hashing and telemetry sends off the read_file critical path.
+		queueMicrotask(() => {
+			const skillStorage = this.getSkillStorage(uri, extensionSkillInfo);
+			const extensionId = extensionSkillInfo?.extensionId;
+			const extensionIdHash = extensionId ? getCachedSha256Hash(extensionId) : '';
+			const extensionVersion = extensionId ? this.extensionsService.getExtension(extensionId)?.packageJSON?.version ?? '' : '';
+			// Use non-caching hash for content to avoid retaining full file text in the
+			// global hash cache indefinitely. Skill names and extension IDs are small and
+			// repeat often, so getCachedSha256Hash is appropriate for those.
+			const contentHash = content ? createSha256HashSyncInsecure(content) : '';
```
pwang347: could consider caching for <=1mb files maybe?

pwang347: the content here are unlikely to change often, so I think cache would be good

src/extension/tools/node/readFileTool.tsx
```diff
@@ -370,6 +376,86 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
...
+				{
+					skillNameHash,
+					skillStorage,
+					extensionIdHash,
+					extensionVersion,
+					contentHash,
+				}
+			);
+
+			this.telemetryService.sendEnhancedGHTelemetryEvent('skillContentRead',
```
digitarald: Should also be send to MSFT internal telemetry if we can.

src/extension/tools/node/readFileTool.tsx
```diff
@@ -370,6 +376,86 @@ export class ReadFileTool implements ICopilotTool<ReadFileParams> {
...
+				{
+					skillName: skillInfo.skillName,
+					skillPath: uri.toString(),
```
digitarald: Can we detect if a skill came from a plugin and report that metadata here as well?

src/extension/tools/node/readFileTool.tsx
pwang347: just another thought, do we need to include the hash at all if the file is not the SKILL.md? I think this will also be called on content in the skill folder - maybe we should add that in the telemetry as well, whether the file is the SKILL.md file versus some other asset?

AbhitejJohn: We only capture this telemetry for skill files, we bail out early for anything that isn't a `skill.md`. So we wont really capture contents of any other assets at the moment (even if they are in the same directory.

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: @digitarald : Sure, can. how do we differentiate between internalMicrosoft and ghEnhanced? I put in gh since most other telemetry we capture is all in gh.

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: We can, I have a todo in place to do that. That needs a cross repo change from what I can tell. We'd want to make a change in VSCode core to expose that data and then consume it here. This would likely be a follow-up.

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: Yeah, I can add back the cache. It probably doesn't make sense skipping it here.

aeschli:
Note: in the core PR, When listing all available skills for a request, I switched this to use our own (cheap) hash, as discussed [here](https://github.com/microsoft/vscode/pull/303110#issuecomment-4121760156).

We should use the same hash, or do you prefer to switch to a well known hash?

https://github.com/microsoft/vscode/blob/b35761f579dc05778eef60227a4c97d349202f83/src/vs/workbench/contrib/chat/common/promptSyntax/computeAutomaticInstructions.ts#L163-L164
https://github.com/microsoft/vscode/blob/b35761f579dc05778eef60227a4c97d349202f83/src/vs/base/common/hash.ts#L17-L18

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: Done — merged the skill telemetry logic into `sendReadFileTelemetry` so it reuses the already-computed `extensionSkillInfo`/`skillInfo`. No more duplicate lookups.

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: Done — removed `queueMicrotask`. `sendReadFileTelemetry` is already `async` and called with `void` (fire-and-forget), so the deferral was redundant.

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: Done — switched back to `getCachedSha256Hash(content)`. Skill files are small (1-5KB) and there are few unique ones per session, so the cache overhead is negligible.

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: Done — added `sendInternalMSFTTelemetryEvent('skillContentRead', plaintextProps)` with the same plaintext properties as the enhanced GH event (skillName, skillPath, extensionId, extensionVersion, skillStorage, contentHash).

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: If this code base uses sha256 for hashing some of the other properties, it might be better to keep this as is for consistency.

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: I think that makes sense. That seems like that would touch a bunch of files. Do we want to make that change as a follow-up?

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: Talked to @aeschli and we decided to use the simple hash here as well, just to make it easier to correlate.

AbhitejJohn: Updated.

src/extension/tools/node/readFileTool.tsx
pwang347: it should just be in one more file, I can make the change myself as well

src/extension/tools/node/readFileTool.tsx
pwang347: to my knowledge this would currently hit assets because we aren't calling the `isSkillMdFile` helper

pwang347:
Hi @AbhitejJohn, I made the changes in https://github.com/microsoft/vscode-copilot-chat/pull/4914 - we can either merge there, or feel free to port the changes to your PR. Thanks!

AbhitejJohn:
@pwang347 : awesome, thank you. If you'd like to change the target branch for your PR to this, happy to pull it in here too. Just waiting on an approval, if we are all good, on this so we can get this in :)

pwang347:
@AbhitejJohn no problem! I think it may be easier for you to just replicate the changes here if possible as I have that branch built off microsoft/main, so there are several merge conflicts

AbhitejJohn:
@pwang347 : Sounds good, updated and pulled this branch up to the latest main as well.

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: Updated in the latest from the PR that Paul had.

src/extension/tools/node/readFileTool.tsx
AbhitejJohn: Updated in the latest based on a PR Paul had for this. thanks for putting that in, it looks much cleaner now.

pwang347 (approved):

PR #4916, opened 2026-04-01
Yemohyle/subagent telem

yemohyleyemohyle:
Add persistent headerRequestId and parentToolCallId to the search and execute subagent telemetry for better tracking.


src/extension/prompt/node/chatMLFetcher.ts
```diff
@@ -1086,6 +1087,8 @@ export class ChatMLFetcherImpl extends AbstractChatMLFetcher {
 		});
 
 		const modelRequestId = getRequestId(connection.responseHeaders);
+		// Preserve ourRequestId as headerRequestId if the server didn't echo x-request-id
+		modelRequestId.headerRequestId = modelRequestId.headerRequestId || ourRequestId;
```
roblourens, line 1091: What's the impact of using `modelRequestId.headerRequestId` though? It's confusing when we use that vs our ID and I don't want to mess up this telemetry

yemohyleyemohyle: x-request-id was empty which caused undefined headerRequestId for subagents.

roblourens: Wasn't it using "our" request id before though? Should we prefer that since we seem to use it in most places?

yemohyleyemohyle: These two lines:
```
1398    const modelRequestId = getRequestId(response.headers);
1399    telemetryData.extendWithRequestId(modelRequestId);
```
in the original version changed the headerRequestId from ourRequestId to potentially empty when the response headers do not contain x-request-id.

roblourens: Ah I follow it now, thanks

roblourens (approved):

justschen (approved):

PR #4931, opened 2026-04-02
Update to Responses API compaction and telemetry

dileepyavan:
## Summary
- Ensure previous responseId is being passed when ignoreStatefulMarker is false.
- add server-side compaction outcome telemetry and regression coverage for enabled and disabled compaction flows

## Details
- `createResponsesRequestBody` now computes the compaction threshold once and uses it to gate both request `context_management` and websocket stateful-marker slicing behavior
- websocket and pass-through `OpenAIResponsesProcessor` call sites now receive the request compaction threshold from the request body
- `responsesApi.compactionOutcome` emits only when compaction is enabled and reports the outcome, request IDs, model, threshold, and token counts
- compaction telemetry intentionally does not include `completionId` or `compactionMessageId`

## Testing
- `runTests` on `src/platform/endpoint/node/test/responsesApi.spec.ts` (16 passed)
- verified `start-watch-tasks` output showed no compile errors

### DIFF
--- a/src/platform/endpoint/node/responsesApi.ts
+++ b/src/platform/endpoint/node/responsesApi.ts
@@ -19,7 +19,7 @@ import { ILogService } from '../../log/common/logService';
 import { FinishedCallback, IResponseDelta, OpenAiResponsesFunctionTool } from '../../networking/common/fetch';
 import { IChatEndpoint, ICreateEndpointBodyOptions, IEndpointBody } from '../../networking/common/networking';
 import { ChatCompletion, FinishedCompletionReason, modelsWithoutResponsesContextManagement, openAIContextManagementCompactionType, OpenAIContextManagementResponse, rawMessageToCAPI, TokenLogProb } from '../../networking/common/openai';
-import { sendEngineMessagesTelemetry } from '../../networking/node/chatStream';
+import { sendEngineMessagesTelemetry, sendResponsesApiCompactionTelemetry } from '../../networking/node/chatStream';
 import { IExperimentationService } from '../../telemetry/common/nullExperimentationService';
 import { ITelemetryService } from '../../telemetry/common/telemetry';
 import { TelemetryData } from '../../telemetry/common/telemetryData';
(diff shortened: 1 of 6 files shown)


src/platform/endpoint/node/responsesApi.ts
```diff
@@ -103,17 +124,33 @@ interface ResponseOutputItemWithPhase {
...
+	if (statefulMarkerAndIndex) {
 		previousResponseId = statefulMarkerAndIndex.statefulMarker;
+
+		// Requests that resume from previous_response_id send only post-marker history,
+		// but they still need the latest compaction item even when that item predates
+		// the marker. This keeps both websocket and non-websocket traffic aligned.
 		messages = messages.slice(statefulMarkerAndIndex.index + 1);
+		if (latestCompactionMessageIndex !== undefined) {
+			if (latestCompactionMessageIndex > statefulMarkerAndIndex.index) {
+				messages = messages.slice(latestCompactionMessageIndex - (statefulMarkerAndIndex.index + 1));
```
chrmarti, line 147: We still set previousResponseId in this case?

src/platform/endpoint/node/responsesApi.ts
```diff
@@ -103,17 +124,33 @@ interface ResponseOutputItemWithPhase {
...
+	readonly outputIndex: number;
+}
+
 function rawMessagesToResponseAPI(modelId: string, messages: readonly Raw.ChatMessage[], ignoreStatefulMarker: boolean): { input: OpenAI.Responses.ResponseInputItem[]; previous_response_id?: string } {
 	const latestCompactionMessageIndex = getLatestCompactionMessageIndex(messages);
-	if (latestCompactionMessageIndex !== undefined) {
-		messages = messages.slice(latestCompactionMessageIndex);
-	}
-
+	const latestCompactionMessage = latestCompactionMessageIndex !== undefined ? createCompactionRoundTripMessage(messages[latestCompactionMessageIndex]) : undefined;
```
chrmarti, line 134: nit: looks like this could be moved into the if block below

src/platform/endpoint/node/responsesApi.ts
dileepyavan: This PR is no longer valid. I had to leave this, thought I closed it and made some changes in this commit: https://github.com/microsoft/vscode/commit/5207a1ab202d32b2b8672e21fcf2658a6f0e7a5c

dileepyavan: validated the changes and for websockets compaction is being sent.

PR #4943, opened 2026-04-02
Add a new fetchedValue concept

lramos15:
Many values in CopilotChat are effectively just values we fetch and then want to keep up to date with an upstream server via polling or refreshing once stale on next fetch. We should provide easy to use utils for this

### DIFF
--- a/src/platform/networking/common/capiClientFetchedValue.ts
+++ b/src/platform/networking/common/capiClientFetchedValue.ts
@@ -17,8 +17,6 @@ import type { IEnvService } from '../../env/common/envService';
 export interface CapiClientFetchedValueOptions<T> {
 	/**
 	 * The request options passed to {@link ICAPIClientService.makeRequest}.
-	 * May be a static object, a synchronous factory, or an async factory
-	 * (useful when headers depend on runtime state such as an auth token).
 	 */
 	readonly request: (() => MakeRequestOptions | Promise<MakeRequestOptions>);
 
@@ -97,7 +95,7 @@ export function createCapiClientFetchedValue<T>(
 		},
 		httpFetch: async (httpRequest) => {
 			const response = await capiClientService.makeRequest<Response>({
-				...httpRequest.state,
+				...(httpRequest.state ?? {}),
 				method: httpRequest.method,
 				// Use the headers from the middleware pipeline (may include
 				// If-None-Match, If-Modified-Since, etc.)


src/platform/networking/common/capiClientFetchedValue.ts
```diff
@@ -0,0 +1,134 @@
...
+ * });
+ *
+ * const fresh = await config.resolve();
+ * ```
+ */
+export function createCapiClientFetchedValue<T>(
+	capiClientService: ICAPIClientService,
+	envService: IEnvService,
+	options: CapiClientFetchedValueOptions<T>,
+): FetchedValue<T> {
```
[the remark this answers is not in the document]
lramos15: All of this is tested in the shared util folder itself. The wrapper is quite small

src/shared-fetch-utils/common/fetchedValue.ts
```diff
@@ -0,0 +1,134 @@
...
+		this._throwIfDisposed();
+		if (!force && this._value !== undefined && !this._isStale(this._value)) {
+			return this._value;
+		}
+		if (this._inflightFetch) {
+			return this._inflightFetch;
+		}
+		this._inflightFetch = this._doFetch();
+		try {
+			return await this._inflightFetch;
```
mjbvz, line 102: If this throws, should we store the `error` and treat it as the cached value? Otherwise when an error occurs we will keep trying to call `doFetch` each time the value is accessed

src/shared-fetch-utils/common/fetchedValue.ts
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * // Synchronous read — may be undefined before first resolve
+ * const current = token.value;
+ *
+ * // Async — fetches only when stale
+ * const fresh = await token.resolve();
+ * ```
+ */
+export class FetchedValue<T> {
+	private _value: T | undefined;
```
mjbvz, line 57: Kind of silly state but what if the value is undefined? Like you have `FetchedValue<string[] | undefined>`

src/shared-fetch-utils/common/fetchedValue.ts
```diff
@@ -0,0 +1,134 @@
...
+	isStale: (value: T) => boolean;
+
+	/**
+	 * When `true`, automatically calls {@link FetchedValue.resolve} once per minute so
+	 * the synchronous {@link FetchedValue.value} getter stays up-to-date.
+	 *
+	 * **Caution:** enabling this will lead to more network requests because the value
+	 * is re-fetched periodically regardless of whether it is being read.
+	 */
+	keepCacheHot?: boolean;
```
mjbvz, line 28: This seems like it could trigger a lot of requests. Could the default be much higher? Also maybe give it a name that makes the potential cost more obvious. Something like `autoRefetchInBackground`?

lramos15: In theory it shouldn't as `isStale` should still be true. This is just basically auto calling `resolve` on a cadence. If the value isn't stale (i.e. the token isn't expired) then it shouldn't be generating many requests.

The worry about shortening the timeout is that then the value might not be hot. An example use case of this is the Auto mode service where we want the auto token to be fresh for when a user starts a new conversation to avoid having to have the latency hit of fetching a fresh token

src/shared-fetch-utils/common/fetchedValue.ts
lramos15: Copilot made this comment and I thought it was a silly / unlikely possibility but I guess we should guard against it just in case will followup with a fix

src/shared-fetch-utils/common/fetchedValue.ts
lramos15: I think the middleware should be responsible for this, but I'm open to your thoughts. 

There's cases where we may throw an error (I was debating changing fetching when window is inactive to throw) where re-calling is ok as the error state could've changed

PR #4972, opened 2026-04-03
Add provider ordering hint for chat model picker

kevin-m-kent:
## Summary
- add `github.copilot.chat.modelPickerVendorOrdering` to emit a provider-based ordering hint for the chat model picker
- assign `vendorPriority` by provider so OpenAI models sort ahead of Anthropic, Gemini, and other providers
- leave top-level promotion to VS Code's existing models control manifest instead of hardcoding featured models in the extension

## Testing
- enabled the setting in a local OSS dev build and verified the picker orders models by provider while top-level featured models come from the manifest

### DIFF
--- a/package.json
+++ b/package.json
@@ -3034,15 +3034,6 @@
 							"onExp"
 						]
 					},
-					"github.copilot.chat.modelPickerVendorOrdering": {
-						"type": "boolean",
-						"default": false,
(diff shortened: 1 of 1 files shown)


justschen:
just a heads up for @lramos15 and @sandy081 !

@kevin-m-kent @cwebster-99  if we want to take this in exp for this iteration, we can merge by eod today and test on monday!

PR #4979, opened 2026-04-04
Allow Mermaid diagrams in Plan mode

OrenMe:
This PR enables Mermaid diagrams in Plan mode behind an explicit configuration gate and updates the Plan prompt so Mermaid diagrams are rendered consistently when they materially help a plan.

Key changes:
- Add a Plan-specific Mermaid setting and expose it in the settings UI
- Gate the Mermaid render tool and Plan prompt instructions behind that setting
- Restore the Plan style-guide wrapper and keep Mermaid instructions concise and explicit
- Add regression coverage for the Plan agent generated content
- Update the switch-to-Plan path so it reflects the same Plan-mode Mermaid guidance

Refs microsoft/vscode#307804.


pierceboggan:
@bhavyaus This looks like a nice addition. What's your perspective on the setting? IMHO we should not introduce another one but curious on your perspective.

OrenMe:
Thanks @pierceboggan 
Agree about the setting option, thought this should only be needed for experimentation to see if it's stable but moving forward I would also remove it

OrenMe:
@pierceboggan @bhavyaus what are your thoughts on this?

PR #5008, opened 2026-04-06
Fix haiku output config

24anisha:
In typical chat conversations, thinking may not be enabled by default, so even if you set Haiku as the model and it has undefined supportsReasoningEffort, the code path never enters the problematic block because thinkingConfig is undefined.
The bug only manifests when:

enableThinking is true in the request options, AND
A reasoningEffort is configured, AND
endpoint.supportsReasoningEffort is undefined/empty
The search subagent explicitly enables thinking (probably by design for better search results), so it hits all three conditions. The main agent in regular chat doesn't enable thinking by default, so it never gets to the problematic code.

My fix ensures that even if thinking is enabled, we don't send output_config to models that don't advertise support for it.


alexdima:

PR #5027, opened 2026-04-07
Fix race condition in resolveAutoModeEndpoint causing duplicate telem…

abadawi591:
## Description

Concurrent calls to `resolveAutoModeEndpoint` for the same conversation race on an empty cache, causing the `automode.routerFallback` telemetry event to fire 2–5× per routing decision instead of once. This inflates telemetry counts by ~2.1×.

## Root cause

In agent-mode, a single tool-calling round calls `getChatEndpoint(request)` multiple times (for `getAvailableTools`, `buildPrompt`, token counting, and fetch). On the first turn of a conversation, the cache entry doesn't exist yet. Since `resolveAutoModeEndpoint` is async, multiple callers read `_autoModelCache` as empty before the first caller finishes and writes the result. Each enters `_tryRouterSelection`, emits telemetry, and redundantly fetches a CAPI token.

    Call 1: cache empty    → async work             → emit telemetry → write cache
    Call 2: cache empty    → async work             → emit telemetry → write cache  ← duplicate
    Call 3: cache empty    → async work             → emit telemetry → write cache  ← duplicate

## Fix

Add per-conversation promise coalescing via a `_pendingResolutions` map. If a resolution is already in-flight for a `conversationId`, subsequent callers await the same promise.

    Call 1: no pending     → start work             → emit telemetry → write cache
    Call 2: pending exists → await Call 1's promise  → done (no duplicate)
    Call 3: pending exists → await Call 1's promise  → done (no duplicate)

## Impact

- **Telemetry:** `automode.routerFallback` event count reduced ~2× to accurate levels
- **Performance:** eliminates redundant CAPI token fetches on first turn
- **User-facing:** none (same model selection behavior)

## Files changed

- `automodeService.ts` — promise coalescing wrapper
- `automodeService.spec.ts` — 5 new tests (concurrent coalescing, single CAPI call, sequential still works, error propagation, retry after failure)

### DIFF
--- a/src/platform/endpoint/node/automodeService.ts
+++ b/src/platform/endpoint/node/automodeService.ts
@@ -186,9 +186,14 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 			return pending;
 		}
 
-		const promise = this._resolveAutoModeEndpointCore(chatRequest, conversationId, knownEndpoints);
-		this._pendingResolutions.set(conversationId, promise);
-		return promise.finally(() => this._pendingResolutions.delete(conversationId));
+		const corePromise = this._resolveAutoModeEndpointCore(chatRequest, conversationId, knownEndpoints);
+		const pendingPromise = corePromise.finally(() => {
+			if (this._pendingResolutions.get(conversationId) === pendingPromise) {
+				this._pendingResolutions.delete(conversationId);
+			}
+		});
+		this._pendingResolutions.set(conversationId, pendingPromise);
+		return pendingPromise;
 	}
 
 	private async _resolveAutoModeEndpointCore(chatRequest: ChatRequest | undefined, conversationId: string, knownEndpoints: IChatEndpoint[]): Promise<IChatEndpoint> {
--- a/src/platform/endpoint/node/automodeService.ts
+++ b/src/platform/endpoint/node/automodeService.ts
@@ -179,21 +179,34 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 
 		const conversationId = chatRequest?.sessionResource?.toString() ?? chatRequest?.sessionId ?? 'unknown';
 
-		// Coalesce concurrent calls for the same conversation to avoid duplicate
-		// router calls and telemetry emissions.
-		const pending = this._pendingResolutions.get(conversationId);
-		if (pending) {
-			return pending;
-		}
+		// Coalesce concurrent calls for the same conversation so we don't fire
+		// duplicate router requests or telemetry. Concurrent callers within a
+		// single turn always share the same prompt, so reusing the result is safe.
+		return this._singleFlight(conversationId, () => this._resolveAutoModeEndpointCore(chatRequest, conversationId, knownEndpoints));
+	}
 
-		const corePromise = this._resolveAutoModeEndpointCore(chatRequest, conversationId, knownEndpoints);
-		const pendingPromise = corePromise.finally(() => {
-			if (this._pendingResolutions.get(conversationId) === pendingPromise) {
-				this._pendingResolutions.delete(conversationId);
+	/**
+	 * Single-flight coalescer: if a resolution for `key` is already in progress,
+	 * return that pending promise instead of starting a new one.
+	 *
+	 * This function MUST remain synchronous (not `async`). The atomicity of the
+	 * `get -> check -> set` sequence relies on no `await` running between them, and
+	 * keeping the function non-`async` makes that invariant structural rather than
+	 * a comment a future edit might miss.
+	 */
+	private _singleFlight(key: string, fn: () => Promise<IChatEndpoint>): Promise<IChatEndpoint> {
+		const existing = this._pendingResolutions.get(key);
+		if (existing) {
+			return existing;
+		}
+		const promise = fn().finally(() => {
+			// Reference-equality guard: only clear if this entry is still ours.
+			if (this._pendingResolutions.get(key) === promise) {
+				this._pendingResolutions.delete(key);
 			}
 		});
-		this._pendingResolutions.set(conversationId, pendingPromise);
-		return pendingPromise;
+		this._pendingResolutions.set(key, promise);
+		return promise;
 	}
 
 	private async _resolveAutoModeEndpointCore(chatRequest: ChatRequest | undefined, conversationId: string, knownEndpoints: IChatEndpoint[]): Promise<IChatEndpoint> {
--- a/src/platform/endpoint/node/automodeService.ts
+++ b/src/platform/endpoint/node/automodeService.ts
@@ -117,6 +117,7 @@ export interface IAutomodeService {
 export class AutomodeService extends Disposable implements IAutomodeService {
 	readonly _serviceBrand: undefined;
 	private readonly _autoModelCache: Map<string, AutoModelCacheEntry> = new Map();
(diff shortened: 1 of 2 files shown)


src/platform/endpoint/node/test/automodeService.spec.ts
```diff
@@ -916,4 +916,101 @@ describe('AutomodeService', () => {
...
+			mockApiResponse(['gpt-4o', 'gpt-4o-mini']);
+			automodeService = createService();
+
+			const chatRequest: Partial<ChatRequest> = {
+				location: ChatLocation.Panel,
+				prompt: 'hello',
+				sessionId: 'session-concurrent'
+			};
+
+			const [result1, result2, result3] = await Promise.all([
```
jinu-jang, line 931: If I understand correctly, shouldn't the real test here be multiple different prompts being asked? Concurrent calls with the same conversationId only breaks our logic if they result in different answers.

My confusion may stem from my lack of understanding on what the `mockApiResponse(['gpt-4o', 'gpt-4o-mini']);` line does

src/platform/endpoint/node/automodeService.ts
```diff
@@ -175,6 +178,20 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 		}
 
 		const conversationId = chatRequest?.sessionResource?.toString() ?? chatRequest?.sessionId ?? 'unknown';
+
+		// Coalesce concurrent calls for the same conversation to avoid duplicate
+		// router calls and telemetry emissions.
+		const pending = this._pendingResolutions.get(conversationId);
```
jinu-jang: Good fix. Can you wrap this out into a synchronous function? Right now this logic is still defined within an `async` function, but the race condition was fixed because there are no `await` being called within this async function.

If people don't remember this detail, this would become very prone to awaits being added again.

Below is a suggestion by Claude on how a synchronous function can be added here.

```typescript
	async resolveAutoModeEndpoint(chatRequest: ChatRequest | undefined, knownEndpoints: IChatEndpoint[]): Promise<IChatEndpoint> {
		if (!knownEndpoints.length) {
			throw new Error('No auto mode endpoints provided.');
		}
		const conversationId = chatRequest?.sessionResource?.toString() ?? chatRequest?.sessionId ?? 'unknown';
		// Coalesce concurrent calls for the same conversation so we don't fire
		// duplicate router requests or telemetry. Concurrent callers within a
		// single turn always share the same prompt, so reusing the result is safe.
		return this._singleFlight(conversationId, () => this._resolveAutoModeEndpointCore(chatRequest, conversationId, knownEndpoints));
	}

	/**
	 * Single-flight coalescer: if a resolution for `key` is already in progress,
	 * return that pending promise instead of starting a new one.
	 *
	 * This function MUST remain synchronous (not `async`). The atomicity of the
	 * `get → check → set` sequence relies on no `await` running between them, and
	 * keeping the function non-`async` makes that invariant structural rather than
	 * a comment a future edit might miss.
	 */
	private _singleFlight(key: string, fn: () => Promise<IChatEndpoint>): Promise<IChatEndpoint> {
		const existing = this._pendingResolutions.get(key);
		if (existing) {
			return existing;
		}
		const promise = fn().finally(() => {
			// Reference-equality guard: only clear if this entry is still ours.
			if (this._pendingResolutions.get(key) === promise) {
				this._pendingResolutions.delete(key);
			}
		});
		this._pendingResolutions.set(key, promise);
		return promise;
	}
```

src/platform/endpoint/node/test/automodeService.spec.ts
abadawi591: Good question! The concurrent calls in the real bug all carry the same prompt and same `chatRequest`.
they're the same user message being resolved multiple times within a single tool-calling round (for getAvailableTools, buildPrompt, token counting, fetch). They're not different user messages arriving at the same time.

So the test is accurate to the real scenario: same chatRequest, same sessionId, same prompt, fired concurrently.

To clarify mockApiResponse(['gpt-4o', 'gpt-4o-mini']): it stubs the CAPI token endpoint to return those as available_models. It controls which models the auto-mode service can choose from — it's not the router response (that would be mockRouterResponse).

Different prompts within the same conversation are already handled by the skipRouter cache (turnCount > 0 → skip), which is unrelated to this fix. This PR only addresses the race window on the first resolution of a conversation.

abadawi591: <html>
<body>
<html><head></head><body><p>Dug a little deeper into this one 👇</p>
<h2>Duplicate <code>automode.routerFallback</code> telemetry from concurrent <code>getChatEndpoint</code> calls</h2>
<h3>Symptom</h3>
<p><code>automode.routerFallback</code> fires 5× per routing decision instead of once. Bursts appear at the same sub-second timestamp, inflating telemetry counts and causing redundant CAPI token fetches.</p>
<h3>Cause</h3>
<p><code>toolCallingLoop.ts</code> calls <code>getChatEndpoint(request)</code> up to 5 times per tool-calling round (for tools, prompt, tokenizer, fetch). On the first resolution of a conversation, these concurrent async calls all see an empty <code>_autoModelCache</code>, all enter <code>_tryRouterSelection</code>, and all emit telemetry before any of them writes the cache.</p>
<p>This pattern dates back to the initial repo when <code>getChatEndpoint</code> was cheap and stateless. As auto-mode features were layered on, existing cross-turn dedup never covered the within-turn concurrent race:</p>

Date | Hash | Author | Change
-- | -- | -- | --
2025-06-27 | af813ee | @kieferrm | Initial repo — multiple getChatEndpoint calls in runOne(), no auto-mode
2025-10-30 | 5aafc56 | @lramos15 | Auto-mode service created, calls now async
2026-01-16 | fe88d62 | @TylerLeonhardt | Model router added, calls now trigger router API
2026-01-20 | 5388442 | @Copilot/@TylerLeonhardt | lastRoutedPrompt dedup — fixes cross-turn but not within-turn race
2026-02-05 | 70b2b4c | @lramos15 | hasImage gate + vision fallback
2026-02-19 | 8f0a331 | @nickthecook | automode.routerFallback telemetry added — made the duplication observable
2026-03-25 | e8d3c5b | @lramos15 | skipRouter/turnCount, conversation-level routing finalized


<h3>Minimal fix (this PR)</h3>
<p>Promise coalescing in <code>resolveAutoModeEndpoint</code>: if a resolution is already in-flight for a <code>conversationId</code>, subsequent callers share the existing promise instead of starting a new one.</p>
<h3>Proper fix (separate PR)</h3>
<p>Resolve the endpoint once at the top of <code>runOne()</code> and pass it down to <code>getAvailableTools</code>, <code>buildPrompt</code>, and <code>fetch</code>. See <code>abadawi/resolve-endpoint-once-per-round</code>. This was attempted in <code>5388442bf</code> (added <code>_cachedEndpointForTurn</code> to <code>ToolCallingLoop</code>) but reverted in favor of dedup inside <code>automodeService</code>. Now implemented as a standalone refactor.</p>
<p>cc @lramos15 @TylerLeonhardt — adding you both since you've touched this area extensively. Would appreciate a sanity check on these findings.</p></body></html>
</body>
</html>

src/platform/endpoint/node/automodeService.ts
abadawi591: good point! Will add this. Thanks

alexdima:

PR #5040, opened 2026-04-07
Send has_image to router instead of skipping it for vision requests

abadawi591:
The server-side auto-intent-service now handles has_image via a vision-aware pre-filter instead of returning HTTP 400. Remove the client-side hasImage early return in _tryRouterSelection() and send has_image: true in the router request body when images are present.

Changes:
- automodeService.ts: remove hasImage bail-out, pass hasImage to router
- routerDecisionFetcher.ts: accept hasImage param, include has_image in request body, parse error code from 400 responses
- automodeService.ts: handle no_vision_models error as a distinct fallback reason; _applyVisionFallback remains as safety net
- automodeService.spec.ts: replace skip-router test with send-has_image test, add no_vision_models error handling test, update transient fallback test

### DIFF
--- a/src/platform/endpoint/node/automodeService.ts
+++ b/src/platform/endpoint/node/automodeService.ts
@@ -201,7 +201,7 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 					"automode.routerFallback" : {
 						"owner": "lramos15",
 						"comment": "Reports when the auto mode router is skipped or fails and falls back to default model selection",
-						"reason": { "classification": "SystemMetaData", "purpose": "PerformanceAndHealth", "comment": "The reason the router was skipped or failed (hasImage, noMatchingEndpoint, routerError, routerTimeout)" }
+						"reason": { "classification": "SystemMetaData", "purpose": "PerformanceAndHealth", "comment": "The reason the router was skipped or failed (noVisionModels, noMatchingEndpoint, routerError, routerTimeout)" }
 					}
 				*/
 				this._telemetryService.sendMSFTTelemetryEvent('automode.routerFallback', {
@@ -252,10 +252,6 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 		const prompt = chatRequest?.prompt?.trim();
 		const lastRoutedPrompt = entry?.lastRoutedPrompt ?? prompt;
 
-		if (hasImage(chatRequest)) {
-			return { lastRoutedPrompt, fallbackReason: 'hasImage' };
-		}
-
 		if (!this._isRouterEnabled(chatRequest) || conversationId === 'unknown') {
 			return { lastRoutedPrompt };
 		}
@@ -277,7 +273,7 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 				previous_model: entry?.endpoint?.model,
 				turn_number: (entry?.turnCount ?? 0) + 1,
 			};
-			const result = await this._routerDecisionFetcher.getRouterDecision(prompt, token.session_token, token.available_models, undefined, contextSignals, chatRequest?.sessionId, chatRequest?.id);
+			const result = await this._routerDecisionFetcher.getRouterDecision(prompt, token.session_token, token.available_models, undefined, contextSignals, chatRequest?.sessionId, chatRequest?.id, hasImage(chatRequest));
 
 			if (!result.candidate_models.length) {
 				return { lastRoutedPrompt: prompt, fallbackReason: 'emptyCandidateList' };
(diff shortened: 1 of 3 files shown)


src/platform/endpoint/node/automodeService.ts
```diff
@@ -297,7 +293,8 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 			return { selectedModel, lastRoutedPrompt: prompt };
 		} catch (e) {
 			const isTimeout = isAbortError(e);
-			const fallbackReason = isTimeout ? 'routerTimeout' : 'routerError';
+			const errorCode = e instanceof RouterDecisionError ? e.errorCode : undefined;
+			const fallbackReason = isTimeout ? 'routerTimeout' : errorCode === 'no_vision_models' ? 'noVisionModels' : 'routerError';
```
jinu-jang, line 297: Nested ternaries are just a no-go for me. Just change this into some if-statements.

src/platform/endpoint/node/automodeService.ts
```diff
@@ -297,7 +293,8 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 			return { selectedModel, lastRoutedPrompt: prompt };
 		} catch (e) {
 			const isTimeout = isAbortError(e);
-			const fallbackReason = isTimeout ? 'routerTimeout' : 'routerError';
+			const errorCode = e instanceof RouterDecisionError ? e.errorCode : undefined;
```
jinu-jang, line 296: Looking at the usage of RouterDecisionError, I'm not sure this class is required. The entire class is declared just to carry `e.errorCode`, a string value read once.

src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -32,6 +32,12 @@ export interface RoutingContextSignals {
 	prompt_char_count?: number;
 }
 
+export class RouterDecisionError extends Error {
```
jinu-jang, line 35: Error classes should have a `this.name` declared.

src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -73,7 +82,12 @@ export class RouterDecisionFetcher {
 		}
 
 		if (!response.ok) {
-			throw new Error(`Router decision request failed with status ${response.status}: ${response.statusText}`);
+			const errorText = await response.text().catch(() => '');
+			let errorCode: string | undefined;
+			try {
+				errorCode = JSON.parse(errorText).error;
+			} catch { /* not JSON */ }
+			throw new RouterDecisionError(`Router decision request failed with status ${response.status}: ${response.statusText}`, errorCode);
```
jinu-jang, line 90: I'm not sure why it is fine to completely ditch `errorText` when we couldn't parse errorText.

```suggestion
			} catch { /* not JSON */ }
			throw new RouterDecisionError(`Router decision request failed with status ${response.status}: ${response.statusText}`, errorCode || errorText);
```
Maybe?

jinu-jang (changes requested):

scutuatua-crypto (approved):

src/platform/endpoint/node/automodeService.ts
```diff
@@ -201,7 +201,7 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 					"automode.routerFallback" : {
 						"owner": "lramos15",
 						"comment": "Reports when the auto mode router is skipped or fails and falls back to default model selection",
-						"reason": { "classification": "SystemMetaData", "purpose": "PerformanceAndHealth", "comment": "The reason the router was skipped or failed (hasImage, noMatchingEndpoint, routerError, routerTimeout)" }
+						"reason": { "classification": "SystemMetaData", "purpose": "PerformanceAndHealth", "comment": "The reason the router was skipped or failed (emptyPrompt, emptyCandidateList, noMatchingEndpoint, noVisionModels, routerError, routerTimeout)" }
 					}
 				*/
 				this._telemetryService.sendMSFTTelemetryEvent('automode.routerFallback', {
```
aashna, line 207: Now that image requests go through the router successfully, there's no way to tell "router picked gpt-4o for a text query" apart from "router picked gpt-4o for an image query" in telemetry. Consider adding `hasImage: String(hasImage(chatRequest))` to `automode.routerFallback` properties here (and to `routerModelSelected` if it still exists) so we can measure vision routing quality separately.

src/platform/endpoint/node/automodeService.ts
aashna, line 297: Separate from the nested ternary issue jinu flagged — this silently swallows the server error code for non-vision 400s. If the server returns 400 with `empty_prompt` or some future error code, this maps it to generic `routerError` and we lose the real reason. Consider passing `errorCode` through directly:

```ts
let fallbackReason: string;
if (isTimeout) {
  fallbackReason = 'routerTimeout';
} else if (errorCode) {
  fallbackReason = errorCode;
} else {
  fallbackReason = 'routerError';
}
```

New server error codes surface in telemetry automatically without needing client-side mapping.

aashna (commented):
Nice work wiring the image signal through to the router. A couple things beyond jinu's comments.

alexdima:

PR #5048, opened 2026-04-08
Add candidateModel and stickyOverride to automode.routerDecision telemetry

aashna:
## Problem

The `automode.routerDecision` telemetry event only logged `predictedLabel` and `confidence`, making it impossible to distinguish what model the router recommended vs. what model was actually used after sticky-provider and vision overrides. Both `model` and `modelInvoked` in `response.success` resolve to the same concrete model ID, so there was no way to detect when the router's recommendation was overridden.

## Fix

Add four new properties to the `automode.routerDecision` event:

| Property | Type | Description |
|---|---|---|
| `candidateModel` | string | The router's top pick (`candidate_models[0]`) before any overrides |
| `stickyOverride` | measurement | Whether the router applied a sticky override (1 or 0) |
| `conversationId` | string | Enables joining with `response.success` events |
| `vscodeRequestId` | string | Enables per-request correlation |

## How to use

Join `automode.routerDecision.candidateModel` with `response.success.model` on the same `conversationId`/`vscodeRequestId`. When they differ, the router recommended one model but sticky-provider affinity or vision fallback selected a different one.

## Changes

- **routerDecisionFetcher.ts**: Extended `getRouterDecision()` signature and GDPR annotations; added `candidateModel`, `stickyOverride`, `conversationId`, `vscodeRequestId` to telemetry payload
- **automodeService.ts**: Pass `conversationId` and `chatRequest.id` through both call sites


TylerLeonhardt:
The code is in the vscode repo now under extensions/copilot

PR #5051, opened 2026-04-09
Fix skillContentRead and readFileToolInvoked telemetry in readFileTool

AbhitejJohn:
Fixes `skillContentRead` telemetry to ensure each field has a proper prefix instead of using shorthand properties.\n\nAlso fixed `readFileToolInvoked` telemetry properties to use explicit key-value pairs instead of shorthand. Both events validated in local telemetry.


AbhitejJohn:
ok, this isn't really required. The reason I wasn't seeing this earlier is because in Dev mode certain events weren't really flushed. Guided by Copilot, I made a local change to add a condition `process.env['COPILOT_LOG_TELEMETRY'] === 'true'` [here](https://github.com/microsoft/vscode-copilot-chat/blob/main/src/extension/extension/vscode-node/services.ts#L315)  and that got these events to flow through, not the changes in this PR. It would be nice to enhance dev mode for scenarios like these.

PR #5056, opened 2026-04-10
Add `vscode-chat-response-resource://` URIs to allowlist to fix chat response resource reads

Bestra:
In microsoft/vscode#308877 we're running into a problem where the resources from the GitHub MCP server (specifically for copilot spaces) come in under URIs like: `vscode-chat-response-resource://...`

These aren't part of the allowlist for readonly files so the model can't access them.

## Summary
- centralize the read-only allowlist for URIs that are readable without workspace membership
- allow `vscode-chat-response-resource` URIs through both read-file access checks
- add regressions for unopened chat response resource reads

## Testing
- npx vitest run src/extension/tools/node/test/toolUtils.spec.ts
- npx vitest run src/extension/tools/node/test/readFile.spec.tsx -t "reads chat response resources without requiring an open tab"

I also tested this locally in my Insiders instance using the launch copilot extension run and debug. 
CleanShot 2026-05-13 at 14 32 49@2x

and it works swimmingly. 

CleanShot 2026-05-13 at 14 30 24@2x

### DIFF
--- a/src/extension/tools/node/test/readFile.spec.tsx
+++ b/src/extension/tools/node/test/readFile.spec.tsx
@@ -343,6 +343,38 @@ suite('ReadFile', () => {
 			testAccessor.dispose();
 		});
 
+		test('reads chat response resources without requiring an open tab', async () => {
+			const resourceUri = URI.parse('vscode-chat-response-resource://session/tool/call-1/0/file.md');
+			const resourceDoc = createTextDocumentData(resourceUri, 'linked resource line 1\nlinked resource line 2', 'markdown').document;
+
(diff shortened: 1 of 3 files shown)


Bestra:
I'll try this out locally before undrafting

Bestra:
I've tested this out on my machine and it works, i've added a screenshot to the description. I think this is ready for review.

Bestra:
Closing this PR, this has been fixed 🙇  https://github.com/microsoft/vscode/commit/19d4617ad02271bfd8d67584c0fe60427a8fc9b0

PR #5072, opened 2026-04-15
Remove default truncation of otel trace and make it configurable via env vars

Rilomilo:
fix(otel): remove default truncation of trace attributes; make it configurable via env var

Previously truncateForOTel() silently capped every attribute at 64 000 characters, producing broken JSON that Langfuse and other observability platforms could not parse.

The new behaviour sends complete data by default.  Truncation can be re-enabled by setting COPILOT_OTEL_MAX_ATTRIBUTE_LENGTH to a positive integer (e.g. export COPILOT_OTEL_MAX_ATTRIBUTE_LENGTH=64000).

- Remove MAX_OTEL_ATTRIBUTE_LENGTH constant and the hardcoded 64 K default
- truncateForOTel() now reads COPILOT_OTEL_MAX_ATTRIBUTE_LENGTH at call time; an explicit maxLength argument still takes precedence
- Update tests: replace the "default 64000" assertions with no-truncation by default, and add coverage for the env-var path and invalid values
- Update stale comment in genAiEvents.ts

https://claude.ai/code/session_01WuZzmxwP2ycygD1kGjJWhV

### DIFF
--- a/src/platform/otel/common/genAiEvents.ts
+++ b/src/platform/otel/common/genAiEvents.ts
@@ -51,7 +51,7 @@ export function emitInferenceDetailsEvent(
 		attributes[StdAttr.ERROR_TYPE] = error.type;
 	}
 
-	// Full content capture with truncation to prevent OTLP batch failures
+	// Full content capture (truncation only if COPILOT_OTEL_MAX_ATTRIBUTE_LENGTH is set)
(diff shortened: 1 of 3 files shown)


zhichli:
please open the PR in vscode repo since we moved chat ext there.

PR #5080, opened 2026-04-20
Change providers to accept extra parameter

glGarg:
Adds support for passing custom fields via an `extraBody` property throughout the BYOK (Bring Your Own Key) model configuration and request pipeline. This enables providers to specify additional request parameters (such as special options for certain endpoints) in a flexible way. The changes ensure that these custom fields are available in model metadata, configuration, and are included in outgoing requests.

### DIFF
--- a/src/extension/byok/common/byokProvider.ts
+++ b/src/extension/byok/common/byokProvider.ts
@@ -61,6 +61,7 @@ export interface BYOKModelCapabilities {
 	supportedEndpoints?: ModelSupportedEndpoint[];
 	zeroDataRetentionEnabled?: boolean;
 	supportsReasoningEffort?: string[];
+	extraBody?: Record<string, unknown>;
 }
 
 export interface BYOKModelRegistry {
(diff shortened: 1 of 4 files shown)


glGarg:
This is how I set the thinking parameter in the config:

```
customModels:
  customoai:
    accounts/msft/deployments/wxbsoe65:
      name: 'wxbsoe65'
      url: 'https://api.fireworks.ai/inference/v1/chat/completions'
      toolCalling: true
      vision: false
      thinking: false
      maxInputTokens: 128000
      maxOutputTokens: 16384
      requiresAPIKey: true
      extraBody:
        thinking:
          type: disabled
```

battaglia01:
I was just looking for this feature and am glad it is being added, as we can then use reasoning in OpenRouter.

@glGarg Would we put this in `chatLanguageModels.json` or somewhere else? The config you have above doesn't look like json.

alexdima:

PR #5093, opened 2026-04-28
Fix read_cell_output incorrectly reporting all outputs as too large

oded-ist:
Two related issues caused even tiny notebook cell outputs (e.g. 5 bytes) to be replaced with "Output is too large to be used as context".

1. Inverted size check in `RunNotebookCellOutput`: `getCharLimit` converts tokens to chars (×4), but it was applied to the byteLength side and compared against a token count. Compare byteLength against getCharLimit(tokenBudget / sizeLimitRatio) instead — the threshold was 16× too small.

2. Remove `ReadCellOutput` from `toolsCalledInParallel`. Tools in that set are invoked eagerly with a sentinel `{ tokenBudget: 1 }` sizing on the premise that they don't consume sizing info. `RunNotebookCellOutput` does — it gates output on `sizing.tokenBudget` — so the sentinel made every non-empty output trip the size check. Letting it use the normal lazy path gives it the endpoint's real prompt budget.

### DIFF
--- a/src/extension/prompts/node/panel/toolCalling.tsx
+++ b/src/extension/prompts/node/panel/toolCalling.tsx
@@ -345,7 +345,6 @@ const toolsCalledInParallel = new Set<ToolName>([
 	ToolName.GetErrors,
 	ToolName.GetScmChanges,
 	ToolName.GetNotebookSummary,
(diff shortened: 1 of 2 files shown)


alexdima:

PR #5105, opened 2026-05-04
Improve linked symbol anchors with tree-sitter fallback

pranavvaid-ac:
Adds an eager tree-sitter fallback for linked-backtick symbol references so links open near the referenced symbol even before an LSP DocumentSymbolProvider result is available. Before this PR, links to referenced symbols would just take you to the top of the file if an LSP was not available.

Changes:
- Resolve initial linked symbol locations from tree-sitter when LSP symbols are unavailable.
- Prefer declaration-like tree-sitter symbols before generic identifier matches.
- Cache parsed file symbols during linkification to avoid repeated parsing for the same file.
- Keep the existing LSP resolve path so symbol kind and more precise document-symbol locations can still upgrade later.
- Add tests for exact matches, qualified-name fallback, parser caching, linkifier integration, and LSP kind/location upgrades.


alexdima:

PR #5090, opened 2026-04-27
fix: block autopilot on user input requests

unveroleone:
## Summary
Autopilot was incorrectly auto-answering clarification prompts and continuing execution without waiting for real user input.  
This PR makes clarification questions in Autopilot a true pause point by removing the self-answering path and always routing through the existing question flow.

## Root Cause
In copilotcliSession.ts, the user_input.requested handler had an Autopilot-only shortcut that immediately replied with a canned freeform answer instead of waiting for the question UI result.

## What Changed
1. Removed the Autopilot auto-response branch from copilotcliSession.ts.
2. Kept the normal askUserQuestion path for user_input.requested in all modes, including Autopilot.
3. Added regression coverage in copilotcliSession.spec.ts to verify:
- Autopilot user_input.requested returns the selected choice from user input.
- Autopilot user_input.requested returns freeform text from user input.
4. Enhanced the test mock session in copilotcliSession.spec.ts so user_input requests can be emitted and awaited end-to-end.

## Behavior Impact
1. Clarification questions in Autopilot now require explicit user input before continuation.
2. Existing Autopilot behavior for permission auto-approval and exit plan mode remains unchanged.

## Validation
1. Targeted tests passed:
- npx vitest run copilotcliSession.spec.ts
- Result: 45 passed, 0 failed
2. Watch compilers reported no TypeScript errors:
- watch:tsc-extension
- watch:tsc-extension-web
- watch:tsc-simulation-workbench
3. esbuild watch completed builds successfully.

PR #5106, opened 2026-05-04
Compress tool output to reduce token usage

### LINKED ISSUE #315376: Compress tool output to reduce token usage
- New setting `chat.tools.compressOutput.enabled`, **off by default**, with `experiment: { mode: 'auto' }`.
- New telemetry event `toolResultCompressed` reporting `{ toolName, filters, beforeBytes, afterBytes }` so we can quantify savings.
- New `IToolResultCompressor` service in `src/vs/workbench/contrib/chat/common/tools/toolResultCompressor.ts` with a per-tool filter registry. Filters are pure functions over text parts. The service skips outputs under 80 bytes and drops any filter that throws (ztk's "never make it worse" rule).
- Three initial filters for `run_in_terminal` in `src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/terminalOutputCompressor.ts`:
  - **`git diff` / `git show`** — collapses unchanged context, drops `index`/`similarity index` headers, omits lockfile and snapshot diff bodies, preserves `+`/`-`/`@@` lines verbatim.
  - **`ls -l` / `ls -la`** — keeps just file names (with `/` for dirs), drops perm/owner/size columns and `total N`.
  - **`npm install` / `yarn` / `pnpm install`** — drops progress bars, `npm warn deprecated`, funding/audit nag lines.

meganrogge:
Fixes https://github.com/microsoft/vscode/issues/315376

Adds a post-processing compression layer for tool results, inspired by [ztk](https://github.com/codejunkie99/ztk). When the model invokes a tool, we now have a chance to filter the result text before it reaches the model — same information, fewer tokens.

## What

- New `IToolResultCompressor` service in `src/extension/tools/common/toolResultCompressor.ts` with a per-tool filter registry. Filters are pure functions over text parts. The service preserves stderr and exit codes, skips outputs under 80 bytes, and drops any filter that throws (ztk's "never make it worse" rule).
- Three initial filters for `run_in_terminal` in `src/extension/tools/node/compressors/terminalOutputCompressor.ts`:
  - **`git diff` / `git show`** — collapses unchanged context, drops `index`/`similarity index` headers, omits lockfile and snapshot diff bodies, preserves `+`/`-`/`@@` lines verbatim.
  - **`ls -l` / `ls -la`** — keeps just file names (with `/` for dirs), drops perm/owner/size columns and `total N`.
  - **`npm install` / `yarn` / `pnpm install`** — drops progress bars, `npm warn deprecated`, funding/audit nag lines.
- Wired into `ToolsService.invokeTool` as a single post-process step right after `vscode.lm.invokeTool` resolves. Works for any tool — including the core `run_in_terminal` — without core changes.
- New setting `github.copilot.chat.tools.compressOutput.enabled`, **off by default**.
- New telemetry event `toolResultCompressed` reporting `{ toolName, filters, beforeBytes, afterBytes }` so we can quantify savings.

## Why

The terminal tool routinely dumps tens of thousands of raw tokens (a single `git diff HEAD~5` can be 90KB+). Most of that is metadata, unchanged context, lockfile churn, or progress noise. Compressing this client-side before it hits the model is a cheap, predictable token win that doesn't change tool semantics.

## Tests

12 unit tests in `src/extension/tools/node/test/terminalOutputCompressor.spec.ts` covering command parsing, each filter's match logic, and the actual compression behavior. All passing.

## Scope notes

- Lives entirely in copilot-chat — no core API changes.
- Default off; opt-in via setting (and later, experiment).
- Easy to extend: add a new `IToolResultFilter` and register it. Future filters: test-runner output, `read_file` skeleton mode, `grep_search` per-file match cap, session-level dedup cache.

Screenshot 2026-05-04 at 8 53 52 PM

### DIFF
--- a/src/extension/tools/node/compressors/terminalOutputCompressor.ts
+++ b/src/extension/tools/node/compressors/terminalOutputCompressor.ts
@@ -77,9 +77,33 @@ export const gitDiffFilter: IToolResultFilter = {
 		let contextRun = 0;
 		let inBinaryOrLock = false;
 
+		// Pending hunk: we buffer the body so we can rewrite the `@@` header to
+		// reflect the line counts we actually emit (otherwise the diff is no
+		// longer valid unified-diff syntax and tools/agents that count lines
+		// inside a hunk get confused).
+		let pendingHunkHeaderIndex = -1;
+		let pendingHunkOldStart = 0;
+		let pendingHunkNewStart = 0;
+		let pendingOldLines = 0;
+		let pendingNewLines = 0;
+
+		const HUNK_RE = /^@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@/;
+
+		const flushHunk = () => {
+			if (pendingHunkHeaderIndex < 0) {
+				return;
+			}
+			out[pendingHunkHeaderIndex] = `@@ -${pendingHunkOldStart},${pendingOldLines} +${pendingHunkNewStart},${pendingNewLines} @@`;
+			pendingHunkHeaderIndex = -1;
+		};
+
 		const flushContextRun = () => {
 			const omitted = contextRun - KEEP_CONTEXT;
 			if (omitted > 0) {
+				// Note: this marker is intentionally not valid unified-diff syntax,
+				// but the surrounding hunk header counts are kept consistent with
+				// the prefixed (' ', '+', '-') lines we actually emit, so a parser
+				// that ignores unknown lines still reads correct counts.
 				out.push(`... ${omitted} unchanged context line${omitted === 1 ? '' : 's'} omitted ...`);
 			}
 			contextRun = 0;
@@ -88,6 +112,7 @@ export const gitDiffFilter: IToolResultFilter = {
 		for (const line of lines) {
 			if (line.startsWith('diff --git')) {
 				flushContextRun();
+				flushHunk();
 				inBinaryOrLock = /package-lock\.json|yarn\.lock|pnpm-lock\.yaml|\.lockb?$|\.snap$/.test(line);
 				if (inBinaryOrLock) {
 					out.push(line);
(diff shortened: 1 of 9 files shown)


roblourens (dismissed):

dmitrivMS (dismissed):

unveroleone:
Nice work! Maybe @roblourens or @meganrogge can look at my [PR](https://github.com/microsoft/vscode-copilot-chat/pull/5090)?

src/extension/tools/node/compressors/terminalOutputCompressor.ts
```diff
@@ -0,0 +1,240 @@
...
+	}
+	return { head, sub };
+}
+
+function isTerminalInput(input: unknown): input is ITerminalInput {
+	return typeof input === 'object' && input !== null && 'command' in input;
+}
+
+/**
+ * Compresses `git diff` / `git show` output by reducing context lines to a
```
roblourens, line 57: This is cool, some thoughts:
Have you considered doing this on the other end- modifying the git diff command to produce fewer context lines, instead of parsing the output? Same with ls.

If you remove context lines, then the line number reference is wrong, should we patch that so the agent doesn't get confused about which lines it's looking at?

How about telling it that the output is truncated, and giving it a way to read the non-truncated output? What if it actually needed to see the stuff that was removed, or what if there's a bug?

roblourens (dismissed):

src/extension/tools/node/compressors/terminalOutputCompressor.ts
meganrogge: Thanks! All three are fair - quick thoughts on each, but TL;DR I'd like to land this behind the experiment first and see whether it actually moves the needle before investing in the deeper versions of any of these.

Re modifying the command instead of parsing output: considered it. Works for `git diff -U1` and `npm install --no-progress`, but `ls -l` has no flag to drop the perms/owner/size columns, and rewriting arbitrary user commands means parsing pipes/redirects/aliases/already-set flags. Output filtering generalizes to future targets (`cargo build`, `pytest`, `docker build`) without per-tool flag specs, so I'll keep it as the backbone and consider command-rewrite as an additive optimization later.

Re line numbers off after dropping context - good catch. The hunk header counts (`@@ -10,7 +10,7 @@`) won't match the emitted body, so it's no longer a valid unified diff. I'll rewrite the counts to match in this PR. 

Re tell it it's truncated + give it a way to read the original: `gitDiffFilter` already emits `... N unchanged context lines omitted ...` markers (you can see in the screenshot above), but `ls`/`npm` filters drop silently and none of them announce which filter ran. I'll add a per-filter header at minimum. 

The "let it fetch the original" piece is more involved — would need `run_in_terminal` to always return a terminal id and `get_terminal_output` to serve completed-session output (it only does for timed out or async currently) — so I'd rather wait on the exp results before going down that path. If compression turns out to be worth keeping, we can consider it.

meganrogge:
Screenshot 2026-05-05 at 2 22 32 PM
Screenshot 2026-05-05 at 2 22 20 PM

dmitrivMS (dismissed):

benvillalobos (approved):

meganrogge:
Apparently this is the old repo. IDK where the new one is.

meganrogge:
Needs to be moved to core

